PR #7, opened 2026-03-26
feat: add review-pr skill and update documentation

alexandrocuma:


skills/review-pr/SKILL.md
```diff
@@ -0,0 +1,288 @@
...
+**If a PR number is given** (e.g. `/review-pr 42`):
+```bash
+gh pr view 42 --json number,title,headRefName,baseRefName,url,state,author
+gh pr diff 42   # GitHub-computed diff — already correct, skip Step 2
+```
+Record: `PR_NUMBER=42`, `BRANCH=headRefName`, `BASE=baseRefName`. Skip Step 2.
+
+**If a branch name is given** (e.g. `/review-pr feat/my-feature`):
+```bash
+git branch --show-current   # or use the given name
```
alexandrocuma: **[S]** `skills/review-pr/SKILL.md:41` — **[BLOCKING]**

`git branch --show-current` returns the checked-out branch, not the argument passed by the user; if the user runs `/review-pr feat/foo` while on `main`, this command returns `main` and the subsequent diff targets the wrong branch entirely.

**Why:** Wrong branch yields wrong diff — the review silently reviews the wrong code.
**Priority:** 90%
**Fix:** Remove this command; use the passed argument directly.

```bash
# Remove git branch --show-current from the branch-arg block.
# The branch name is already known from the user's argument — use it as-is.
```

skills/review-pr/SKILL.md
```diff
@@ -0,0 +1,288 @@
...
+gh pr view --head <branch> --json number,baseRefName 2>/dev/null   # check if PR exists
```
alexandrocuma: **[S]** `skills/review-pr/SKILL.md:42` — **[BLOCKING]**

`gh pr view` has no `--head` flag; this command fails silently (redirected via `2>/dev/null`) and returns nothing, causing the skill to assume no PR exists even when one does — skipping draft review posting entirely.

**Why:** Silent failure drops PR context; all inline comments become unreachable.
**Priority:** 95%
**Fix:** Replace with `gh pr list --head <branch>`.

```bash
gh pr list --head <branch> --json number,baseRefName --limit 1 \
  | python3 -c "import sys,json; d=json.load(sys.stdin); print(json.dumps(d[0]) if d else '')" 2>/dev/null
```

skills/review-pr/SKILL.md
```diff
@@ -0,0 +1,288 @@
...
+```
+```
+
+#### 5b. Create the pending review via gh api
+
+```bash
+REPO=$(gh repo view --json nameWithOwner --jq '.nameWithOwner')
+
+# Write the review payload to a temp file
+cat > /tmp/review_payload.json << 'PAYLOAD'
```
alexandrocuma: **[S]** `skills/review-pr/SKILL.md:170` — **[NON-BLOCKING]**

Review comment bodies containing backticks, double-quotes, or embedded newlines will produce malformed JSON from the heredoc approach; the GitHub API returns 422 with no actionable error, and the entire review is silently lost.

**Why:** Code suggestions routinely contain backticks and quotes — these break heredoc JSON.
**Priority:** 75%
**Fix:** Build the JSON payload with Python, not a heredoc.

```bash
python3 -c "
import json, sys
payload = {'body': REVIEW_BODY, 'comments': COMMENTS}
with open('/tmp/review_payload.json', 'w') as f:
    json.dump(payload, f)
"
```

agents/pr-review.md
```diff
@@ -54,12 +54,30 @@ Before reviewing, check if `codebase-navigator` has already mapped this project:
...
+- If a PR number is given: `gh pr diff <number>` and `gh pr view <number>` — GitHub computes the correct diff, skip fetch
+- If a branch is given or no argument: fetch first, then diff against remote refs (see below)
+
+**Always fetch before diffing local branches** — local refs may be stale:
+```bash
+git fetch origin --prune
+```
+
+**Detect base branch** (in order):
+```bash
```
alexandrocuma, line 66: **[I]** `agents/pr-review.md:66` — **[NON-BLOCKING]**

The agent's base branch detection stops at two sources (`gh pr view`, `gh repo view`) but omits the third fallback present in the skill (`git branch -r | grep -E 'origin/(main|master|develop)$'`); when both `gh` calls fail (e.g. no auth, no network), the agent has no base ref and the diff silently fails.

**Why:** Divergence from skill means offline/unauth runs behave differently per invocation path.
**Priority:** 50%
**Fix:** Add `git branch -r` fallback after `gh repo view`.

skills/review-pr/SKILL.md
```diff
@@ -0,0 +1,288 @@
...
+**If a branch name is given** (e.g. `/review-pr feat/my-feature`):
+```bash
+git branch --show-current   # or use the given name
+gh pr view --head <branch> --json number,baseRefName 2>/dev/null   # check if PR exists
+```
+
+**If no argument given**:
+```bash
+git branch --show-current       # <branch>
+gh pr view --json number,baseRefName 2>/dev/null   # check if PR exists for current branch
```
alexandrocuma: **[I]** `skills/review-pr/SKILL.md:48` — **[NON-BLOCKING]**

The no-arg flow queries PR number via `gh pr view --json number,baseRefName` but never shows an explicit shell assignment; Step 5b then uses `$PR_NUMBER` directly in the `gh api` URL, relying on the LLM to infer the variable name — which may produce a malformed API path.

**Why:** Implicit variable tracking across multi-step instructions is error-prone.
**Priority:** 55%
**Fix:** Add explicit assignment: `PR_NUMBER=$(gh pr view --json number --jq '.number')`.

skills/review-pr/SKILL.md
```diff
@@ -0,0 +1,288 @@
...
+
+You are a **Senior Staff Engineer** conducting a surgical pre-merge code review. Prioritize architecture, logic, and correctness over stylistic perfection. Be pragmatic, not pedantic.
+
+## Triggered by
+
+- User typing `/review-pr [branch-or-PR-number]`
+- `dev-agent` — after completing a feature branch, before opening a PR
+- `pr-review` agent — when a RISEN-format review is needed
+
+## When to Use
```
alexandrocuma, line 15: **[N]** `skills/review-pr/SKILL.md:16` — **[NON-BLOCKING]**

The claim "`pr-review` agent — when a RISEN-format review is needed" implies the agent delegates to this skill, but the agent has no skill invocation and produces a different format (Critical/Important/Minor vs RISEN); this creates a false dependency in the chaining graph.

**Why:** Dead-end relationship claim; following it produces no invocation path.
**Priority:** 30%
**Fix:** Remove claim or add explicit skill invocation to `pr-review.md`.

Issue #77, opened 2026-09-14
Epic: promo-campaign skill — app-agnostic reels, videos and carousels


alexandrocuma:
## Epic
Promotion has no home in the toolkit. The one working pipeline is buried in a consumer repo, tied to a single app and to the Android emulator, and it speeds footage up to fit. Make it a skill that works for any app.

## Goal
`/promo-campaign`: find an app's main features, choose one hook, storyboard a ≤20s video that is not rushed, capture it on the platform the app runs on, and compose a 9:16 reel (1080x1920) and a 4:5 carousel (1080x1350).

## Sub-tickets & critical path
- **A — #75 method, hook selection and storyboard rules** — blocks B (defines the instance-file contract)
- **B1 — #76 iOS Simulator capture and segment reel composer** — depends on A
- **B2 — #78 Android emulator adapter and carousel composer** — depends on B1; not on the first consumer's critical path

Critical path: **A → B1**. B2 follows B1. First consumer: the first consumer (a private app) (iOS reel, ≤20s, es), which depends on both.

## Non-Goals
- Posting, scheduling, music, paid ads
- Web or desktop capture in the first cut
- Physical-device capture

## Estimate: L (decomposed) — A: M, B1: M (new Maestro integration), B2: S–M

🤖 Refined via `/refine`.

alexandrocuma:
Progress: A (#75, PR #79) and B1 (#76, PR #82) are merged and released in **v0.6.0**. `/promo-campaign` now plans a campaign and, on iOS, seeds, captures and cuts a 1080x1920 reel at 1.0x. Remaining: **B2 #78** (Android emulator adapter + 4:5 carousel composer). The skill reports those as not installed until then. Follow-ups filed during delivery: #80 (opencode installer strips body `name:` lines), #81 (secret-guard blocks program text).

alexandrocuma:
Epic complete, and its home has moved. `/promo-campaign` shipped here (#75, #76) and was then removed in #86: marketing is not a phase of the development lifecycle, so it does not belong in a developer-experience toolkit.

The method, contract schema, both trap catalogues and the safety rules were written up in marketing-toolkit's `docs/SPEC.md` before deletion, and the remaining work is tracked there: https://gitlab.com/acumadev/marketing-toolkit/-/issues

PR #79, opened 2026-09-14
#75: promo-campaign skill — method, hook selection, storyboard rules

### LINKED ISSUE #75: A: promo-campaign skill — method, hook selection and storyboard rules
## Problem
There is no toolkit skill for promoting an app. The only working promo pipeline lives inside one consumer repo, a private consumer repo (`docs/marketing/` + the consumer's promo scripts). It is welded to that app and to one platform (Android emulator via `adb`), and it fits a 25.6s cut by playing at 1.4x, which is exactly the rushed result a promo should avoid.

This ticket writes the **method** as an app-agnostic skill: `skills/promo-campaign/SKILL.md`.

## Acceptance Criteria
- [ ] **Placement**: a skill only (`skills/promo-campaign/`), not an agent. It is filed as the toolkit's first **domain playbook**, a new category beside the lifecycle orchestrators and utilities, with a written entry test: app-agnostic, user-invoked, interactive, and not absorbable by an orchestrator. Front matter sets `disable-model-invocation: true`, so its description costs no per-session context; it stays invocable as `/promo-campaign`.
- [ ] Follows `docs/development/skill-authoring-guide.md` as a **User-Invoked Skill**: `name` matches the directory, role statement, **Triggered by** (user-invoked only), **When to Use** (with an out-of-scope list), a numbered and named process, safety rules, and a defined output format. SKILL.md stays under 500 lines.
- [ ] **Feature extraction** separates *discovery sources* (changelog, README, code) from the *claim authority* (the current store listing or landing page). It produces a ranked feature list, each feature tagged *visual* or *not visual* and citing a claim-authority quote. Features found only in discovery sources are reported as unclaimed and never captioned.
- [ ] **Hook selection**: exactly one hook per video, defined as the single on-screen moment that proves the rank-1 visual claim, landing at ≤ 5.0s.
- [ ] **Storyboard rules**, as checkable limits measured on the render (`ffprobe`): duration ≤ 20.0s (shorter wins); **1.0x, never sped up**; one idea per caption; each caption on screen ≥ 2.0s; at most 8 captions; end card ≤ 3.0s *including* its transition, with no caption overlapping it. **Cut order** when it runs long: (1) compress pre-hook setup until the hook lands ≤ 5.0s, (2) drop beats from the end, (3) drop repeated beats.
- [ ] **Claim rule**: every caption maps to a quoted line of the current store listing or landing page. **Badge rule**: a store badge appears only for a listing that is publicly installable when checked signed-out, or is marked `launch_day: true` with a do-not-post-before note.
- [ ] **Safety rules**: seed demo state before the first launch; never film real user data; no ad requests from release builds on unregistered devices; generated UI never stands in for the real app; no host leaks (clipboard, notifications, real clock or carrier); never fabricate testimonials, ratings or download counts; never modify store badge artwork; never write outputs into the skill's install directory.
- [ ] **Output set** defined: 9:16 reel (1080x1920) and 4:5 carousel (1080x1350). A 1:1 variant is out of scope for this epic unless #76 adds a composer for it.
- [ ] Defines the **instance-file contract** *inside SKILL.md*, because opencode installs SKILL.md only. The contract is a per-repo `docs/marketing/campaign.md` with YAML front matter `schema: promo-campaign/v1` and top-level keys:
  - `app` (name, icon, tagline per locale), `locales`, `max_duration_s`, `claim_sources`, `features`, `hook`, `beats`
  - `captions.<locale>[]` (start_s, end_s, text, emphasis, claim)
  - `end_card` (duration_s, badges[]: store, locale, path, public, launch_day)
  - `outputs` (dir, formats)
  - `carousel.<locale>[]` (still, text, emphasis, claim), naming stills captured under `capture.stills[]`
  - reserved `seed` and `capture` blocks, whose internals #76 defines

  The skill also creates or updates `docs/marketing/README.md` and the `docs/README.md` entry, and git-ignores `outputs.dir`.
- [ ] **Degrades without #76**: when the skill's `scripts/` directory is absent, it stops after writing the instance file and reports that capture and compose tooling is not installed.
- [ ] No app-specific names, coordinates or copy anywhere under `skills/promo-campaign/` (grep check). The toolkit is public and the consumer app is private, so **no the consumer app content is copied in**. Worked examples live in `references/`, are labelled EXAMPLE, use a fictional app, and pass every storyboard limit.
- [ ] Registered everywhere the toolkit catalogues skills, reframing "seven commands" to "eight: five lifecycle orchestrators, two utilities, one domain playbook":
  - `CLAUDE.md:5`
  - `docs/reference/skills.md`: `:3`, `:5`, the table, a new Domain Playbooks section, `:91`, Adding a New Skill
  - `skills/README.md`: `:9`, `:11`, the table, the diagram
  - `README.md`: `:5`, `:46`, the skills table, the repo tree
  - `docs/README.md`: `:11`, `:27`
  - `docs/coverage.md`: counts, plus a Domain playbooks row
  - `CHANGELOG.md` [Unreleased]
- [ ] `./install.sh --skills-only --dry-run`, run from the delivery worktree, lists `promo-campaign/`, and a real install places `SKILL.md` and `references/` under `~/.claude/skills/promo-campaign/`. `go test ./...` stays green (no Go changes).

## Non-Goals
- Capture and compose scripts (sibling ticket)
- Any consumer app's campaign content
- Posting, scheduling, music selection, paid ads

## Estimate: M
Documentation only, with no Go changes: one new skill directory, plus a count reframe across seven indexes and CHANGELOG.

Supporting files are not new (graphify already ships `references/`). The new patterns are a domain-playbook category and a versioned input contract that lives in consumer repos.

Source material to *generalise from, not copy*: the consumer app `docs/marketing/promo-video.md` and `social-post.md`, verified end to end in `es` on 2026-09-09. That cut breaks this ticket's own limits (25.6s at 1.4x, hook caption at 13.3s, a 1.7s caption), so it serves as a counter-example, and it lives in a private repo.

First consumer: the first consumer (a private app) (iOS reel, ≤20s, es).

Decided with the owner (2026-09-13): manual invocation only (`disable-model-invocation: true`); `premium-tier` does not join the domain-playbook category in this epic.

🤖 Refined via `/refine`. Groomed: https://github.com/alexandrocuma/devexp-toolkit/issues/75#issuecomment-5658832816

alexandrocuma:
Closes #75 · Epic #77

Adds `/promo-campaign`, the toolkit's first **domain playbook**: an app-agnostic, user-invoked skill that turns an app's real footage into a plan for a ≤20s reel and a 4:5 carousel. It writes that plan to a versioned instance file in the consumer repo. Capture and compose tooling is #76; until it lands, the skill stops after writing the instance file and says so.

## What changed
- **`skills/promo-campaign/SKILL.md`** (391 lines, under the 500-line guard). Front matter: `disable-model-invocation: true` and `argument-hint`.
  - Process phases 0–8 with three user checkpoints.
  - Storyboard limits, measured on the render: ≤ 20.0s, render = plan within one frame, 1.0x never sped up, captions ≥ 2.0s and ≤ 8, hook ≤ 5.0s, end card ≤ 3.0s including its transition.
  - Cut order, the claim rule and the badge rule.
  - An inline, parser-proof `promo-campaign/v1` contract. `seed` and `capture` are reserved for #76.
  - 8 safety rules, an output template, and an embedded Ruby limits check (stdlib only; PyYAML is not assumed).
- **`references/example-campaign.md`**: a fictional app ("Tallybird") in en and es that passes every limit, with the arithmetic shown. **`references/campaign-template.md`**: a blank skeleton.
- **Registration**, reframed from seven commands to eight (5 lifecycle + 2 utilities + 1 domain playbook): `CLAUDE.md`, `README.md`, `docs/README.md`, `docs/coverage.md`, `docs/reference/skills.md` (new Domain Playbooks section with the entry test), `skills/README.md`, and `CHANGELOG.md` [Unreleased]. **No version bump**: the release waits for #76.
- **No Go changes.** The installer discovers `skills/*/SKILL.md` and copies whole directories (`installer_test.go:44-59`).

## Verification (run in the worktree)
| Check | Result |
|---|---|
| `scripts/stage-assets.sh` + `go test ./... -race -cover` | 10 packages ok (baseline before changes: the same 10) |
| `./install.sh --skills-only --dry-run` | 8 skills, including `promo-campaign/` (baseline: 7) |
| Sections, front matter, line count | 6 required sections, `disable-model-invocation: true`, 391 lines |
| Grep for app-specific or private content under `skills/promo-campaign/` | empty |
| Grep for stale "seven commands" wording (excluding CHANGELOG) | empty |
| Non-markdown files in the skill directory | none (nothing binary for `go:embed`) |
| Embedded Ruby check extracted from SKILL.md, run on the example with `/usr/bin/ruby` | `all limits pass`, exit 0 |

**Mutation record for the embedded check.** It is the guard, so it was made to fail on purpose.
- **Red (exit 1):** caption shortened to 1.9s; unquoted `schema`. Earlier implementation pass, all 12 red: hook at 5.1s, hook outside its beat, 20.5s total, end card 3.2s, caption over the end card, two emphasised captions, empty claim, undeclared still, unquoted string, `0:02.4`, schema v2.
- **Green controls (exit 0):** a shorter valid cut, `max_duration_s` lowered to 16, no carousel.
- **Parser disagreement, observed:** PyYAML reads `lands_at_s: 0:02.4` as 2.4 and passes it; Ruby's Psych reads 144.0 and fails. That is why the contract requires decimal seconds and quoted strings.

## Decisions and deviations
- **Placement** (owner decision): a skill only, not an agent, filed as a new domain-playbook category with a four-part entry test.
- **Folded in from #76's grooming:**
  - the `carousel.<locale>[]` key, whose `still` names an id under `capture.stills[]`
  - the parser-proof YAML rules
  - render duration = plan within one frame, since simulator recorders drop static tails silently
  - `${CLAUDE_SKILL_DIR}` treated as a SKILL.md text substitution, not an environment variable
- **Safety rule 5** says save and restore the host clipboard, never just empty it (#76 grooming W8).
- **Optional contract fields:** `end_card.transition_s` and `badges[].do_not_post_before`, so the badge rule can be checked by script.
- **The skill touches the reserved `capture` block:** it may declare `capture.stills[].id` while planning and never writes anything else in `seed` or `capture`.

## Follow-ups (not in this PR)
- The opencode installer strips **every** `name:` line from SKILL.md, not only the front matter's (`cli/internal/skills/installer.go:106-112`). The contract skeleton works around it, and the gotcha is documented under Adding a New Skill. Filed as #80.
- Found during grooming: `skill-authoring-guide.md:13-19` and `:327` contradict all seven existing skills; `templates/skill-template.md:4` contradicts the guide on name prefixes; `docs/guides/install.md:46` gives the wrong opencode path.
- A real `./install.sh` plus a `/promo-campaign` run in a throwaway repo (plan Step 5, check 7) happens after merge, from the main checkout.

https://claude.ai/code/session_01C16btGMR8NcNwfGtcgYiYy


alexandrocuma (commented):
## Review of #79 (`5e6b768`)

**Verdict: MERGEABLE AFTER FIXES.** The method, contract and registration are sound, and the example obeys every stated rule. Three correctness bugs need fixing first: the embedded check crashes on the most likely YAML mistake, the `.gitignore` append can corrupt a consumer repo, and Phase 8 can pass on stale frames.

### What I ran
- Extracted the Phase 6 script with the awk one-liner and ran it under `/usr/bin/ruby` 2.6.10 (Psych 3.1.0) and Homebrew Ruby 4.0.6 (Psych 5.3.1). Both give identical results.
  - `example-campaign.md`: `all limits pass`, exit 0.
  - `campaign-template.md`: parses, then fails the limits as expected (exit 1).
- Ran 30 fixtures, each a one-line mutation of the example, with exit codes captured under bash (the table under Blocking 1 lists them).
- Ran the Phase 6 bash block exactly as it sits in SKILL.md, with its 3-space list indentation, on a pass case and a fail case.
- Ran the `.gitignore` snippet on edge cases, and ffmpeg 9.0.1 on the Phase 8 frame-pull line.
- Grepped the diff for private or app-specific content: clean. Grepped the whole repo for stale counts: none outside CHANGELOG history.

### Acceptance criteria (#75)

| AC | Status | Evidence |
|---|---|---|
| Placement / domain playbook / `disable-model-invocation` | Met | `SKILL.md:5`, `:12`; `docs/reference/skills.md:80-87` |
| Authoring guide, user-invoked, <500 lines | Met | name `:2`, role `:10`, Triggered by `:16-22`, When to Use + out of scope `:24-36`, Process `:42-259`, Safety `:334-346`, Output `:348-391`; 391 lines |
| Feature extraction: discovery vs. claim authority | Met | `:54-68`, `:85-98` |
| Hook: exactly one, rank-1 visual, ≤ 5.0s | Met in prose | `:100-109`. The script does not enforce "exactly one" (Should fix 2), and "rank-1 visual" is ambiguous (Should fix 6) |
| Storyboard limits measured on the render, cut order | **Partly** | Limits and cut order are at `:121-145`. Phase 8 (`:251-259`) measures neither speed nor the render's end-card length (Should fix 7) |
| Claim rule, badge rule | Met | `:155-167` |
| Safety rules (8) | Met | `:336-346` |
| Output set 9:16 + 4:5, no 1:1 | Met | `:34`, `:147`, `:320` |
| Instance-file contract inline, README index, gitignore | Met, with a bug | contract `:263-330`, index `:172`, gitignore `:173-178` (Blocking 2) |
| Degrades without `scripts/` | Met | `:237-240`. `${CLAUDE_SKILL_DIR}` is a real skill-mode substitution in Claude Code (confirmed in the 2.1.270 binary) |
| No app-specific content; EXAMPLE uses a fictional app and passes | Met | grep clean; example passes, and also obeys the rules the script doesn't check (hook caption 1.2–4.5 covers `lands_at_s` 2.4; 4 slides = 4 captions) |
| Registration everywhere | Met, one overclaim | `docs/README.md:11` (Should fix 5). 35 `agents/*.md` minus `README.md` = 34, and 34 + 8 = 42 is correct |
| `install.sh` dry run lists it; real install places `SKILL.md` + `references/`; `go test` | **Partly** | The real install was deferred to after merge (Should fix 8) |

---

### Blocking

**1. The limits check crashes instead of reporting, and the parser-proof lint never runs on unquoted dates.** `SKILL.md:185-197`
- `YAML.safe_load` runs at `:186`, before the lint at `:197`. An unquoted `confirmed_live: 2026-09-01` or `do_not_post_before: 2026-10-15` raises `Psych::DisallowedClass: Tried to load unspecified class: Date` with a stack trace, on both Rubies. Dates are one of the value types `:272` says must be quoted, and the lint exists to catch them.
- Missing or mistyped keys also crash with stack traces:

  | Input | Result |
  |---|---|
  | no `beats` | `undefined method 'each' for nil` |
  | renamed `end_card` | `undefined method '[]' for nil` |
  | caption without `start_s` | `comparison of Float with nil failed` |
  | no front matter | `no implicit conversion of nil into String` |
  | `target_s: "1.2"` | `String can't be coerced into Float` |
  | `lands_at_s: "2.4"` | `comparison of String with 5.0 failed` |

- The last two rows matter most. The lint's own message ("unquoted value") and the rule "every string value is double-quoted" both push an agent to quote `0:02.4` rather than convert it to `2.4`, and the next run then crashes.
- All of these exit 1, so the check fails closed. But an agent gets a Ruby trace instead of a reason to take back to Phase 4.
- **Fix** (tested on Ruby 2.6 and 4.0): lint from `Psych.parse` first, abort on lint errors before `safe_load`, and type-check the keys the arithmetic touches. This also fixes Should fix 1:
  ```ruby
  parts = File.read(ARGV[0]).split(/^---[ \t]*$/, 3)
  abort "#{ARGV[0]}: no YAML front matter between --- lines" unless parts.size == 3 && parts[0].strip.empty?
  text = parts[1]; err = []
  # ... lint lambda unchanged, but `n.start_line + 1` ...
  begin
    lint.call(Psych.parse(text))
  rescue Psych::SyntaxError => e
    abort "YAML syntax error: #{e.message}"
  end
  abort(err.join("\n")) unless err.empty?
  fm = YAML.safe_load(text)
  num = ->(v, where) { v.is_a?(Numeric) ? v : abort("#{where} must be a number in decimal seconds, got #{v.inspect}") }
  %w[beats end_card hook locales captions].each { |k| abort "missing key: #{k}" unless fm[k] }
  ```
  Then wrap every `target_s`, `duration_s`, `lands_at_s`, `start_s` and `end_s` read in `num.(...)`.
  - The unquoted date now reports `line 106: unquoted value "2026-10-15"`.
  - `target_s: "1.2"` now reports `beats[0].target_s must be a number in decimal seconds, got "1.2"`.

**2. The `.gitignore` append corrupts the consumer's last ignore rule.** `SKILL.md:176-177`
- If `.gitignore` doesn't end with a newline (common), `printf '%s\n' "$dir/" >> .gitignore` glues onto the last line. Measured: `node_modules/\n*.log` became `node_modules/\n*.logreleases/promo/\n`. That silently stops ignoring `*.log` in the user's repo, and the output never ignored the directory either.
- The `grep -qxF` also misses equivalent existing rules (`/releases/promo`, `releases/`), and an `outputs.dir` with a trailing slash appends `releases/promo//`.
- **Fix:**
  ```bash
  dir="${dir%/}"
  git check-ignore -q "$dir/" 2>/dev/null || {
    [ -s .gitignore ] && [ -n "$(tail -c1 .gitignore)" ] && printf '\n' >> .gitignore
    printf '%s/\n' "$dir" >> .gitignore
  }
  ```
  `git check-ignore` also makes the Output line "already ignored" (`:387`) accurate.

**3. Phase 8 frame pulls can inspect stale frames and still pass.** `SKILL.md:248`
- `ffmpeg -v error -ss <time> -i "$reel" -frames:v 1 /tmp/.promo-campaign-frame-<n>.png` has no `-y`.
- The normal loop is render, check, fix, re-render, check. On the second pass the file exists, stdin is not a TTY, and ffmpeg prints `Not overwriting - exiting` and **exits 0** (ffmpeg 9.0.1: rc=0, previous frame left in place). Every visual checklist item then passes against the previous render.
- **Fix:** add `-y`, or put a render id or `$$` in the frame filename.

---

### Should fix

**1. Lint line numbers are off by one.** `SKILL.md:193`
- `start_line + 2` reports line 22 for `title: no` on line 21, line 9 for `icon: 9:41` on line 8, and line 72 for `lands_at_s: 0:02.4` on line 71.
- It is correct only when a blank line follows the opening `---`, because `\s*` in `/^---\s*$/` consumes the newline.
- **Fix:** `split(/^---[ \t]*$/, 3)` with `start_line + 1`, as in Blocking 1. That form gives the correct line with or without the blank line.
- The `split(..., 3)` limit itself is fine: a `---` horizontal rule in the body still passes (tested).

**2. The script skips rules that are cheap to check, and the PR description claims one of them.**

All of these pass today (exit 0):
- a second `role: "hook"` beat ("exactly one hook per video" is an AC)
- `hook.beat` naming a `setup` beat
- `emphasis: true` on the last caption instead of the hook's. The rule at `:136` needs `start_s ≤ lands_at_s ≤ end_s`.
- `launch_day: true` with no `do_not_post_before`, and a badge with `public: false, launch_day: false`. The PR description says `do_not_post_before` was added "so the badge rule can be checked by script", but the script checks no badge at all.
- `transition_s: 9.0` with `duration_s: 2.5`
- every `confirmed_live: null` while captions exist (`:66` says nothing is captioned until then)
- `max_duration_s: 25.0`: silently capped at 20, although `:282` says it may never be raised
- `outputs.formats` containing `"carousel_4x5"` with no `carousel` block
- a caption `claim` found in no `features[].claim.<locale>`
- a feature with `status: "unclaimed"` and `rank: 6`

**Fix:** add these as one-liners. Or, if the script is meant to stay arithmetic-only, remove the badge claim from the PR description and say in `:180` which rules it does not check.

**3. "Keys stay unquoted" (`:272`) is not parser-proof for locale keys.**
- A Norwegian campaign (`locales: ["no"]`, `captions: { no: [...] }`) loads the key as `false`. The check then reports the misleading `no: captions need exactly one emphasis`, and PyYAML does the same.
- The template already quotes `"<locale>":`, so the rule and the template disagree.
- **Fix:** require quoted locale keys (`"en":`). Optionally, lint plain mapping keys matching `/\A(y|yes|n|no|on|off|true|false)\z/i`.

**4. Platform scope has nowhere to go in the contract.**
- Phase 0 (`:46-48`) and `argument-hint` scope a run to a platform, but `promo-campaign/v1` has no platform key.
- So `/promo-campaign ios` and `/promo-campaign android` write the same file with nothing saying which platform the beats, stills and authority were planned for. #76's capture tooling will need exactly that.
- **Fix:** add an optional `platforms: ["ios"]` key (additive within v1), or state that platform lives under `capture`. Also say what a scoped refresh does to the other platform's authority and badge entries.

**5. `docs/README.md:11` says the Quick Start uses "the 8 commands end-to-end".** `docs/guides/quickstart.md` covers only the five lifecycle commands, with no `/graphify`, `/cleanup` or `/promo-campaign`. The claim was already false at 7, and the PR restates it. **Fix:** "install, then use the lifecycle commands end-to-end", or add a short "Utilities and playbooks" section to the quickstart.

**6. "Rank-1 visual" can be unsatisfiable as written.**
- Ranking (`:92-95`) puts prominence first and prefers visual only at equal prominence. So rank 1 can be a feature that isn't visual, such as "Works offline" in a subtitle.
- `:102`, `:315` and the template's `hook.feature` comment then point at a feature that doesn't exist.
- **Fix:** "the highest-ranked claimed feature with `visual: true`".

**7. Phase 8 does not measure two limits it says are measured on the render.** `:251-259`
- **Speed.** No checklist item covers it; "watched once at 1.0x" is not a check.
- **End card ≤ 3.0s including transition.** Only "no caption" is checked.
- **Fix:** add "a frame at `card_start − 1/fps` shows the last beat with no transition started", plus "the compose tooling's segment lengths sum to the beat targets (no `setpts`/`atempo`/speed filter)". The second item can be marked pending #76.

**8. Real-install AC not yet shown.** `HOME=$(mktemp -d) ./install.sh --skills-only`, run from the worktree, then `ls "$HOME/.claude/skills/promo-campaign/references"`, satisfies the AC without touching the real home.

**9. Phase 1 "cookieless fetch" (`:64`) does not name a tool, and a fetch cannot show installability.**
- An agent with a browser-automation MCP will reach for it, and that browser is usually signed in: the opposite of a signed-out read.
- Store pages are region-dependent, and a returned page does not show that the listing "offers the install" (`:164`).
- **Fix:** name a plain HTTP fetch (`WebFetch` or `curl -sL`, with the country stated), forbid browser tools that share the user's session, and ask the user to confirm the Install button in a private window whenever the fetch cannot show it.

---

### Nit

- **Caption-count derivation (`:130`).** `(20.0 − 3.0) ÷ 2.0 = 8.5 → 8` uses the *longest* end card, which gives the smallest room, not an upper bound (a 1.0s end card leaves room for 9). Call 8 a chosen cap that a full 3.0s end card happens to fit.
- **Confusing lint and cap messages.** `max_duration_s: null` produces `duration 16.0 > 0.0`. `~` and an empty `found_in:` produce `unquoted value "~"` / `""`; say "use null".
- **Hook window upper bound.** `between?(lo, hi)` includes `hi`, but `hi` is the next beat's first frame, so `lands_at_s: 4.5` passes on a beat that ends at 4.5. Use `lo <= x && x < hi`.
- **Heredoc terminator indentation (`:183`/`:230`).** The terminator is indented inside the list, so as copied `   RB` never closes the heredoc. It works only because bash reads to EOF and Ruby parses `RB` as a constant after `exit`. Move the block out of the list, or say to strip the indentation.
- **ffprobe duration (`:246`).** `format=duration` is the longest stream, so an audio track that outlasts the video passes. Use `-select_streams v:0 -show_entries stream=duration`.
- **`outputs.dir` validation (`:173`).** Also reject absolute paths. "Not under the skill's install directory" can't happen for a repo-relative path without `..`.
- **Specialist count.** `docs/reference/skills.md:117` says "~40 specialist capabilities" on a line this PR edits; `README.md:46` and `skills/README.md:11` say ~30.
- **Coverage date.** `docs/coverage.md:3`'s `Generated: 2026-06-09` is unchanged although the header was edited.

alexandrocuma:
Review findings addressed in `7dd1321`.

**Blocking**
- **B1: the limits check crashed instead of reporting.** It now lints before `safe_load`, rescues load errors, and type-checks keys by path before any arithmetic. An unquoted date, a missing `beats`/`end_card`/`start_s`, a quoted number, no front matter or a YAML syntax error each give a readable message and exit 1.
- **B2: the `.gitignore` append could corrupt the file.** It now skips a path that is already ignored (`git check-ignore`), adds a missing final newline first, and rejects absolute paths, `..` and `~`.
- **B3: Phase 8 could check stale frames.** Frames now go into a fresh `mktemp -d` with `-y`. Before the fix, a re-render left the stale frame in place with exit 0; after it, the frame reflects the new render.

**Should fix**
- **S1:** lint line numbers are correct with or without a blank line after `---`.
- **S2:** the check now enforces:
  - exactly one hook beat, and `hook.beat` names it
  - the emphasised caption is on screen at `lands_at_s`
  - the badge rule
  - `transition_s` ≤ `duration_s`
  - no captions or slides while `confirmed_live` is unset
  - `max_duration_s` may lower 20, never raise it
  - `carousel_4x5` requires slides
  - claims are verbatim in `features[].claim` or `app.tagline`
- **S3:** locale keys are quoted, and a `no:` key is linted.
- **S4:** `platforms` key added.
- **S5:** the Quick Start row no longer claims the 8 commands.
- **S6:** the hook proves the highest-ranked **visual** claimed feature, and a campaign with none stops.
- **S7:** Phase 8 checks 1.0x and the rendered end-card length.
- **S8:** a real install into a temporary `HOME` placed `SKILL.md` and `references/`, identical to the worktree.
- **S9:** Phase 1 names a plain HTTP fetch; `public: true` needs the user's signed-out confirmation.
- **Nits taken:** the caption-count wording, `ffprobe` reading the video stream, the hook window as `[start, end)`, and the `RB` terminator at column 0.

**Mutation record for this round**
- **Red (exit 1, readable message):** 18 B1 cases (no front matter, unquoted date, `0:02.4`, `9:41`, `no`, `~`, empty value, a Norwegian `no:` key, quoted numbers, missing keys, YAML syntax error, …) and 17 S2 cases.
- **Green controls (exit 0):** a `---` rule in the body, a shorter valid cut, `max_duration_s` 16, reel only with no carousel, a public badge with no date, a hook at the beat's start, a tagline claim.
- **Also verified:** `go test ./... -race` green; dry run lists 8 skills; SKILL.md is 468 lines; app-specific grep empty.

Issue #81, opened 2026-09-14
secret-guard blocks commands whose program text or heredoc merely mentions a key or dotenv file name


alexandrocuma:
## Problem
`hooks/claude-code/secret-guard.sh` treats every shell token of a Bash command as a file path (lines 22-51). It blocks the **whole command** when any token's basename ends in `.key`, `.pem`, `.p12` or `.pfx`, or starts with `.env.`. Program text and prose that merely contain such a string are refused:

- **jq filters.** `jq -r '.entries[].key'` is shlex-split into a token whose basename ends in `.key`, so the command is blocked. This happened during `promo-campaign` (#76): inline `jq` over seed entries was refused with `Blocked access to ".key"`, and the work had to move into a script file to get past the hook. Run from a file, the same filter is not blocked, so the guard obstructs without protecting anything.
- **Heredocs and prose.** A `gh issue create` whose heredoc body *mentions* a dotenv file name, while drafting this very issue, was blocked with `Blocked access to ".env.production"`. Nothing was being read.

## Acceptance Criteria
- [ ] A token counts as a path only if it plausibly is one: not inside a heredoc body; no program syntax (`|`, `[`, `]`, `{`, `}`, `(`, `)`, `$`, spaces after unquoting); and a basename that is more than the bare extension.
- [ ] Still blocked (red): reading a private key file with `cat`/`head`/`less`, copying a production dotenv file, reading a `.p12` bundle, and `Read` on a dotenv file.
- [ ] Allowed (green controls): `jq '.[].key' entries.json`, `jq -r '.key' x.json`, a quoted `python3 -c` program that mentions a `.pem` name, `echo .key`, and a heredoc whose text names a dotenv file.
- [ ] A test script beside the hook runs every case through the hook with the tool JSON it receives and asserts the exit code. Each red case is proven by reverting the fix.

## Non-Goals
- Scanning file contents for secrets (that is `secret-in-write-guard.sh`)
- Changing the Read-tool path check, which already works on real paths

Issue #78, opened 2026-09-14
B2: promo-campaign Android emulator capture adapter and carousel composer


alexandrocuma:
## Problem
#76 (B1) ships iOS capture and the reel composer. The epic's Android capture and 4:5 carousel remain. The consumer-side Android pipeline taps density-locked coordinates and seeds by root-only SQLite writes.

## Acceptance Criteria
- [ ] **Android emulator capture** (`capture-android.sh`) reusing B1's `lib/`: refuses anything that is not an emulator (`emulator-` serial and `ro.kernel.qemu=1`); every `adb` call pins `-s <serial>` and every Maestro call `--device <serial>`; drives the app with consumer-authored Maestro flows by on-screen text (verified to work on an API 34 emulator; off-screen elements need `scrollUntilVisible`), not coordinates.
- [ ] **Traps kept, generically**: SystemUI demo mode applied after any `wm density` change (the density value comes from the contract); a wait after focusing an input unless Maestro's `inputText` is shown to make it unnecessary; the host-clipboard chip handled by **saving and restoring** the host clipboard, never by silently emptying it.
- [ ] **Seeding backend `rn-asyncstorage` on Android** requires a `google_apis` image (release builds refuse `run-as`; `adb root` works) and writes `RKStorage` `catalystLocalStorage`. The first launch that creates `RKStorage` happens with the emulator's network off, or the schema is pre-created — whichever is proven. Ownership and SELinux context are restored (`chown` to the app uid, `restorecon`). A `google_apis_playstore` image fails with an actionable error.
- [ ] Records with `screenrecord` (respecting its time limit) and writes the same `<take>.cues.json` as B1, so `compose-reel.sh` works unchanged.
- [ ] **Carousel composer** (`compose-carousel.sh`): 1080x1350 slides from lossless screenshots (`simctl io screenshot` / `adb exec-out screencap -p`) plus captions, window derived from the source aspect (e.g. 444x966+318+300 for 1206x2622, 470x966+305+300 for 1080x2220), same frame and caption self-checks as the reel, closing slide from B1's end-card builder.
- [ ] Consumes the carousel slide key #75 adds (`carousel.<locale>[]`: still id, caption, emphasis, claim).
- [ ] Same script rules as B1: `/bin/bash` 3.2, `bash <path>`, no writes into the install dir, no binaries, guards proven by mutation.

## Non-Goals
- Physical devices; Play Store images; web or desktop capture

## Estimate: S–M
[...]

Issue #13, opened 2026-06-12
Story 1: Normalize Category B agents to canonical Persistent Agent Memory section (8 agents)


alexandrocuma:
Part of #12.

## Problem
8 agents already have a memory-write section, but not in the canonical `## Persistent Agent Memory` (h2) form from `templates/agent-template.md` (shipped in Gap 2, commit `e84345b`):

| Agent | Current heading | Line |
|---|---|---|
| `backend-senior-dev` | `# Persistent Agent Memory` (h1) | ~155 |
| `dev-agent` | `# Persistent Agent Memory` (h1) — also has separate `## Memory Protocol` at ~154 (Phase-0 consult step, leave untouched) | ~164 |
| `frontend-senior-dev` | `# Persistent Agent Memory` (h1) | ~124 |
| `dep-audit` | `## Memory Protocol` (~43) — this IS the write section here, no separate one | n/a |
| `runbook` | `## Memory Protocol` (~43, Phase-0 consult — leave untouched) + `### Phase 5: Update Memory` (~260, write section) | n/a |
| `migration` | `## Memory` | ~214 |
| `test-gen` | `## Memory` | ~155 |
| `grooming-agent` | `### Phase 9: Memory` | ~277 |

Reference implementations: `agents/codebase-navigator.md` and `agents/feature-path-tracer.md` (both already migrated in Gap 2).

## Acceptance Criteria
- [ ] Each of the 8 files has a `## Persistent Agent Memory` (h2) section, positioned per `templates/agent-template.md` (after Output Format / before closing TIPS block)
- [ ] Each retains its existing domain-specific "what to save" / "what NOT to save" content (e.g. dep-audit's known-acceptable-vulnerability list, grooming-agent's per-project ticket history, migration's gotchas/patterns)
- [ ] Both Gap 2 guardrail bullets are present verbatim or adapted to the agent's existing phrasing:
  - "Don't duplicate codebase-navigator's atlas (`~/.claude/agent-memory/codebase-navigator/<project-name>.md`) — record only what's unique to this agent's domain"
  - "If your memory files use a structure from an older version of this section, migrate them to the current structure as part of normal write-back"
- [ ] Earlier Phase-0-style "consult memory first" sections (dep-audit/runbook/dev-agent's `## Memory Protocol`) are left untouched — only the write-guidelines section is normalized
- [ ] No file ends up with two top-level "Memory"-named sections after the edit

## Non-Goals
[...]

PR #82, opened 2026-09-14
#76: promo-campaign iOS Simulator capture and segment reel composer

### LINKED ISSUE #76: B1: promo-campaign iOS Simulator capture and segment reel composer
## Problem
The method (sibling ticket) needs tooling that turns an instance file into footage and posts on any platform. The only existing implementation (the consumer app the consumer's promo scripts, about 900 lines of verified zsh) captures only from an Android emulator, using density-locked `adb input tap` coordinates and root-only SQLite seeding, and composes for one fixed 1080x2220 source.

## Acceptance Criteria
- [ ] **Contract parsing** without new dependencies: the YAML front matter of `docs/marketing/campaign.md` (`schema: promo-campaign/v1`, #75) is converted with `ruby` (`YAML.safe_load`, ships with macOS) to JSON and read with `jq`, falling back to `python3` with PyYAML, then `yq`. With none present the script exits non-zero naming all three. A `jq` pass type-checks every key the scripts read and names the failing path. The contract requires quoted strings and decimal seconds: YAML 1.1 reads `no` as false and `9:41` as a number, and Psych and PyYAML disagree on `0:05.5` (330 vs 5.5).
- [ ] **Defines the reserved `seed` and `capture` blocks** (a compact key table in SKILL.md, the full schema and a fictional example in `references/`): `seed.backend` + `seed.entries[]` (key, already-serialised value); `capture.ios` (app path, bundle id, device, status bar, starting appearance); `capture.takes[].steps[]` interleaving consumer-authored Maestro flow files with shell steps (`appearance`, `hold_s`); `capture.stills[]`; `capture.cut[]` segments (take, in_s, out_s, join `cut`|`fade`, fade_s). Captions stay #75's `captions.<locale>[]` on the output timeline.
- [ ] **Seeding is a declared backend**: `rn-asyncstorage` on iOS writes `Library/Application Support/<bundle-id>/RCTAsyncLocalStorage_V1/manifest.json` after `simctl uninstall` + `simctl install` and before the first `simctl launch`; a value over 1024 chars goes to a file named the lowercase MD5 hex of its key, with `null` in the manifest. Unknown backends fail naming the supported ones.
- [ ] **iOS Simulator capture** (`capture-ios.sh`): resolves one booted UDID; status bar `clear` then `override … --batteryState discharging`; records with `simctl io recordVideo --codec=h264` and drives only after `Recording started`; every Maestro call exports `DEVELOPER_DIR` when `xcode-select -p` is the Command Line Tools and passes `--device <UDID>`; suggestion taps use a relative selector (`below:`); the keyboard is dismissed with `pressKey: Enter`; appearance flips fire from the shell. A non-zero Maestro exit discards the take. On exit a trap clears the status bar and restores the starting appearance. Stills come from `simctl io screenshot` (lossless), never from take frames.
- [ ] **Cue sheet**: each take writes `<take>.cues.json` beside it with take seconds for every Maestro command (from `--test-output-dir` `commands.json` timestamps, named by `label:`), every shell step, `take_len_s` from the logged stop time, and the file's own duration.
- [ ] **Reel composer** (`compose-reel.sh`) builds the reel from `capture.cut[]` segments at 1.0x — `trim` + `setpts=PTS-STARTPTS`, hard `concat` or `xfade` — with no speed factor anywhere. It pads each take to its logged length with `tpad=stop_mode=clone`, because `recordVideo` writes no frames for a static tail and a trailing hold is otherwise lost. It applies `settb=AVTB,fps=30` immediately before every `concat`/`xfade` input, not only at the source, because `concat` re-stamps at 1/1000000 and `xfade` refuses mismatched timebases. It derives the phone window from the source aspect ratio (fit inside the band between captions and footer, even dimensions), overlays the `PNG32:` + `-alpha set` frame and ImageMagick caption PNGs with `shortest=1` on every overlay, and crossfades into an end card built with `-respect-parentheses`. Output: 1080x1920 H.264.
- [ ] **Self-checks fail the run**: frame `%[channels]` is `srgba`; each caption block fits its band; every segment out-point lies within its take's logged length; the filtergraph has no PTS scaling; `ffprobe` duration **equals** the planned length (Σ segments − Σ crossfades + end card − its fade) within one frame **and** is ≤ `max_duration_s` (default 20); size is 1080x1920; one frame per caption is written to `outputs.dir` for review.
- [ ] **Guards proven by mutation**, recorded in the PR: drop `PNG32:`/`-alpha set` (red), drop `shortest=1` (red, bounded by a timeout), drop `tpad` (red), normalise only at the source (red). Green controls: a single segment with no join, and a synthetic 1080x2220 `testsrc2` take that yields a 716x1470 window.
- [ ] No app-specific values inside the scripts; everything comes from the instance file, flags or env. Fonts come from the contract or env (macOS Arial default, existence-checked). Prerequisites (ffmpeg, ImageMagick, Maestro, Xcode, jq, perl, a YAML parser) are checked up front with install hints.
- [ ] Scripts target `/bin/bash` 3.2 (no `timeout`, `EPOCHREALTIME`, associative arrays or `mapfile`), are invoked as `bash <path>` because `devexp install` copies skill files with mode 0644 (`cli/internal/skills/installer.go:68`), and locate their siblings from `${BASH_SOURCE[0]}`. `${CLAUDE_SKILL_DIR}` is a SKILL.md substitution, not an environment variable, so SKILL.md passes paths explicitly.
- [ ] Scripts never write into `~/.claude/skills/promo-campaign/` (reinstall overwrites it, stale-manifest cleanup deletes it recursively). Takes, cue sheets, checks and renders go to `outputs.dir`; temporary files to `mktemp -d`.
- [ ] No store badge artwork, fonts or other binaries under `skills/promo-campaign/` (`go:embed all:skills`); badges are consumer-repo paths from `end_card.badges[]`, and are never modified.
- [ ] SKILL.md hands off to the scripts and degrades per component: until B2 ships it reports that the Android adapter and carousel composer are not installed. Documented limitation: opencode installs SKILL.md only, so capture and compose work in Claude Code only.
- [ ] Validated: `./install.sh --skills-only`, run from the delivery worktree, deploys `scripts/` and `references/` to `~/.claude/skills/promo-campaign/`; each script runs via `bash` from the deployed path; `go test ./...` stays green.

## Non-Goals
- Android emulator capture and the carousel composer (B2, separate ticket)
- Web or desktop capture; physical-device capture
- Generating Maestro flows from `beats` (flows are consumer-authored)
- Replacing the consumer app's the consumer's promo scripts (the consumer decides)

## Estimate: M
Proven while grooming on an iOS 26.5 Simulator:
- AsyncStorage seeding including a spilled value.
- `recordVideo` VFR output and its lost static tail.
- Shell appearance flips mapped onto the take to within about 0.1s.
- Maestro `commands.json` timestamps as a cue sheet.
- A 3-segment 1.0x cut with a crossfade, captions and an end card whose duration matched the plan exactly, with a measured on-screen event landing within one frame of prediction.
- Four guards proven red by mutation.

Critical path: #75 → #76 → the consumer app #26.

B2 (Android adapter + carousel) split out to #78.

🤖 Refined via `/refine`. Groomed: https://github.com/alexandrocuma/devexp-toolkit/issues/76#issuecomment-5658833247

alexandrocuma:
Closes #76 · Epic #77 · Follows #75 (#79)

B1 of the promo-campaign epic: the tooling that turns a `promo-campaign/v1` instance file into a 1080x1920 reel **cut** from real iOS Simulator takes at 1.0x. Android capture and the carousel composer are B2 (#78). SKILL.md says so and degrades per component.

## What's added
| File | Lines | What it does |
|---|---|---|
| `scripts/lib/common.sh` | 140 | `die` with exit codes, prerequisite check with install hints, perl hi-res time and timeouts (bash 3.2 has neither), `DEVELOPER_DIR` only when xcode-select points at the Command Line Tools, scratch dirs, font lookup |
| `scripts/lib/contract.sh` | 320 | YAML front matter → JSON via ruby, then python3+PyYAML, then yq. Same parser-proof lint as SKILL.md's check; jq type checks name the failing key path. `validate` / `json` subcommands |
| `scripts/lib/geometry.sh` | 127 | phone window from source aspect (even dims), frame layer with alpha check, caption layer with band check, end card (`-respect-parentheses`) |
| `scripts/seed-rn-asyncstorage-ios.sh` | 65 | AsyncStorage `manifest.json` plus spill files (>1024 UTF-16 units, named by the MD5 of the key) after install and before first launch. Refuses storage that already exists |
| `scripts/capture-ios.sh` | 262 | resolves one booted UDID. Per take: reinstall, seed, status bar, appearance, launch, record, run steps (consumer Maestro flows plus shell `appearance`/`hold_s`), write `<take>.cues.json`. Stills come from a separate non-recording pass. A trap restores the Simulator |
| `scripts/compose-reel.sh` | 256 | `capture.cut[]` segments at 1.0x, `tpad` to the logged take length, `settb=AVTB,fps=30` before every join, `shortest=1` on every overlay, end-card crossfade, self-checks |
| `references/capture-and-compose.md` | 243 | full `seed`/`capture` schema, cue-sheet format, render arithmetic, Maestro flow rules, trap catalogue, measured timings, mutation record |
| `SKILL.md` | 468 → 492 | Phase 7 names the real entry points, passes absolute paths via `bash "${CLAUDE_SKILL_DIR}/scripts/…"`, and degrades per component. Adds a compact `seed`/`capture` key table |
| `references/example-campaign.md`, `campaign-template.md` | | fictional `seed`/`capture` blocks. Example: 4.5 + 3.0 + 3.9 + 2.5 − 0.4 = 13.5s = Σ beats |
| `CHANGELOG.md`, `docs/reference/skills.md` | | #76 entry under Epic #77; one bullet naming the scripts, `bash <path>`, Claude Code only |

No Go changes. Scripts are `#!/bin/bash`, bash 3.2-compatible, invoked as `bash <path>` (install writes 0644), write only to `outputs.dir` and `mktemp -d`, and ship no binaries.

## Verification
**Composer, no device** (synthetic `testsrc2` takes with a cue sheet logging a longer take than the file, which reproduces the lost static tail):
- **3 segments** (a cut, a 0.3s fade), 2 captions, 3.0s end card: **10.300000s vs a 10.300000s plan**, 1080x1920, 309 frames. Re-run independently by the orchestrator on the pushed commit.
- **Single segment, no join:** 5.500000s = plan.
- **1080x2220 source:** window 716x1470+182+250, 7.000000s = plan.

**Mutation record.** Mutations were applied to copies of `scripts/`.

| # | Mutation | Result |
|---|---|---|
| M1 | drop `-alpha set` + `PNG32:` | **red** 70: frame hole alpha is 1, expected 0 |
| M1b | drop `PNG32:` only | green. ImageMagick 7.1.2 keeps alpha without it (recorded, not relied on) |
| M2 | drop `shortest=1` everywhere | **red** 70 after the 60s watchdog |
| M3 | drop the take `tpad` | **red** 70: 6.066016s is not the planned 10.300000s within one frame. **Re-run independently** |
| M4 | normalise only at the source | **red** 1: ffmpeg refuses `xfade` (timebase 1/1000000 vs 1/30) |
| M5 | Σ segments ≠ Σ beats | **red** 70: 7.6s against 7.3s. **Re-run independently** |
| M6 | `setpts=0.5*PTS…` in a trim | **red** 70: filtergraph re-times frames |
| M7 | `out_s` past the logged take length | **red** 70 |
| M8 | caption too tall for its band | **red** 70 |
| M9 | M3 with the duration check disabled | green at 6.07s, which proves the duration check is what catches M3 |

**Parser fixtures.** Each was run through SKILL.md's Ruby check and through `contract.sh` under ruby, python3 (ruby hidden) and yq (both hidden). All parsers agree:
- **Fully quoted:** passes.
- **Rejected, naming key path and line:** `locale: no`, unquoted `9:41`, `0:05.5`, an unquoted `no:` locale key.
- **Unknown `seed.backend`:** rejected with "supported backends: rn-asyncstorage".

**iOS end to end** on a booted iPhone 17 Pro Simulator (iOS 26.5) with a real React Native Release build. Campaign and flow were throwaway files outside the repo; no app identifier is in this PR.
- **Seed:** 4 entries including a 2273-char value. The spill file is named by the key's MD5 and matches the value byte for byte, with `null` in the manifest. First paint showed the spilled data.
- **Capture:** exit 0 in 27s. The cue sheet has Maestro command timestamps plus shell events: take 12.447s, file 10.487s.
- **Compose:** a segment ending inside the lost static tail, padded by 1.960s, rendered **7.400000s vs a 7.400000s plan**.
- **1.0x proof:** the dark-appearance flip crosses the luminance midpoint at 3.067s against a predicted 3.052s, within one frame.
- **Afterwards:** app uninstalled, no status-bar overrides, appearance light.

**Other checks:**
- `bash -n` passes on every script.
- `go test ./... -race -cover` is green.
- `./install.sh --skills-only --dry-run` lists 8 skills.
- A temp-HOME install deploys `scripts/` (with `lib/`) and `references/`, all `-rw-r--r--`. The deployed `compose-reel.sh --help` exits 0, and deployed `contract.sh validate` passes on the example.
- **Hygiene:** the grep for consumer literals, UDIDs and `/Users/` paths is empty; there are no non-md/sh files.

## Deviations from #76's text
1. **Planned length is `(Σ segments − Σ fades) + end_card.duration_s`**, not "… + end card − its fade". #75's rule makes `end_card.duration_s` include its transition. The crossfade runs over the footage's held last frame, with no caption over it. So the ticket's formula double-counts the fade, and the code follows #75.
2. **Frame self-check is `srgba` plus a transparent hole centre**, not channel count alone. With the frame built hole-first, dropping `-alpha set` left an opaque hole that still reported `srgba` (M1 vs M1b).
3. **`join` is the transition into the next segment.** Fonts and colours come from an optional `capture.compose` key, which env overrides, falling back to macOS Arial.
4. **`take_len_s` is logged before the stop signal** so it excludes file finalisation.
5. **`contract.sh` is stricter than SKILL.md's check on a few keys it reads:** `locales[]`, caption `text`, `app.name`/`icon`, badge `path`/`locale`.
6. **Traps found and fixed during the work,** each recorded in the references:
   - ffmpeg reads the caller's stdin, so every call passes `-nostdin`.
   - yq v4 drops fields when stdin is piped.
   - ImageMagick 7.1 prints `srgba 4.0`.

## Follow-ups (not in this PR)
- **#78 (B2):** Android emulator adapter and carousel composer.
- **#81:** `secret-guard` blocks commands whose program text or heredoc merely mentions a key or dotenv file name. It forced inline `jq` into script files during this work, and blocked filing its own issue.
- **Not automated:** CI still runs no shell checks. A `bash -n` step would catch syntax errors but not bash 3.2 incompatibilities, because Ubuntu ships bash 5.

https://claude.ai/code/session_01C16btGMR8NcNwfGtcgYiYy


alexandrocuma (commented):
## Review of #82 (closes #76) at `4c927c3`

**Verdict: mergeable after fixes.** The composer math, the device pinning and the trap all held up under test. The fixes below are a missing caption guard in the composer, one output-naming footgun, capture robustness, and a Phase 7 doc step that fails by construction.

### Blocking

1. **The composer doesn't check that captions stay clear of the end card, and its own comments say they do.**
   - **Where:** `compose-reel.sh:175-177` says "no caption rides into the transition", and so does `references/capture-and-compose.md:133`. That holds only when every `end_s` ≤ Σ beats. Nothing in `compose-reel.sh` or `contract.sh` checks it; only SKILL.md's Phase 6 Ruby check does.
   - **How it's reached:** Phase 7 only runs `contract.sh validate`. Step 3 then says "fix the cut or the storyboard" and re-run compose, which never re-runs the Phase 6 check.
   - **Reproduced:** in green3 I moved caption 2's `end_s` to 9.5, past the end-card start at 7.3. The SKILL check rejects it. `compose-reel.sh` exits 0 with 10.300000s = plan.
   - **Proof on the render:** the top-band mean at t=7.55 (mid-transition) is 0.9037, against 0.9151 for the baseline reel. At t=7.10 and t=8.5 the two are identical. So the caption is being blended into the crossfade.
   - **Fix:** in the render-arithmetic block (after `:94`), die 70 when any caption in any rendered locale has `end_s > BEATS_S + FRAME_S/2`. Also `start_s < 0` and `end_s ≤ start_s`, since the `enable` expression silently accepts both. #75 lists this limit as "measured on the render", so it belongs with the other self-checks.

2. **`--launch-day` writes the same file as the public render.**
   - **Where:** `compose-reel.sh:170` and `:246` write `reel-<locale>.mp4` either way. Only a stderr line (`:206-209`) marks it as not postable before `do_not_post_before`.
   - **Why it matters:** a launch-day run deletes the postable reel and puts a do-not-post reel under the identical name. #75's badge rule treats a badge for a store the app isn't live in as the strongest claim on the card.
   - **Reproduced** with a fixture whose only badge is `launch_day: true`. Without the flag the end card has no badge. With it, `reel-en.mp4` is replaced in place.
   - **Fix:** write `reel-<locale>.launch-day.mp4` (review frames likewise) when `launch_day=1`, and name the date in the final summary line. Update the outputs list in `references/capture-and-compose.md:24-33`.

### Should fix

3. **A failed recapture leaves the previous run's take in place.**
   - **Where:** `capture-ios.sh:184-189` deletes only the new `.mov` in `$T`. An earlier successful `takes/<id>.mov` + `<id>.cues.json` survive, and `compose-reel.sh` later renders from them without complaint.
   - **Reproduced:** stub Maestro exiting 1, with a stale `main.mov`/`main.cues.json` already in `takes/`. Both are still there after exit 1.
   - The composer already guards the equivalent case for reels (`compose-reel.sh:169-170`).
   - **Fix:** `rm -f "$OUT/takes/$id".{mov,cues.json} "$OUT/takes/$id".failed-step-*.log` at the top of `record_take`. Likewise remove each still's PNG before its pass.

4. **A skipped Maestro command crashes the run after a good recording.**
   - **Where:** `capture-ios.sh:213` divides `.metadata.timestamp` and `.metadata.duration` without a null guard. As far as I can tell, Maestro's command metadata leaves both unset for commands that never started (`SKIPPED`, e.g. a `runFlow` with an unmet `when:`). I reproduced with a stub `commands.json`, not real Maestro, so please confirm on a flow with a conditional.
   - **Result:** `jq: error … null (null) and number (1000) cannot be divided`, exit 5, the recording lost, and an empty `takes/main.cues.json` left behind, because the redirect truncates before jq fails.
   - **Fix:** `select(.metadata.timestamp != null)` before building the cue, `dur_s: ((.metadata.duration // 0) / 1000 | r3)`. Also write to a temp file and `mv` it into place.

5. **Phase 7 step 1 fails by construction.**
   - **Where:** `SKILL.md:318` runs `contract.sh validate <file>`, which defaults to `all`. `all` requires a non-empty `capture.cut` (`contract.sh:262`), but step 2 is where the in/out points come from, out of the cue sheets.
   - **Reproduced:** a campaign with `seed` + `capture.ios/takes` and no cut gives `capture.cut must be a non-empty list of segments, got nothing`. `validate … capture` passes.
   - **Fix:** step 1 runs `validate "<file>" capture`, and step 2's "validate again" runs `all`.

6. **Parser parity: anchors/aliases and duplicate keys.** The PR says all parsers agree. They don't on these two:
   - **Aliases.** `hold_s: &hold 1.0` then `hold_s: *hold`:
     - SKILL.md's check and `contract.sh` under ruby reject it: `Unknown alias: hold`, because `safe_load` has aliases off.
     - `contract.sh` under python3 with PyYAML (ruby hidden) prints `contract ok` in both `all` and `capture` modes.
     - By reading `yq_convert` (`:117-135`), yq's `-o=json '.'` resolves aliases too, so it accepts as well.
   - **Duplicate keys.** Ruby and PyYAML keep the last value silently. My fixture repeated `max_duration_s`, and both `json` outputs gave `30.0`. go-yaml v3 (yq v4) errors on a duplicate mapping key, again by reading only. SKILL.md's check caught that one instance only because 30 > 20; a duplicated `captions:` would silently drop a whole block.
   - **Fix:** add both to the lint on every path, so all parsers refuse before loading:
     - Psych: node `anchor` and `Psych::Nodes::Alias`, plus a seen-set of key values per mapping.
     - PyYAML: walk `yaml.parse()` events for `AliasEvent` or `anchor`, plus a seen-set of keys in `walk`.
     - yq: `.. | select(anchor != "" or alias != "")`.
     - Mirror it in SKILL.md's Ruby check so the two stay one rule.

7. **Capture deletes an existing install's data without asking, so the seeder's refusal never triggers in the capture path.**
   - **Where:** `prepare_app` (`capture-ios.sh:123-124`) uninstalls unconditionally, and the trap uninstalls again (`:64-67`).
   - The seeder correctly refuses existing storage. I verified exit 65 on a second seed, and the UTF-16 boundary: 512 astral characters stay inline, 513 spill, and the spill file names match `md5`. But whatever the developer had in that app on that Simulator is gone before seeding runs.
   - **Fix:** before the first take, `simctl get_app_container <U> <bid> data`. If it succeeds, die 65 unless a `--replace-installed` flag is given. Document it in the header and the reference.

8. **Safety Rule 5 is left to the operator.**
   - SKILL.md:440-443 (clipboard saved, cleared and restored; notifications silenced) is marked load-bearing. `capture-ios.sh` does neither, and references trap 15 hands it back as manual.
   - **Fix:** at minimum print a pre-flight reminder before the first take. Better: save `pbpaste`, clear with `pbcopy </dev/null`, and restore it in `cleanup`, text only, and say so.

### Nit

9. **`--locale` is matched as a regex.** `grep -qx -- "$only_locale"` (`compose-reel.sh:164`) lets `--locale e.` through: it rendered `reel-e..mp4` with 0 captions and exited 0. Use `grep -Fqx`.
10. **Locales are type-checked but not format-checked** (`contract.sh:249`). `for L in $LOCALES` (`compose-reel.sh:168`) word-splits and globs, so `"*"` expands to files in the cwd. Locales also build paths inside `rm -f` (`:170`). Add a slug rule, e.g. `^[A-Za-z]{2,3}([-_][A-Za-z0-9]+)*$`.
11. **`outputs.dir: "."` passes** (`contract.sh:185-186`), so `takes/`, `stills/`, `checks/` and reels land in the repo root. Reject `.` and a leading `./`.
12. **`capture.ios.bundle_id` is never validated**, yet it is joined into a filesystem path (`seed-rn-asyncstorage-ios.sh:49`). `get_app_container` gates it in practice. Check `^[A-Za-z0-9.-]+$` in `contract.sh:219`.
13. **An unfinalised recording fails without a `promo-campaign:` reason.** `stop_recording` (`capture-ios.sh:161-165`) ignores recordVideo's exit status. With a stub that leaves an unfinalised `.mov`, the run ends with ffprobe's `moov atom not found` and exit 1. Capture the status and die with a message.
14. **`run_step` runs with `set -e` off.** It is called as `run_step … || rc=$?` (`capture-ios.sh:183`), which disables `set -e` for its whole body. A failure in the `cj` substitutions at `:140-143` leaves `kind` empty, and the step is skipped with rc 0. Add `[ -n "$kind" ] || die 1 …`.
15. **The watchdog can be disabled by accident.** `with_timeout` passes its value to `alarm`, which takes whole seconds (`common.sh:40`), so `PROMO_RENDER_TIMEOUT_S=0.5` turns the watchdog off. Validate a positive integer.
16. **A failed still screenshot exits silently** (`capture-ios.sh:238`: `>/dev/null 2>&1` under `set -e`). The stills-pass failure message (`:235`) also tells the user the log will be deleted. Keep it the way take logs are kept.
17. **Still ids are slug-checked here but have no format rule in SKILL.md's Phases 1-6.** A planned `{ id: "hook total" }` passes Phase 6, then fails `contract.sh` in both modes. Add the slug rule to SKILL.md's contract table, or list it in deviation 5.

### Acceptance criteria

| # | AC | Status |
|---|----|--------|
| 1 | Contract parsing: ruby → python3+PyYAML → yq, die naming all three, jq type check naming paths | **Met, with a parity gap** (#6). `contract.sh:25-31`, `:143`, `:158-291` |
| 2 | `seed`/`capture` defined (SKILL.md table, full schema, example) | Met. `SKILL.md:415-421`, `references/…:46-84`, `example-campaign.md` |
| 3 | `rn-asyncstorage` seeding after uninstall+install, before launch; spill; unknown backend | Met. `capture-ios.sh:122-133`, `seed-rn-asyncstorage-ios.sh:47-64`, `contract.sh:208-209` |
| 4 | iOS capture: one UDID, status bar, recordVideo after `Recording started`, `DEVELOPER_DIR`, `--device`, shell flips, discard on failure, trap, lossless stills | Met. `capture-ios.sh:109-117`, `:129-131`, `:173-179`, `:146`, `:184-189`, `:57-74`, `:238`. Flow authoring rules: `references/…:147-154` |
| 5 | Cue sheet | **Partly** (#4). `capture-ios.sh:202-217` |
| 6 | Reel composer at 1.0x, `tpad`, per-join normalisation, window, PNG32 frame, `shortest=1`, `-respect-parentheses` | Met. `compose-reel.sh:128-159`, `:183-221`, `geometry.sh` |
| 7 | Self-checks | Met as listed, with deviations 1 and 2. **The caption/end-card limit is missing** (#1) |
| 8 | Guards proven by mutation | Met (recorded in the PR; I didn't re-run the mutants) |
| 9 | No app-specific values; fonts from contract or env; prerequisites up front | Met |
| 10 | bash 3.2, `bash <path>`, `BASH_SOURCE` | Met. Empty-array expansion is guarded (`compose-reel.sh:211`); no `((i++))`, `mapfile` or `local x=$(…)` |
| 11 | Never writes into the skill directory | Met |
| 12 | No binaries | Met |
| 13 | SKILL.md degrades per component; opencode limitation | Met, but step 1 is wrong (#5). `SKILL.md:316-324` |
| 14 | Install and `go test` validation | Claimed in the PR, not re-run by me. `installer.go:62-68` copies nested dirs at 0644, consistent with the claim |

**Deviation 1 is right.** #75 says the end card starts at Σ beats and is "≤ 3.0s including its transition", measured from the first frame of the transition (`SKILL.md:126-127`, `:141`). So the render is Σ beats + `duration_s`, and the ticket's "− its fade" would double-count. The code starts the xfade at the end of the footage (`compose-reel.sh:217`). On green3, the footage frames up to 7.3 match the baseline, the crossfade runs 7.3–7.8, and the render is 10.300000s. The only way a caption overlaps the transition is #1. Nominal-vs-frame rounding stays within one frame: the worst case I measured was 5.3ms, on drift3.

### What I ran

Nothing touched a Simulator or a device. Scratch fixtures and harness are outside the repo.
- **Baseline:** green3 re-run gives 10.300000s = plan.
- **Composer fixtures:**
  - **5 segments from two takes** (one of them 60 fps), 3 fades, millisecond cut points: 13.600000s vs a 13.593000s plan.
  - **6 cut segments** at millisecond points: 6.100000s vs 6.096000s.
  - **3 segments, 2 fades**, millisecond points: 10.066667s vs 10.072000s.
  - **0 captions, no badges**, and a take file longer than its logged length (pad 0.000): 10.3s.
  - **8 captions**, including `%` and `@` in the text: renders.
  - **A launch-day-only badge**, with and without `--launch-day`.
  - **The caption-overlap case** (#1) and **`--locale e.`** (#9).
- **Parity:** the alias, duplicate-key and no-cut fixtures, run through SKILL.md's Ruby check (extracted from `SKILL.md:203-310`), `contract.sh` under ruby, and under python3 with ruby hidden from PATH. The yq path was reviewed by reading only. The example campaign validates in `all` mode.
- **`capture-ios.sh` against fake `xcrun`/`maestro` on PATH:**
  - **Happy path:** every simctl call names the UDID, every Maestro call has `--device`, cue sheet and still are written, and the trap clears the status bar, restores the appearance and uninstalls.
  - **Failure cases:** Maestro exit 1 (#3), the skipped command (#4), two booted Simulators with the same name (exit 69, no UDID-addressed call), and the unfinalised recording (#13).
  - **Ctrl-C to the process group** mid-Maestro, and **SIGTERM to the script alone** mid-hold: both exit 130 with the full restore and no recorder left running.
- **Seeder:** the UTF-16 boundary, MD5 file names, and the refusal to overwrite existing storage.
- **Hygiene:** grepped the diff for `/Users/`, UDID patterns, store URLs and consumer names. Only the fictional `com.example.tallybird` appears.
- **Docs:** SKILL.md is 492 lines. Script `--help` headers match the entry-point table and the outputs list in `references/capture-and-compose.md`, apart from #2 and #5.

alexandrocuma:
Review findings addressed in `27ab9e7`.

**Blocking**
- **1. Captions into the end card:** `contract.sh` now refuses a caption with `end_s` past Σ `beats[].target_s`, `start_s` < 0, or `start_s` ≥ `end_s`, and `compose-reel.sh` runs that check before any render. The reviewer's 9.5s fixture exits 65 with "past the end-card start at 7.3s". green3 still renders at 10.300000s = plan.
- **2. `--launch-day` overwrote the reel:** it now writes `reel-<locale>.launch-day.mp4` (and matching check frames). The postable `reel-<locale>.mp4` is byte-identical before and after a launch-day run, and the summary names the do-not-post-before date.

**Should fix**
- **3. Stale take:** a take's old `.mov`, cue sheet and failure logs are removed before recording, and the cue sheet is moved into place only on success. With a Maestro stub exiting 1, only the failure log remains.
- **4. Skipped Maestro commands:** a command with no timestamp becomes a cue with `t_s: null`, its status kept. Real Maestro 2.10.0 still stamped both a skipped `when:` command and a failed `optional: true` command. The null case is guarded but was only seen with a stub.
- **5. Phase 7 validation order:** step 1 is `validate … capture`, step 2 `validate … all`. `all` before capture explains that the cut comes from the cue sheets.
- **6. Parser parity:** anchors, aliases and duplicate keys are refused by ruby, python3 (strict SafeLoader) and yq (explicit detection), and by SKILL.md's embedded check, with matching paths and lines.
- **7. Silent uninstall:** an installed app is refused (65) unless `--replace-installed`, and the trap only undoes what the run changed. On the Simulator the app and status bar were untouched after the refusal.
- **8. Safety Rule 5 on iOS:**
  - The host clipboard is saved, cleared and restored on every exit, including SIGTERM. `--keep-clipboard` opts out, and non-text clipboard content gets a warning that restore is text-only.
  - The Simulator pasteboard is emptied before each take.
  - simctl has no notification switch, so notifications are documented rather than automated.

**Nits:** all taken:
- literal `--locale`
- locale, `outputs.dir` and `bundle_id` format checks
- readable reasons for an unreadable recording or failed still
- explicit `run_step` status checks
- integer render timeout
- still ids documented as slugs

**Re-verified**
- **Controls:** green3, single segment and 1080x2220 each equal their plan.
- **Mutants:** M1–M9 unchanged (M1/M1a/M2–M8 red, M1b/M9 green).
- **Parser parity:** agrees across all four checkers.
- **Build and install:** `bash -n`; SKILL.md 497 lines; hygiene greps empty; `go test ./... -race` green; dry run lists 8 skills; temp-HOME install with deployed `--help` exit 0.
- **Real Simulator:** end to end at 7.400000s = plan, then restored (app not installed, no overrides, light).
- **Orchestrator spot check** on the pushed commit, asserting messages and not only exit codes: green3 = 10.300000s, the 9.5s caption red, the alias red in both `contract.sh` and SKILL.md's check, the example green in both, literal `--locale` red, hygiene clean.

Issue #114, opened 2026-09-16
Kimi hooks: registry block, adapter, config.toml hooks block


alexandrocuma:
Part of epic #110 — **sub-ticket 4**, depends on #111 and #107 (target-generic hook registry).

## Acceptance Criteria
- [ ] Each hook in `hooks/registry.json` gains a `kimi` target block (per #107's target-generic shape) stating event, matcher, command and whether it is enabled for Kimi
- [ ] An adapter lets the Claude Code guard scripts run under Kimi (maps Read/Write/Edit `path` to `file_path`; turns `ask` into a block with the reason) — guards block Read/Write/Edit of secrets despite Kimi's argument names
- [ ] In a Kimi session, reading `.env`, a destructive shell command, and a write containing a secret are each blocked with the guard's reason
- [ ] format-on-save and graphify-grep-nudge are not installed for Kimi, and the output says why
- [ ] devexp hooks are written as a marker-delimited `[[hooks]]` block in `$KIMI_CODE_HOME/config.toml`; user settings and comments preserved; re-install replaces only that block; the result passes `kimi doctor config`
- [ ] `hooks.disabled` / wizard selection apply to Kimi; `--dry-run` writes nothing
- [ ] Fail-open behaviour handled per the grooming decision (documented or mitigated in the adapter)
- [ ] Tests written and passing (adapter tests mirroring guard tests; TOML block writer tests)

## Non-Goals
- Kimi plugin-system install; Windows hook shells

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
# Grooming plan: #114 Kimi hooks (registry block, adapter, config.toml hooks block)

## Ticket Health Report: #114

**Ticket**: Kimi hooks: registry block, adapter, config.toml hooks block
**Type**: New feature, security-adjacent (safety guards on a new CLI)
**Groomed**: 2026-09-16
**Validated against**: commit `ff7246905a2662f7e9a6a3ccdda82d7e8d927f4f` (`feat/release-targets`).
- **Re-validated at `8ec0d09dcb77969ce3849097f5fa8a04a18e28cd`.** #107 merged into `feat/release-targets` during grooming (`Merge #116`). Every `hooks/registry.json`, `cli/internal/hooks/*` and hook-docs line below is quoted at `8ec0d09`.
[...]

Issue #108, opened 2026-09-16
Go CLI installs the opencode hook plugin


alexandrocuma:
Part of epic #106 — **sub-ticket B**, depends on #107. Groomed plan: https://github.com/alexandrocuma/devexp-toolkit/issues/106#issuecomment-5702068041 (steps 4–6, 8).

## Problem
The Go CLI's opencode install (`cli/cmd/install_opencode.go`) never deploys the hook plugin, so opencode users get no guards while the installer reports success. opencode auto-loads every file directly in `~/.config/opencode/plugins/` (non-recursive `{plugin,plugins}/*.{ts,js}`) and calls every export, so modules must live in a subdirectory behind a single entry file, copied (not referenced) so binary installs survive cache re-extraction.

## Acceptance Criteria
- [ ] After installing for opencode, `~/.config/opencode/plugins/` contains one devexp entry file and a `devexp/` subdirectory with only the selected modules, `utils.js`, `package.json` and the selection file — and a new opencode session blocks reading a `.env` file
- [ ] opencode starts without plugin load errors; each hook runs exactly once per tool call
- [ ] A hook listed in `hooks.disabled` or deselected in the wizard is not installed for opencode (same resolution as the Claude Code path)
- [ ] With every hook disabled, no plugin is installed and the output says why
- [ ] Hook test files (`*.test.js`) are never installed; the installer never writes `plugins/package.json`
- [ ] `--dry-run` prints every file that would be written and writes nothing
- [ ] Re-running install is idempotent; a hook removed or disabled since the last install is removed on re-install
- [ ] Files left by pre-`46ca772` flat installs are removed only when they match the known legacy set and content signature; a legacy `config.json` `plugin` entry is removed only on exact match; same-named non-matching files are kept with a warning
- [ ] Non-devexp files in `plugins/` are never touched
- [ ] Installed plugin files are tracked in the opencode manifest and survive `--agents-only` / `--skills-only` runs
- [ ] Install output lists the hooks installed for opencode; the "feature subset" warning no longer implies hooks are missing
- [ ] Docs flipped (known gap closed): `overview.md`, `CLAUDE.md`, `install.md`, `README.md`, `hooks/README.md`, `reference/hooks.md`, `hook-authoring-guide.md`, `workflows.md`, `conventions.md`, `testing.md`; CHANGELOG `### Fixed` entry telling clone users to `rm bin/devexp && ./install.sh`
[...]

Issue #106, opened 2026-09-16
Install the opencode hook plugin from the Go CLI


alexandrocuma:
> **Epic.** Groomed and split into sub-tickets — deliver in order:
>
> - [ ] #107 — opencode plugin: data-driven entry, explicit registry mapping, parity fixes (M)
> - [ ] #108 — Go CLI installs the opencode hook plugin (M) — depends on #107
> - [ ] #109 — uninstall.sh: remove the opencode hook plugin and fix top-level `local` crash (S) — depends on #108
>
> **Grooming corrections** (plan: https://github.com/alexandrocuma/devexp-toolkit/issues/106#issuecomment-5702068041):
> - opencode's `plugins/` scan is non-recursive (`{plugin,plugins}/*.{ts,js}`) → layout is one entry file + `devexp/` subdirectory, copied (not referenced)
> - `hooks.disabled` / wizard selection resolved at install time (same resolver as Claude Code), written to a selection file the entry reads
> - Legacy flat-install cleanup is in scope (known file set + content signature; exact-match config entry)
> - The config `plugin` array also accepts paths, not only npm packages
> - `uninstall.sh` crashes earlier than stated (L216 as well as L331) — fix moved into scope (#109)
> - lint/format/test-on-save never ran in opencode (`file.edited` isn't a plugin hook key)
>
> **Decisions:** D3 — opencode secret-guard message changed to match Claude Code (AC1 kept). D4 — entry adapts `event` → `file.edited`, enabling the three advisory hooks in opencode (format-on-save will rewrite files on edit).
>
> **Estimate:** XL overall → decomposed into M + M + S.

---

## Problem

opencode users get none of devexp's safety guards (secret reads, destructive commands, large-file overwrites) — yet the installer reports success. The Go CLI's opencode path (`cli/cmd/install_opencode.go`) installs MCPs, agents and skills but never deploys the hook plugin; the step was lost when the Go CLI replaced the bash installer in `46ca772`. Recorded as a known gap in `docs/architecture/overview.md#known-gaps`.

[...]

PR #116, opened 2026-09-16
fix(hooks): data-driven opencode plugin with a target-generic hook registry (#107)

### LINKED ISSUE #107: opencode plugin: data-driven entry, explicit registry mapping, parity fixes
Part of epic #106 — **sub-ticket A** (no dependencies). Groomed plan: https://github.com/alexandrocuma/devexp-toolkit/issues/106#issuecomment-5702068041 (steps 1–3).

## Problem
The opencode plugin entry (`hooks/opencode/devexp-plugin.js`) statically imports and composes all 10 modules, so it cannot honour hook selection, and one broken module disables every guard (fails open). The registry has no hook → JS module/export mapping (the Go `Hook` struct ignores the `opencode` block). The opencode secret-guard message differs from Claude Code's, and lint/format/test-on-save never run in opencode because `'file.edited'` is not a plugin hook key — file events only reach plugins via the `event` hook.

## Acceptance Criteria
- [ ] Every hook in `hooks/registry.json` declares its mapping **per install target** in a target-generic shape (existing `claude_code` block, an `opencode` block with module file, export name, fail-closed and enabled) so a future target (Kimi Code CLI — see follow-up epic) adds a sibling block without changing the Go registry type's structure; the Go registry type parses it and existing Claude Code install behaviour is unchanged
- [ ] The plugin entry exports exactly one function and composes only the modules listed in an installed selection file (`devexp/hooks.json`)
- [ ] A module that fails to import or initialise does not disable the other hooks
- [ ] If a fail-closed guard (secret-guard, secret-in-write-guard, dangerous-cmd-guard) fails to load, its tool calls are blocked with an internal-error message rather than allowed
- [ ] Blocking a `.env` read in opencode shows the same message text as the Claude Code guard (`Blocked access to "…"`)
- [ ] Editing a source file in opencode triggers lint-on-save, format-on-save and test-on-save (event → `file.edited` adapter), each respecting selection
- [ ] graphify-* hooks stay available for opencode (explicit opencode enablement), still self-gating on `graphify-out/graph.json`
- [ ] Tests written and passing: new `hooks/opencode/devexp-plugin.test.js` (composition from selection, failure isolation, fail-closed stubs, event adaptation); existing opencode and Claude Code hook suites

## Non-Goals
- Adding a Kimi Code CLI target (separate epic) — only the registry shape must accommodate it
- Installing the plugin (#108) or uninstalling it (#109)
- Changing any hook's detection logic beyond the message text

## Release Targets
- `cli` — ships in the next tagged release (assets embedded at build)
- `toolkit-clone` — on `git pull`; no effect for opencode users until #108 ships

alexandrocuma:
Closes #107 · Part of epic #106 · **Stacked on #105** (base `feat/release-targets`; retarget to `main` once #105 merges)

## Summary

Makes the opencode hook plugin data-driven and safe to compose, with a target-generic registry — the foundation #108 (installer) and #114 (Kimi hooks) build on.

- **Target-generic registry** — each hook in `hooks/registry.json` has sibling per-target blocks; `opencode` now declares `module`, `export`, `fail_closed`, `enabled`. Go `Hook{Name, Description, Enabled, Targets map[string]TargetSpec}` with `Target(id)` / `EnabledFor(id)`; a new target is a new block, not a new type (tested with an arbitrary `example_target`).
- **Data-driven entry** — `hooks/opencode/devexp-plugin.js` exports one function and composes only the modules listed in `devexp/hooks.json` (contract consumed by #108). Each module loads in isolation; a fail-closed guard that fails to load becomes a blocking stub; missing/invalid `hooks.json` blocks; module names with `/` or `..` are rejected.
- **Parity (D3)** — opencode secret-guard message now matches Claude Code: `Blocked access to "…"`.
- **file.edited adapter (D4)** — opencode delivers file events only via the `event` hook; the entry adapts `file.edited` so lint/format/test-on-save can run in opencode.
- **Bug fixed** — `test-on-save.js` threw `ReferenceError: path is not defined` on every source edit (never ran in opencode).
- **Docs** — "add a hook" instructions no longer point at `Promise.all`: CLAUDE.md, conventions, workflows, hooks reference, authoring guide, overview, hooks/README, docs-sync agent, testing.md; CHANGELOG.

Claude Code behaviour unchanged: `settings.json` produced by the base binary and the branch binary is byte-identical (also with base-registry vs branch-registry, roots normalised).

## Behaviour notes for reviewers
- Composition order is now registry order — when two guards match the same call, which block message appears can differ; whether it blocks cannot.
- No effect for opencode users until #108 installs the plugin.

## Out-of-scope findings
None open — the three found during delivery (test-on-save import, docs-sync wording, testing.md counts) were in scope and fixed in this PR.

## Test plan
- [x] Go `./scripts/stage-assets.sh && go test ./... -race -cover` — all pass; `internal/hooks` 89.4% → 91.6%; `go vet`/`gofmt` clean
- [x] Claude Code hook tests — 25 / 10 / 37 passed
- [x] opencode hook tests — dangerous-cmd-guard 25, **devexp-plugin 83 (new)**, secret-guard 38 passed
- [x] `uninstall.test.sh` — 8 passed
- [x] Claude Code `settings.json` byte-identity check (base vs branch)
- [x] Mutation checks: removing the fail-closed stub, path guard, event filter, `{ file }` payload mapping, the D3 message, or the test-on-save fix each fails the new test
- [ ] Real opencode session — requires #108 (installer)


hooks/opencode/devexp-plugin.js
```diff
@@ -1,72 +1,117 @@
...
+      const file = entry?.module;
+      if (typeof file !== 'string' || !file || file.includes('/') || file.includes('..')) {
+        throw new Error(`module ${JSON.stringify(file)} is not a bare file name inside devexp/`);
+      }
+      const mod = await import(new URL(file, dir).href);
+      const factory = mod[entry.export];
+      if (typeof factory !== 'function') throw new Error(`export ${entry.export} is not a function`);
+      loaded.push({ name, hooks: (await factory(ctx)) ?? {} });
+    } catch (err) {
+      if (entry?.failClosed) {
```
alexandrocuma: **Should-fix: fail-closed depends on one exact key in `hooks.json`. If the key is missing or spelled differently, a security guard that fails to load fails open.**

Whether a guard fails closed is decided only by `entry.failClosed` from the installed selection. The registry spells the field `fail_closed` and `hooks.json` spells it `failClosed`. #108 has to translate between them, so a snake_case slip in #108 is an easy mistake to make, and nothing would catch it.

Reproduced in a scratch tree:
- `hooks.json` = `[{"name":"secret-guard","module":"secret-guard.js","export":"secretGuard","fail_closed":true}]`
- `secret-guard.js` removed

Result: one `console.error` line (`[devexp secret-guard] failed to load; skipped`), then `tool.execute.before({tool:'read'},{args:{filePath:'/p/.env'}})` **resolves**, and the `.env` read is allowed. An entry that is `null`, not an object, or missing `name` also takes the fail-open branch.

This is the exact failure #107 exists to prevent, moved from "one broken import" to "one wrong key". A console line is easy to miss in opencode.

**Suggested fix (either works, both is better):**
- Make the entry defensive, not purely data-driven. Keep a small built-in set, `const FAIL_CLOSED = new Set(['secret-guard','secret-in-write-guard','dangerous-cmd-guard'])`, and treat `entry.failClosed === true || entry.fail_closed === true || FAIL_CLOSED.has(entry.name)` as fail-closed. Also treat an entry that isn't a valid object as a blocking stub, not a skip.
- Add a test case: a fail-closed guard's import fails with the key as `fail_closed`, or with no key but a known guard name. Expect a block. Also pin the `hooks.json` key spelling in a contract test that #108's writer must satisfy.

Question: is `hooks.json` = `[]` (no guards at all, silently) intended to mean "user deselected everything"? If so, it's worth one sentence in the contract docs, so an installer bug that writes an empty selection isn't mistaken for a valid state.

hooks/opencode/devexp-plugin.js
```diff
@@ -1,72 +1,117 @@
...
+      },
+    };
+  }
+
+  const loaded = [];
+  for (const entry of entries) {
+    const name = entry?.name ?? '(unnamed)';
+    try {
+      const file = entry?.module;
+      if (typeof file !== 'string' || !file || file.includes('/') || file.includes('..')) {
```
alexandrocuma: **Nit (defense in depth, and the code doesn't match its stated guarantee): the "bare file name inside `devexp/`" check can be bypassed with a URL scheme or an encoded dot segment.**

The check only rejects `/` and `..`, but the value goes through `new URL(file, dir)`, which also accepts:
- `"module": "node:process", "export": "cwd"` resolves to the `node:process` builtin. It imports, and the entry calls `cwd(ctx)` **with no error logged**. I reproduced this: the plugin loads and the entry is silently "composed". Any builtin export (`node:child_process`, `node:fs`, …) can be called this way with `ctx` as its first argument.
- `"module": "%2e%2e"` resolves to the parent `plugins/` directory. Here the only thing that stops it is that importing a directory fails.

The practical risk is low. Anyone who can write `hooks.json` can also write the modules next to it. But the header comment, the authoring guide and the PR description all promise that only bare file names inside `devexp/` are loaded.

**Suggested fix:** use an allowlist, not a denylist, and check the resolved URL:
```js
if (typeof file !== 'string' || !/^[A-Za-z0-9][A-Za-z0-9._-]*\.js$/.test(file)) throw …;
const url = new URL(file, dir);
if (url.protocol !== 'file:' || !url.href.startsWith(dir.href)) throw …;
```
Then add `node:process` and `%2e%2e` cases to test 7.

hooks/opencode/devexp-plugin.js
```diff
@@ -1,72 +1,117 @@
...
+    // Adapt opencode's bus events to the modules' file.edited handlers.
+    // opencode dispatches `event` with `void`, so nothing here may reject.
+    event: async (input) => {
+      const event = input?.event;
+      if (event?.type !== 'file.edited') return;
+      const payload = { file: event.properties?.file };
+      for (const m of loaded) {
+        if (!m.hooks['file.edited']) continue;
+        try {
+          await m.hooks['file.edited'](payload);
```
alexandrocuma: **Should-fix before #108 ships (or file a follow-up issue): the adapter newly runs synchronous child processes on opencode's event loop on every edit.**

Until now, `lint-on-save`, `format-on-save` and `test-on-save` never ran in opencode. This adapter turns them on.
- They use `execFileSync` and `spawnSync`, and `which` also runs `execFileSync`. Timeouts are 10s, 15s and 20s, so a worst case of about 45s per edit.
- Their handler bodies have no `await` before the spawn. The first handler therefore runs synchronously inside opencode's `events.listen(... Effect.sync(() => void hook.event(...)))` callback. The others run on the next microtasks.

For the whole run, the opencode server process is frozen:
- no streaming
- no abort/cancel handling
- other sessions stall

On a Go file, every edit runs `go vet ./pkg`, `gofmt -w` and `go test ./pkg` back to back. The agent waiting is roughly what Claude Code's PostToolUse hooks do, but those run in a separate process and don't freeze the host.

Also, `format-on-save` rewrites the file after opencode's own `format.file()` and after the edit tool has computed `contentNew`/`diff` (`tool/edit.ts:155-171`). The diff and LSP view the tool reports can then differ from what's on disk.

The module internals are a non-goal for #107, so this isn't blocking here. But this PR is what activates the behaviour, and #108 makes it user-visible.

**Suggested fix:** track it before #108.
- Switch the on-save helpers to async `execFile`/`spawn` with the same timeouts.
- At minimum, have the adapter start each handler with `setImmediate`/`queueMicrotask`, so the publish path isn't blocked.
- Consider whether `format-on-save` should be off by default for opencode, since opencode already formats on edit.

alexandrocuma (commented):
## PR review: #116 (diff against `feat/release-targets` @ ff72469)

**Verdict: no blockers.** There are 2 should-fix findings and 1 nit, all posted inline.

**Acceptance criteria for #107: all met.**
- Every registry entry has a per-target block.
- The Go `Hook.Targets` map parses an arbitrary `example_target` block.
- The Claude Code install path only changes field access (`installer.go:160` and `:280`). The `!hook.Enabled` check and the empty-event/script skip are untouched.
- The entry has a single function export and composes from `hooks.json`.
- A module that fails to load is isolated, and fail-closed stubs block.
- `Blocked access to` now matches `claude-code/secret-guard.sh:116`.
- `event` is adapted to `file.edited` with `{ file }`.
- The three graphify hooks have `opencode.enabled: true`.

**Verified locally at 2e75300:**
- `go test ./...` passes.
- `devexp-plugin.test.js` passes 83/83, `secret-guard.test.js` 38, `dangerous-cmd-guard.test.js` 25.
- Against the pinned opencode source (`anomalyco/opencode@2e018f70`):
  - The plugin glob is `{plugin,plugins}/*.{ts,js}`, which is not recursive, so the `devexp/` modules are never auto-loaded as plugins.
  - `getLegacyPlugins` throws on any export that isn't a function. The single export is correct.
  - `event` is dispatched with `void` from `Effect.sync`. The adapter can't reject, which is correct.
  - `file.edited` data is `{ file }`, published by `edit.ts:115,159`.

**Checked, no finding:**
- **Runtime error in an advisory module's `tool.execute.before`.** It still blocks the call, but that's unchanged from the base: the same sequential loop where the first throw wins. The entry can't tell a bug from an intentional block, because `large-file-guard` and `graphify-read-guard` are advisory yet block on purpose. Failing closed on runtime errors is the safer posture, so this is acceptable.
- **Ordering.** Registry order only changes which block message wins. No current module mutates `output.args`, and the graphify handlers that write state still run in the same relative order.
- **Import failures, init failures and missing exports.** With `failClosed`, each of these installs a blocking stub. Tests 4, 5 and 7 exercise them for real.
- **Missing or invalid `hooks.json`.** It blocks, and test 6 covers it.
- **Callers of `hooks.Hook`.** Nothing marshals it. `cli/cmd/registry.go` reads only `Name`/`Enabled`, and `uninstall.sh` reads only `claude_code.script`.
- **Registry JSON shape.** Consistent across all 10 hooks.
- **Docs.** They match the code.

hooks/opencode/devexp-plugin.js
alexandrocuma: Fixed in 0667807.

- An entry is now fail-closed if it has `failClosed: true`, **or** `fail_closed: true`, **or** a security-guard name. The three names (`secret-guard`, `secret-in-write-guard`, `dangerous-cmd-guard`) are hard-coded in `SECURITY_GUARDS`, and a comment explains why that duplicates the registry: defense in depth, so a dropped or misspelled flag can't make a guard fail open.
- An entry that is `null`, not an object, or has no non-empty string `name` now installs a blocking stub (`… devexp/hooks.json entry <i> is not a hook entry …`). It is never skipped.
- On your question: `hooks.json` = `[]` is **misconfigured**, not "user deselected everything". With every hook disabled, #108 installs no plugin at all, so `[]` can only come from an installer bug. It now blocks the same way an unreadable file does. The contract in `hook-authoring-guide.md` says so.
- Tests (`devexp-plugin.test.js`, case 11):
  - your reproduction: secret-guard.js removed, key spelled `fail_closed`, and the `.env` read now blocks
  - `fail_closed` on a guard whose module is missing
  - each guard name with no flag, or with `failClosed: false`
  - null, number, string, array, nameless, empty-name and numeric-name entries
  - `[]` (case 6)
  - a contract test that parses the guide's `hooks.json` example, asserts its keys are exactly `name`/`module`/`export`/`failClosed`, loads it, and shows that `module`/`export` under other key names are not read
- Mutation checks, each removed on its own:
  - dropping the name set fails 3 checks
  - dropping the `fail_closed` alias fails 1
  - skipping invalid entries fails 7
  - allowing `[]` fails 2

hooks/opencode/devexp-plugin.js
alexandrocuma: Fixed in 0667807.

- `module` now has to match the allowlist `^[A-Za-z0-9][A-Za-z0-9._-]*\.js$`.
- The resolved URL must also have protocol `file:` and start with `dir.href` before anything is imported.
- Test 7 now covers `node:process`, `%2e%2e`, `%2e%2e/evil.js`, `%2e%2e\evil.js`, `.%2e\evil.js`, `probe-count.js?x` and `file:///etc/evil.js`. Each must be refused as not a bare file name, `evil.js` must never be imported, and a `node:` builtin marked `failClosed` must block.
- I found one more bypass while writing the tests. `%2e%2e\evil.js` contains neither `/` nor `..`, but WHATWG URL treats `\` as `/` in `file:` URLs, so the old check really imported `plugins/evil.js`.
- Mutation check: against the previous entry, the `node:process`, `%2e%2e`, both backslash cases and the `?x` case fail, and so does "no sneaky module escapes to evil.js". With only the old denylist swapped back in, the new URL check still stops the escapes, but the refusal-message checks fail (5 failures).

hooks/opencode/devexp-plugin.js
alexandrocuma: Fixed in 0667807.

- **Non-blocking dispatch.** `event` no longer awaits the handlers. It queues them and returns at once. The queue waits one `setImmediate`, then runs one edit at a time, calling the modules in selection order, and logs each module's errors.
- **No synchronous spawns.** `utils.js` has a new `runCommand` (async `spawn`: stdin closed, utf-8 output or `output: 'ignore'`, SIGTERM on timeout, never rejects). `which`, `runLinter`, `format-on-save` and `test-on-save` all use it. Tool choice, cwd, messages and the 10s/15s/20s timeouts are unchanged. `format-on-save` stays silent as before.
- **format-on-save** stays enabled for opencode (owner decision D4). The module header, `hook-authoring-guide.md` and `reference/hooks.md` now say its rewrite lands after the edit tool computed its diff, so the reported diff can differ from the file on disk.
- **Tests.** Case 12 uses a project whose local `eslint` sleeps 1s:
  - `event` returns in under 200ms while the linter still runs
  - a 50ms timer fires on time
  - the lint output is still printed, and dispatch continues after it

  Case 13 covers `runCommand`: a 200ms timeout kills `sleep 5` with SIGTERM, exit code/stdout/stderr and ENOENT behave as before, the timeout constants are 10000/15000/20000, and no on-save module imports a sync spawn.
- **Mutation checks.** Awaiting the handlers inside `event` fails case 12 (returned after ~1.2s). A synchronous `runLinter` fails the timer check (fired ~1.1s late) and the no-sync-spawn check. `runCommand` without `timeout` fails case 13 (took 5s).

Issue #117, opened 2026-09-16
Stale agent/skill removal trusts manifest names without filtering


alexandrocuma:
Found during delivery of #108 (out of its scope).

## Problem
On re-install, stale agents and skills are removed by name from the previous install manifest (`removeStale(target, manifest.Stale(old, new), os.Remove, dryRun)` in `cli/cmd/install_claude.go` and `cli/cmd/install_opencode.go`). The names are joined onto the target directory without validation, so a corrupted or hand-edited manifest entry (e.g. one containing path separators or `..`) can cause a file outside the agents/skills directory to be deleted.

#108 added filtering for opencode plugin paths (`hooks.OwnedStalePlugins`); agents and skills have no equivalent.

## Acceptance Criteria
- [ ] A manifest agent/skill entry that is not a bare name matching the installed-file shape (no separators, no `..`, expected extension/dir form) is never removed; a warning names it
- [ ] Removal is confined to the target agents/skills directory for both Claude Code and opencode
- [ ] Existing stale-removal behaviour for valid names is unchanged
- [ ] Tests written and passing, including a tampered-manifest case (mutation-checked)

## Non-Goals
- Changing the manifest format

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
## Ticket Health Report — #117

**Ticket**: Stale agent/skill removal trusts manifest names without filtering
**Type**: Bug fix (security-adjacent: file deletion driven by an on-disk manifest)
**Groomed**: 2026-09-16
**Validated against**: commit `d359b6e4ba7e46c402272d050d9088f16517ed3a` (`main`, v0.8.0)

---

### Confirmed Claims
- Stale agents and skills are removed by manifest name with no validation → `cli/cmd/install_claude.go:63` (agents, `os.Remove`), `:82` (skills, `os.RemoveAll`); `cli/cmd/install_opencode.go:72` (agents, `os.Remove`), `:90-95` (commands, `<entry>.md`, `os.Remove`).
- Names are joined straight onto the target dir → `removeStale` in `cli/cmd/backup.go:69-86` does `filepath.Join(dir, name)` and calls `removeFn`; `manifest.Stale` (`cli/internal/manifest/manifest.go:69`) is a pure set difference.
- #108 filters opencode plugin paths; agents/skills have no equivalent → confirmed.

### Incorrect or Misleading Claims
[...]

Issue #109, opened 2026-09-16
uninstall.sh: remove the opencode hook plugin and fix top-level local crash


alexandrocuma:
Part of epic #106 — **sub-ticket C**, depends on #108 (merged). Re-groomed plan: https://github.com/alexandrocuma/devexp-toolkit/issues/109#issuecomment-5704076997

## Problem
`uninstall.sh` crashes for every opencode uninstall: `local` is used outside a function under `set -euo pipefail`, so it exits 1 after removing agents and skills — leaving the #108 hook plugin behind. Its plugin removal also hardcodes 6 legacy flat file names and doesn't know the #108 layout (`plugins/devexp.js` + `plugins/devexp/`).

**Approach (owner decisions O1–O4):** removal rules live in Go, shared with #108's install code — `uninstall.sh` delegates to a hidden `devexp uninstall --target opencode [--dry-run] [--yes]`. With no usable binary, uninstall warns, leaves the plugin untouched, prints a rebuild hint and exits 0.

## Acceptance Criteria
- [ ] `./uninstall.sh` for opencode completes with exit 0 (no top-level `local`; the opencode MCP block tolerates an unreadable, malformed or oddly shaped `config.json`)
- [ ] An opencode install is detected even when no agents are installed (plugin-only)
- [ ] Uninstall removes only devexp-owned plugin paths — recorded in the manifest or recognised on disk by the same rules install uses — never deletes through a symlinked `plugins/`, refuses a symlinked `devexp/` or a dangling `plugins/` link, and removes `devexp.js` and `devexp/` together or not at all
- [ ] Legacy flat-install files are removed only by the same rule as install (closed name set, `/**` header, name header, regular file, no symlinks); the legacy `config.json` `plugin` entry only on exact match with the rest of the file byte-preserved; a symlinked `config.json` is left untouched; nothing (config included) is changed when the root checks fail
- [ ] Non-devexp files in `~/.config/opencode/plugins/` and unrelated `config.json` content are left untouched
- [ ] After uninstall the manifest's `plugins` key lists only what had to stay (only when the manifest loaded cleanly; never on dry-run)
- [ ] With no usable `devexp` binary: warning, plugin untouched, rebuild hint, exit 0
- [ ] Tests written and passing: removal-rule cases (incl. install-vs-uninstall parity) in Go tests; script wiring (detection, binary lookup, step order, crash regressions, no top-level `local`) in `uninstall.test.sh`

## Non-Goals
[...]

PR #118, opened 2026-09-16
fix(cli): install the opencode hook plugin (#108)

### LINKED ISSUE #108: Go CLI installs the opencode hook plugin

alexandrocuma:
Closes #108 · Part of epic #106 · **Stacked on #105** (base `feat/release-targets`, which already contains #107)

## Summary

`devexp install` now installs the opencode hook plugin — opencode users get devexp's safety guards for the first time since the Go CLI replaced the bash installer (`46ca772`).

- **Layout** — `~/.config/opencode/plugins/devexp.js` (the #107 entry) + `plugins/devexp/` with only the selected modules, `utils.js`, `package.json`, and `hooks.json` (exact contract from #107: `name`/`module`/`export`/`failClosed`, registry order, never `[]`). opencode's plugin scan is non-recursive, so only the entry is auto-loaded. Files are copied, not referenced, so binary installs survive cache re-extraction.
- **Selection** — `hooks.disabled` and the wizard's hook choices apply to opencode through the same `resolveHookDisabled` + `EnabledFor(opencode)` as Claude Code. Every hook disabled → no plugin, reason printed.
- **Never installed** — `*.test.js`, `plugins/package.json`.
- **Legacy cleanup** — pre-`46ca772` flat-install files are removed only when name ∈ closed legacy set **and** content signature matches (shared fixtures in `cli/internal/hooks/testdata/legacy-opencode/` for #109); a legacy `config.json` `plugin` entry is removed only on exact match, by splicing the original bytes (formatting, key order and file mode preserved). Mismatches are kept with a warning.
- **Manifest** — new `plugins` list; stale plugin files removed only if `OwnedStalePlugins` confirms they're devexp-owned; preserved on `--agents-only` / `--skills-only`.
- **Refusals** — won't overwrite a non-devexp `plugins/devexp.js`, write through symlinks, or proceed when `devexp` is a file; validates every source/destination before writing.
- **Unreadable manifest** — no longer panics (Claude Code and opencode): warns, removes nothing stale that run, rewrites the manifest.
- **Docs** — opencode hooks known gap closed across install/README/hooks/overview/CLAUDE.md; CHANGELOG tells clone users to `rm bin/devexp && ./install.sh`.

## Release note
Must not be cut before #109 — `uninstall.sh` doesn't yet know this layout. The `v0.8.0` cut is deferred until the epic completes.

## Out-of-scope findings
- #117 — stale agent/skill removal trusts manifest names without filtering (plugins are now filtered)
- A legacy `config.json` entry written with a trailing-slash `$HOME` (`//.config`) isn't matched and is left in place (safe direction)

## Test plan
- [x] Go `go vet` + `go test ./... -race -cover` — 377 tests/subtests; `cmd` 35.4% → 65.0%, `manifest` 91.3%, `hooks` 91.4%
- [x] Cross-language: Go-installed tree loaded by node with opencode's rules → 1 plugin, 1 guard chain, Claude's exact `.env` message, no load errors
- [x] Contract test: writer output matches the documented `hooks.json` example exactly
- [x] Claude Code / opencode hook suites (25/10/37 · 25/130/38) and `uninstall.test.sh` (8)
- [x] Scripted temp-`HOME` e2e (stub opencode, legacy + foreign fixtures): dry-run writes nothing · install tree exact · legacy removed, user file kept · loader emulation · idempotent (identical bytes) · disable one · disable all · partial runs — 27 PASS
- [x] Mutation checks: 32/32 caught (deletion-safety rules, exact config match, byte-splice, refusals, contract keys, selection, manifest carry-forward, unreadable manifest)
- [ ] Manual: wizard deselection (TTY)
- [ ] Manual: real opencode — `opencode debug config` shows `plugins/devexp.js`, no "Failed to load plugin"; `.env` read blocked once; lint-on-save runs on a `.ts` edit


alexandrocuma (commented):
## PR Review: fix(cli): install the opencode hook plugin (#108)

**Recommendation: Request changes.** One blocking data-safety bug and four should-fix issues. All five are inside this diff, and each was reproduced against a binary built from `8f9084c` with a temp `HOME`, a stub `opencode` on `PATH` and `DEVEXP_DIR` pointing at a copy of the repo.

| # | Severity | Where | What |
|---|---|---|---|
| 1 | **blocking** | `opencode.go:335` / `:375`, `install_opencode.go:113-115` | Stale removal and prune follow a symlinked `plugins/devexp/`, deleting files outside `plugins/` |
| 2 | should-fix | `install_opencode.go:104` | Legacy cleanup runs before `InstallOpencode` validates, so a refusal leaves the user with no guards |
| 3 | should-fix | `opencode.go:308` | Writes aren't atomic. An interrupted `devexp.js` write fails open, and every later install refuses to overwrite it |
| 4 | should-fix | `backup.go:61` / `install_opencode.go:43` | An unreadable manifest drops plugin tracking for good. After that, "every hook is disabled" leaves all 10 guards running |
| 5 | should-fix (low likelihood) | `opencode.go:338` | A kept `devexp.js` loses its `hooks.json`, so every opencode tool call is blocked and re-running install doesn't fix it |

### What I checked that holds up
- **Tests:** `go vet` and `go test ./...` are green.
- **`--dry-run`:** with legacy fixtures and a CRLF `config.json`, a before/after snapshot of every file (content hash plus mode) is identical.
- **Idempotency:** a second run leaves every byte the same, manifest and `config.json` included.
- **`hooks.json` contract:** matches the entry and `hook-authoring-guide.md` (key names, registry order, never `[]`). Modules, then `hooks.json`, are written before the entry.
- **Manifest paths:** `OwnedStalePlugins` rejects `../`, nested paths, absolute paths, dotfiles and non-`.js` names.
- **Legacy file detection:** matches only on name set, `/**` header and regular file. Symlinked or same-named user files are kept.
- **`config.json` byte splice:** tested with CRLF, duplicate entries, escaped `\/`, unicode keys (`plu\u0067in`), nested `plugin` keys, non-string elements, trailing commas, comments, BOM, trailing documents, and `plugin` as a string or duplicated key. Each case is either correct or left untouched; the duplicate-key case returns a warning. A 60s fuzz run (31M execs) that includes the `DeepEqual` guard found no panic and no write that failed the guard.
- **Claude Code path:** unchanged apart from the unreadable-manifest fix. Its manifest keeps its shape (`omitempty`).

Nice work on the byte splice: verifying the result against the semantic edit is a solid design. The refusal set and the contract test are also well done.

cli/internal/hooks/opencode.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+	return nil
+}
+
+// OwnedStalePlugins filters manifest-recorded stale plugin paths down to the
+// ones devexp may delete: devexp.js (only while it still is a devexp entry) or
+// a file directly in devexp/ with a name devexp installs. Anything else — a
+// hand-edited manifest with ../, a nested path, a replaced entry — is kept and
+// reported, never removed.
+func OwnedStalePlugins(pluginsDir string, stale []string) []string {
```
alexandrocuma: **[blocking] Stale removal deletes through a symlinked `plugins/devexp/`, outside `plugins/`.**

`InstallOpencode` refuses a symlinked `devexp/` (`checkPluginDest`). But with every hook disabled it returns `nil, nil` before any check. `doInstallOpencode` then passes the whole old manifest through `OwnedStalePlugins`, which checks names only, and on to `removeStale(os.Remove)`. So every `devexp/<name>` resolves through the link.

A realistic trigger is a maintainer live-linking the plugin to a working checkout. Repro:
```sh
devexp install --reinstall-mcps                       # normal install, manifest has 14 plugin paths
cp -R repo/hooks checkout-hooks
rm -rf $P/devexp && ln -s $T/checkout-hooks/opencode $P/devexp
# set hooks.disabled to every hook
devexp install --reinstall-mcps
ls checkout-hooks/opencode   # 16 files -> 4 left: only *.test.js and devexp-plugin.js survive
ls -la $P                    # the devexp symlink itself is gone too (PruneOpencodeDir)
```
Twelve source files in the checkout are deleted. This contradicts the CHANGELOG line "never deletes a manifest path outside `devexp.js` and `devexp/`".

**Suggested fix:** do the removal in `hooks`, with the same parent check `checkPluginDest` uses. Before removing any `devexp/...` path, `Lstat` `plugins/devexp` and skip every removal under it with a warning unless it is a real directory (`fi.IsDir()`, not a symlink). The same check belongs in `PruneOpencodeDir` (see the comment on line 375). Add a regression case: a symlinked `devexp/` with every hook disabled, plus a manifest that lists files.

cli/internal/hooks/opencode.go
```diff
@@ -0,0 +1,615 @@
...
+	return err == nil && isDevexpEntry(content)
+}
+
+// PruneOpencodeDir removes pluginsDir/devexp once it is empty. os.Remove
+// refuses a non-empty directory, so anything left in it survives.
+func PruneOpencodeDir(pluginsDir string, dryRun bool) {
+	if dryRun {
+		return
+	}
+	os.Remove(filepath.Join(pluginsDir, opencodeDir)) //nolint:errcheck
```
alexandrocuma: **[blocking, same root cause as the comment on line 335]** `os.Remove` on a symlink deletes the link even when the target directory isn't empty, so "os.Remove refuses a non-empty directory" doesn't protect a symlinked `devexp`. The repro in the line 335 comment leaves `plugins/` empty. `Lstat` first, and only remove a real directory.

cli/cmd/install_opencode.go
```diff
@@ -94,6 +95,28 @@ func doInstallOpencode(opts *installOpts) error {
 		removeStale(skillsTarget, staleSkillFiles, os.Remove, opts.dryRun)
 	}
 
+	if !opts.agentsOnly && !opts.skillsOnly {
+		registry, err := hooks.LoadRegistry(filepath.Join(opts.repoDir, "hooks", "registry.json"))
+		if err != nil {
+			ui.Warn(fmt.Sprintf("hooks registry: %v", err))
+		} else {
+			ui.Info(fmt.Sprintf("Installing hooks (opencode plugin → %s)...", p.plugins))
+			if err := hooks.CleanLegacyOpencode(p.plugins, configPath, opts.dryRun); err != nil {
```
alexandrocuma: **[should-fix] Legacy cleanup runs before `InstallOpencode` has validated anything, so a refusal leaves the user with no guards.**

`CleanLegacyOpencode` deletes the working legacy plugin files and the `config.json` entry. Only then does `InstallOpencode` run `checkPluginDest`. When it refuses, the command returns an error with the old plugin already gone and the new one never written. Repro:
```sh
cp testdata/legacy-opencode/* $P/                      # working pre-v0.1.0 install
echo 'export const Mine = async () => ({})' > $P/devexp.js   # an unrelated user plugin with that name
printf '{"plugin":["%s/devexp-plugin.js"]}' $P > config.json
devexp install --reinstall-mcps
#  - devexp-plugin.js (legacy flat install) ... - package.json ... - plugin entry ...
#  Error: .../plugins/devexp.js exists and is not a devexp plugin entry
ls $P    # devexp.js   <- only the user's file; no guards at all
```
A missing source or an I/O error during validation has the same effect.

**Suggested fix:** run `CleanLegacyOpencode` only after `InstallOpencode` returns `nil`. The legacy files don't collide with the new layout, and opencode needs a restart to see either one. Alternatively, split `InstallOpencode` into validate and apply steps, and clean legacy files between them.

cli/internal/hooks/opencode.go
```diff
@@ -0,0 +1,615 @@
...
+		if !isDevexpEntry(content) {
+			return fmt.Errorf("%s exists and is not a devexp plugin entry — move it aside and re-run", p)
+		}
+	}
+	return nil
+}
+
+// writeIfChanged writes content to pluginsDir/dest unless it already holds
+// exactly those bytes, reporting added or updated files.
+func writeIfChanged(pluginsDir, dest string, content []byte) error {
```
alexandrocuma, line 394: **[should-fix] Plugin writes aren't atomic. An interrupted `devexp.js` write fails open and leaves the install stuck.**

`os.WriteFile` truncates, then writes. A crash, `SIGKILL` or `ENOSPC` midway leaves a partial file:
- **Partial `hooks.json`:** the entry blocks every tool call. That fails closed, which is fine.
- **Partial `devexp.js`:** opencode logs "Failed to load plugin" and runs with **no guards**. Every later `devexp install` then refuses the file, because `isDevexpEntry` needs the first two header lines. Repro:
```sh
devexp install --reinstall-mcps
head -c 2 $P/devexp.js > x && mv x $P/devexp.js   # what an interrupted write leaves
devexp install --reinstall-mcps
# Error: .../plugins/devexp.js exists and is not a devexp plugin entry — move it aside and re-run
```
It also leaves a check-then-use gap: the path is checked by `Lstat` in `checkPluginDest` and then opened by `os.WriteFile`, which follows a symlink created between the two.

**Suggested fix:** write to `os.CreateTemp(filepath.Dir(p), "."+base+".tmp-*")`, then `Sync`, `Chmod(0644)` and `os.Rename` it over `p`. The rename is atomic, and it replaces a symlink instead of following it, which closes both problems. The same applies to the `config.json` rewrite in `removeLegacyConfigEntry` (line 491). That file is user-owned, so a truncated write there loses the user's opencode config.

cli/cmd/backup.go
```diff
@@ -53,6 +54,18 @@ func backupExistingDirs(dir, backupDir string, dryRun bool) {
 	}
 }
 
+// loadOldManifest reads the previous run's manifest. An unreadable or corrupt
+// manifest is reported and treated as empty: the install carries on, nothing
+// counts as stale on this run (an empty old list marks nothing for removal),
+// and the manifest is rewritten at the end as usual.
+func loadOldManifest(path string) *manifest.Manifest {
```
alexandrocuma, line 61: **[should-fix] An unreadable manifest loses the plugin list for good, not just for "this run".**

After a corrupt manifest, `old.Plugins` is empty, so the rewritten manifest has no `plugins` key. Every later run then has nothing to diff against. The same happens through a partial run: `--agents-only` carries the empty `Plugins` forward (`install_opencode.go:43`). Result: once every hook is disabled, the installer prints "nothing to install", yet the whole plugin stays installed and active. This breaks the acceptance criteria "With every hook disabled, no plugin is installed" and "a hook removed or disabled since the last install is removed on re-install". Repro:
```sh
devexp install --reinstall-mcps
echo '{"agents": "oops"' > ~/.config/opencode/.devexp-manifest.json
# set hooks.disabled to every hook
devexp install --reinstall-mcps   # warns; [skip] opencode plugin — every hook is disabled — nothing to install
devexp install --reinstall-mcps   # same output again
ls $P                             # devexp  devexp.js   <- all 10 guards still load, on every future run
```

**Suggested fix:** don't rely only on the manifest to find plugin files. When the manifest was unreadable, or when nothing is selected, also treat these paths as owned candidates when they exist on disk: `devexp.js` (while `isDevexpEntry`), plus `devexp/{hooks.json,package.json,utils.js}` and each registry `opencode.module` basename (inside a real `devexp/` directory). Pass them through the same ownership filter. `loadOldManifest` could return a `trusted bool` so the opencode path knows when to fall back to this.

cli/internal/hooks/opencode.go
```diff
@@ -0,0 +1,615 @@
...
+
+// OwnedStalePlugins filters manifest-recorded stale plugin paths down to the
+// ones devexp may delete: devexp.js (only while it still is a devexp entry) or
+// a file directly in devexp/ with a name devexp installs. Anything else — a
+// hand-edited manifest with ../, a nested path, a replaced entry — is kept and
+// reported, never removed.
+func OwnedStalePlugins(pluginsDir string, stale []string) []string {
+	var owned []string
+	for _, rel := range stale {
+		if isOwnedPluginPath(rel) && (rel != opencodeEntry || entryStillDevexp(pluginsDir)) {
```
alexandrocuma: **[should-fix, low likelihood] Keeping `devexp.js` while removing the rest of the plugin blocks every opencode tool call, and re-running install can't fix it.**

When the entry fails `entryStillDevexp`, for example because it is a symlink to a real devexp entry in someone's dotfiles, only `devexp.js` is kept. `devexp/hooks.json` and the modules are still removed. opencode then loads an entry with no `hooks.json`, so every tool call is blocked with "re-run devexp install". Re-running with every hook disabled does nothing, so the remedy in that message doesn't work. Repro:
```sh
devexp install --reinstall-mcps
mv $P/devexp.js dotfiles/devexp.js && ln -s $T/dotfiles/devexp.js $P/devexp.js
# set hooks.disabled to every hook
devexp install --reinstall-mcps   # "devexp.js left untouched"; devexp/ removed
ls $P                             # devexp.js -> dotfiles/devexp.js ; no devexp/  => entry blocks all calls
```

**Suggested fix:** make removal all-or-nothing when the entry is in the stale set. If `devexp.js` can't be removed, keep `devexp/hooks.json` too, and warn that the plugin was left installed.

cli/internal/hooks/opencode.go
alexandrocuma: Fixed in d5c0838.

`InstallOpencode` now runs `checkPluginRoots` before it writes or removes anything, and that includes the every-hook-disabled path: the early return is gone. The check `Lstat`s `plugins/` and `plugins/devexp/` and refuses either one if it is a symlink or not a directory.

Stale removal has moved from `cmd` into `hooks.removePluginFiles`, which runs the same check again right before removing and only removes regular files. `ownedOnDisk` only looks inside a real `devexp/`. `CleanLegacyOpencode` also refuses a symlinked `plugins/`, because a checkout of `hooks/opencode` carries exactly the legacy names and headers.

Your repro, before and after (`review-repros.sh` on `h.sh`):
- **Before (8f9084c):** checkout files 16 → 4, and the symlink was removed.
- **After:** checkout files 16 → 16, the link is kept, and the install stops with `…/plugins/devexp is a symlink — devexp writes and removes opencode plugin files only in a real directory; replace the link with a directory and re-run`.
- A symlinked `plugins/` behaves the same way: nothing behind it changes.

Regression tests:
- `TestInstallOpencode_SymlinkedRoots`: `devexp/` and `plugins/`, hooks enabled and all disabled, dry-run and real.
- `TestDoInstallOpencode_SymlinkedDevexpAllDisabled`: your scenario end to end.
- `TestRemovePluginFiles_RechecksRoots`, `TestOwnedOnDisk_SymlinkedDevexpDir`, and a symlinked-`plugins/` case in `TestCleanLegacyOpencode_Config`.

Mutation-checked: removing any of these checks turns a test red.

The CHANGELOG now says "never deletes anything outside `devexp.js` and a real `devexp/` directory", and that holds.

cli/internal/hooks/opencode.go
alexandrocuma: Fixed in d5c0838. `pruneOpencodeDir` now `Lstat`s `devexp` and calls `os.Remove` only when it is a real directory, so a symlink is never removed. `TestPruneOpencodeDir` has an "a symlink is never removed" case, and dropping the `Lstat` check turns it red. Beyond that, a symlinked `devexp/` never reaches the prune: `InstallOpencode` refuses it first (see the reply on line 335).

cli/cmd/install_opencode.go
alexandrocuma: Fixed in d5c0838. `doInstallOpencode` now calls `CleanLegacyOpencode` only after `InstallOpencode` returns without an error, and `InstallOpencode` validates everything before it writes anything: sources, destinations, the `devexp.js` ownership and the symlink checks.

Your repro, before and after:
- **Before (8f9084c):** the legacy files were deleted, then the install errored, leaving the user with no guards.
- **After:** `Error: …/plugins/devexp.js exists and is not a devexp plugin entry`. Every legacy file in `plugins/` is byte-identical and the `config.json` `plugin` array is unchanged. The only change to `config.json` is the `mcp` key the MCP step writes before hooks.

Test: `TestDoInstallOpencode_RefusalKeepsLegacyInstall` (legacy fixtures, a user `devexp.js` and the legacy config entry; all bytes are compared). Moving the cleanup back before the install turns it red.

cli/internal/hooks/opencode.go
alexandrocuma: Fixed in d5c0838.

**Atomic writes.** Plugin files and the `config.json` rewrite now go through `writeFileAtomic`: `os.CreateTemp` in the same directory (`.<name>.tmp-*`), then `Write`, `Sync`, `Chmod` (0644 for plugin files, the original mode for `config.json`), `Close` and `Rename`. The temp file is removed on any failure. `Rename` replaces whatever is at the path, including a symlink created after the `Lstat` check, instead of writing through it. A symlinked `config.json` is `Lstat`-refused before anything happens: it is left untouched, with a warning to remove the entry by hand, so a dotfiles link is never followed or replaced.

**Damaged entry.** A `devexp.js` that the manifest records is now replaced even if its header is gone. One that isn't recorded and has no entry header is still refused.

Your repro, before and after:
- **Before:** `Error: …/plugins/devexp.js exists and is not a devexp plugin entry`.
- **After:** `~ devexp.js — updated`, and the file equals the source again. No `.tmp-` files are left behind.

Tests:
- `TestWriteFileAtomic`: mode, no leftover temp files on success or on a failed rename, a symlink replaced rather than written through.
- `TestWrites_NeverInPlace`: in a read-only directory the write fails and the old bytes stay, for both plugin files and `config.json`.
- `TestInstallOpencode_EntryOwnership`: recorded and truncated → repaired; unrecorded → refused.
- A symlinked `config.json` case.

Each guard is mutation-checked.

cli/cmd/backup.go
alexandrocuma: Fixed in d5c0838.

The files that may be removed are now the manifest's `plugins` list **plus** the devexp plugin files recognised on disk (`ownedOnDisk`):
- `devexp.js`, only while it carries the entry header;
- inside a real, non-symlink `devexp/` only: `hooks.json`, `package.json`, `utils.js` and every registry `opencode.module` basename that exists as a regular file.

These go through the same path filter as manifest entries. That covers both a corrupt manifest and the `--agents-only` carry-forward of an empty list. Agents and skills are unchanged: an unreadable manifest still removes none of them, and the warning now says so.

Your repro, before and after:
- **Before:** `devexp` and `devexp.js` were still installed after two all-disabled runs.
- **After:** both are gone after the first run, and foreign files stay.
- Corrupt manifest followed by a normal install rewrites `plugins` with all 14 entries.

Tests: `TestInstallOpencode_LostManifest` and `TestDoInstallOpencode_LostManifest` (your repro, and a normal install restoring the list), plus `TestStalePlugins` cases for on-disk discovery (foreign `devexp/mine.txt`, `unknown.js` and `my-plugin.js` are never picked up, nor is a `devexp.js` without the header). Dropping the on-disk candidates, or the header check on them, turns tests red.

cli/internal/hooks/opencode.go
alexandrocuma: Fixed in d5c0838. Removing the plugin is now all-or-nothing.

When every hook is disabled and `devexp.js` must stay (a symlink or another non-regular file), nothing is removed. The kept files stay recorded in the manifest, and the output says:

`opencode plugin left installed and its hooks remain active: …/devexp.js is a symlink, so devexp.js and devexp/ are both kept (removing only devexp/ would block every tool call). Move devexp.js aside and re-run to remove the plugin.`

The same applies if deleting the entry itself fails: `removePluginFiles` removes the entry first and stops, keeping the rest.

Your dotfiles repro, before and after:
- **Before:** `devexp/` was removed and only the symlinked entry was left, which blocks every call.
- **After:** the symlink, `devexp/hooks.json` and the modules are all kept, and the warning is printed.

Tests: `TestInstallOpencode_KeptEntryKeepsPlugin` (the symlinked entry) and `TestInstallOpencode_EntryRemovalFailureKeepsPlugin` (a read-only `plugins/` whose `devexp.js` can't be unlinked while `devexp/` stays writable). Both guards are mutation-checked.

Issue #110, opened 2026-09-16
Kimi Code CLI as a third install target


alexandrocuma:
> **Epic.** Deliver in dependency order. Refined with `/refine`; grooming follows per sub-ticket.
>
> - [ ] #111 — Kimi target plumbing: paths, detection, N-target selection (S)
> - [ ] #112 — Kimi MCP installer (S) — depends on #111
> - [ ] #113 — Kimi agents and skills installer (M) — depends on #111
> - [ ] #114 — Kimi hooks: registry block, adapter, config.toml block (M) — depends on #111, #107
> - [ ] #115 — Kimi uninstall, feature-subset warning, docs (S) — depends on #112, #113, #114

## Problem
Teams on **Kimi Code CLI** (moonshotai/kimi-code — binary `kimi`, config in `$KIMI_CODE_HOME` or `~/.kimi-code`) can't install devexp: the installer only knows Claude Code and opencode, so Kimi users get no agents, skills, MCPs or safety guards.

## Research summary (Kimi Code CLI `0.43.1`, commit `bd06178`)
- MCPs: `$KIMI_CODE_HOME/mcp.json` `{"mcpServers":{…}}` (stdio `command/args/env`, `url/headers`, `transport:"sse"`); no `kimi mcp add` CLI; one invalid entry fails the whole file.
- Agents: `$KIMI_CODE_HOME/agents/*.md`; `description` required, `name` kebab-case; mostly a tool-name remap (`WebFetch`→`FetchURL`, Task* todo tools→`TodoList`); custom agents are genuinely spawnable (`subagents: "*"` for orchestrators).
- Skills: `$KIMI_CODE_HOME/skills/<n>/SKILL.md`, dir copy; invoked `/skill:<n>` or `/<n>`.
- Hooks: `[[hooks]]` in `$KIMI_CODE_HOME/config.toml` (`event`, `matcher`, `command`, `timeout`; strict schema). Same stdin shape as Claude (`tool_name`, `tool_input`) and exit 2 blocks — but Read/Write/Edit use `path` not `file_path`, `ask` is treated as allow, context/PostToolUse feedback is ignored, and spawn errors/timeouts fail open. Reuse via an adapter.
- Kimi reads `AGENTS.md`, never `CLAUDE.md`. Legacy Python kimi-cli also ships a `kimi` binary (uses `~/.kimi/`).

## User Stories & Acceptance Criteria

**1. Detect and select Kimi**
[...]

PR #120, opened 2026-09-16
fix(hooks): harden hook input handling (#119)

### LINKED ISSUE #119: security: harden large-file-guard hook input handling
## Problem
A Claude Code hook script handles tool-supplied input unsafely, which can lead to unintended command execution under crafted input. Affects released versions (installed and enabled by default for Claude Code).

Details are intentionally omitted from this public issue.

## Acceptance Criteria
- [ ] Tool-supplied values are never interpreted as code by the hook
- [ ] The hook's existing behaviour for normal files is unchanged
- [ ] Regression tests cover hostile input shapes
- [ ] Other hook scripts audited for the same class of issue; any found are fixed in the same change
- [ ] Tests written and passing

## Non-Goals
- Changing what the hook decides for legitimate inputs

## Release Targets
- `cli`, `toolkit-clone` — patch release from `main`, independent of the #106/#110 epics

alexandrocuma:
Closes #119 · Security · Base `main` — ships as a patch independent of the #106 / #110 epics

## Summary
A Claude Code hook script could let tool-supplied values be interpreted as code under crafted input. Those values are now passed to the hook's interpreter strictly as data, using the same pattern the on-save hooks already use. The hook's decisions for legitimate inputs are unchanged (old vs new output compared byte-for-byte across normal inputs, including size boundaries).

All Claude Code and opencode hooks were audited for the same class of issue; no other instance was found.

Also adds a "treat tool input as data, never as code" rule to the hook authoring guide, and a CHANGELOG `### Security` entry.

## Test plan
- [x] New `hooks/claude-code/large-file-guard.test.sh` — 32 checks: baseline decisions plus hostile input shapes; asserts nothing is executed and decisions match a benign input of the same size (picked up by existing CI glob)
- [x] Mutation: reverting the fix fails the new test
- [x] Go `go test ./... -race`; Claude Code hook tests (25/10/32/37); opencode hook tests (25/38); `uninstall.test.sh` (8)


alexandrocuma (commented):
## PR Review: fix(hooks): harden hook input handling (#119)

**Recommendation**: Approve with minor comments. Nothing blocks merge, and the one comment below is about text outside this diff.

**Summary**: The fix passes tool-supplied values to the interpreter as arguments, under a quoted heredoc delimiter, so they never become part of the program text. The output JSON is still built by the serializer, so a hostile value cannot add or change fields. I checked it independently and agree the fix is complete for this class, the behaviour for legitimate inputs is unchanged, and the audit claim holds.

---

### What I verified

- **Fix completeness.** I traced every step from the tool envelope to the output: the extractor, the file-exists and line-count checks, the arithmetic comparison, the interpreter call and the JSON output. After the fix, no tool-supplied value reaches program text. Every shell expansion is quoted. The line count comes from `wc`, not from tool input. Nothing inside the heredoc is expanded because the delimiter is quoted.
- **Hostile inputs, run independently** (not the PR's own test), with a sentinel file. I ran each case as an absolute and as a relative path: quote and string-literal breakouts, attempts to break out of the JSON, arguments that look like options, format-string characters, Unicode line separators and control and escape sequences, NUL, lone surrogates, non-string `file_path` types, a non-object `tool_input` or envelope, FIFOs and directories. On this branch nothing executed. Every output was valid JSON with exactly `hookSpecificOutput` and `systemMessage`, the decision was always `ask`, and the path came back intact. Against `main`, the string-literal class does execute, which confirms the test targets the right thing.
- **Behaviour parity.** I compared `main` and this branch byte-for-byte on stdout, stderr and exit code: 146 cases, 0 differences. The cases covered benign names (spaces, Unicode, brackets, tabs), line counts of 0/1/499/500/501/600/100000 as absolute and relative paths, a missing file, an empty or absent `tool_input`, a directory, malformed JSON and empty stdin. The padded `wc -l` count on macOS is kept verbatim.
- **Advisory failure semantics.** Malformed input still exits 0 with the `internal error` message, and `fail-closed.test.sh` passes (10/10).
- **Audit claim.** I spot-checked every `hooks/claude-code/*.sh` and `hooks/opencode/*.js` file and found no other place where tool data reaches program text:
  - The other shell hooks either use static interpreter source with data on stdin or argv, or use quoted expansions with constant `grep` patterns.
  - The on-save hooks already pass argv under a quoted heredoc, and their `subprocess.run` calls use list arguments.
  - The opencode modules use only `execFileSync`/`spawnSync` with argument arrays and never `shell: true`, `eval` or `Function`.
- **Tests.**
  - I reverted the hook to `main` in a scratch copy and the new test fails (7 checks, including the non-execution probe).
  - Suites on this branch: claude-code 25/10/32/37, opencode 25/38, `uninstall.test.sh` 8, all passing.
  - `stage-assets.sh` followed by `go test ./... -race` passes (ran in a scratch copy).
- **Public text.** The commit message, CHANGELOG, PR body and issue all stay at class level.

### 🔵 Minor (outside this diff)

**The docs examples just below the new rule still show the old pattern.**
`docs/development/hook-authoring-guide.md`, *Soft block (ask)* example and *Boilerplate*

The new "Treat tool input as data" rule is good. But the soft-block example directly below it still uses the `python3 -c "..."` form, and the boilerplate's extractor still ends in `2>/dev/null || echo ""`. That fallback fails open silently, which contradicts the contract in `fail-closed.test.sh`. The example text is static today, but authors tend to copy it and add the file path to the message. Consider rewriting the soft-block example in the argv/quoted-heredoc form that `large-file-guard.sh` now uses, and making the boilerplate's error branch match the hooks. This could be a follow-up; it does not need to block this PR.

### Optional nit

`large-file-guard.sh` L26: the second interpreter call has no `|| { internal error …; exit 0; }` fallback, so a failure there would exit non-zero without the `internal error` prefix. I could not find an input that makes it fail: its arguments are always valid strings and the output is ASCII-escaped. This matches `main`, so it is only a consistency note.

### ✅ What's done well

- The fix uses the pattern the on-save hooks already use, instead of adding a new escaping helper.
- The test is well built. Before the hostile cases it checks that a plain name really asks above 500 lines and stays silent at or under 500, so "same decision as a plain name" means something. It also checks the round-tripped message exactly, which catches mangled output as well as execution.
- Advisory semantics and the macOS `wc` padding are both preserved, and the parity claim in the PR description is accurate.

Issue #124, opened 2026-09-16
Config and manifest writes: atomic and symlink-safe everywhere


alexandrocuma:
Found during #108/#109 (out of their scope); also noted by Kimi grooming (F4).

## Problem
Several writers still use truncate-and-write and may follow symlinks: `manifest.Save` (`os.WriteFile`), the opencode MCP merge on install (`cli/internal/mcp/opencode.go`), and the opencode MCP removal in `uninstall.sh` (python `json.dump`, which also reformats `config.json` and writes through a symlinked `config.json`). #108 introduced `writeFileAtomic` for plugin files and the legacy config splice; the rest don't use it.

## Acceptance Criteria
- [ ] Manifest and opencode `config.json` writes use temp file + fsync + rename, preserving file mode
- [ ] A symlinked `config.json` is never written through; the user gets a warning (consistent with #108)
- [ ] opencode MCP removal preserves unrelated formatting/bytes (move it to the Go uninstall command added in #109, or byte-splice)
- [ ] Parse errors are reported, never silently overwritten
- [ ] Tests incl. interrupted-write simulation and symlink cases

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
Also in scope here (found delivering #137, PR #147): `InstallClaude` writes `~/.claude/settings.json` with a plain `os.WriteFile`, and the `uninstall.sh` python step writes it non-atomically too (unlike its opencode `config.json` step). A plain rename would replace a symlinked `settings.json` with a regular file, so the fix needs to resolve the link target first.

alexandrocuma:
# Execution plan — #124 (atomic, symlink-safe config and manifest writes) + #150 (hook registration gaps)

Delivered together in the v0.9.2 batch, branch `fix/124`.

## Ticket Health Report — #124

**Ticket**: Config and manifest writes: atomic and symlink-safe everywhere
**Type**: Bug fix / hardening
**Groomed**: 2026-09-16
**Validated against**: commit `10ad67c39707e1f8734ba44e8f7f7e64f0db824d` (`main`)

### Confirmed claims
- `manifest.Save` truncates and writes: `os.WriteFile` → `cli/internal/manifest/manifest.go:63`.
- The opencode MCP merge on install truncates and writes → `cli/internal/mcp/opencode.go:120`.
[...]

Issue #123, opened 2026-09-16
uninstall.sh: skills handling is wrong for opencode-only uninstalls


alexandrocuma:
Found during #109 (out of its scope).

## Problem
On an opencode-only uninstall, `uninstall.sh` prints "Skills will be kept (still in use by other installed CLI)" and its skills logic targets `~/.claude/skills`. opencode skills are installed as `~/.config/opencode/commands/<name>.md`, which uninstall never removes.

## Acceptance Criteria
- [ ] opencode uninstall removes the devexp commands it installed (manifest-driven, same ownership discipline as agents) and leaves user commands untouched
- [ ] Claude Code skills are only considered when Claude Code is being uninstalled; messages name the right CLI and path
- [ ] opencode manifest is removed when nothing devexp-owned remains
- [ ] Tests in `uninstall.test.sh`

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
Related, from the PR #156 review: `uninstall.sh` matches agents and skills by repo name with `test -f`. On a case-insensitive filesystem (macOS default), it would remove a user's own file whose name differs only in case from a repo agent (e.g. `Dev-Agent.md` for `dev-agent.md`). Install already guards against this by comparing exact on-disk names (#128). Uninstall should do the same, ideally driven by the manifest.

Issue #115, opened 2026-09-16
Kimi uninstall, feature-subset warning and docs


alexandrocuma:
Part of epic #110 — **sub-ticket 5**, depends on #112, #113, #114.

## Acceptance Criteria
- [ ] `./uninstall.sh` detects Kimi installs and offers it among the targets
- [ ] Uninstall removes only what devexp installed: manifest-listed agents and skills, registry MCP keys from `mcp.json`, the devexp hooks block from `config.toml`, the manifest; user content untouched; exits 0
- [ ] Install output shows a Kimi feature-subset warning (no agent memory, colours, per-agent model; hooks fail open on errors; no CLAUDE.md)
- [ ] Docs updated: `overview.md` (install flow for Kimi), `install.md`, `README.md`, `reference/hooks.md`, `reference/mcps.md`, `agent-authoring-guide.md` (Kimi compatibility), `setup.md`; CHANGELOG
- [ ] `uninstall.test.sh` covers the Kimi target; tests passing

## Non-Goals
- Legacy kimi-cli cleanup

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
# Grooming plan: #115 Kimi uninstall, feature-subset warning, docs

## Ticket Health Report: #115

**Ticket**: Kimi uninstall, feature-subset warning and docs
**Type**: New feature (uninstall path) + docs
**Groomed**: 2026-09-16
**Validated against**: commit `ff7246905a2662f7e9a6a3ccdda82d7e8d927f4f` (`feat/release-targets`).
- **Re-validated at `8ec0d09dcb77969ce3849097f5fa8a04a18e28cd`.** #107 merged into `feat/release-targets` during grooming. Docs line numbers below are quoted at `8ec0d09`.
- `uninstall.sh` and `uninstall.test.sh` are identical in both commits.
- #108 and #109 will shift the `uninstall.sh`, `install.md`, `README.md` and `overview.md` lines again. Every edit is therefore also named by section, so it can be re-anchored.

**Kimi source checked**: `moonshotai/kimi-code` @ `bd06178` (0.43.1). `core/` = `packages/agent-core-v2/src/`.
**Epic**: #110, sub-ticket 5 (last).
**Dependencies**:
- **#112, #113, #114**: the manifest fields `mcps` (hashes), `agents`, `skills` and `hook_files`; `hooks.WriteKimiConfig` / `stripDevexpHooks`; `safeNames`.
[...]

PR #125, opened 2026-09-16
fix(uninstall): remove the opencode hook plugin via shared Go code; stop crashing (#109)

### LINKED ISSUE #109: uninstall.sh: remove the opencode hook plugin and fix top-level local crash

alexandrocuma:
Closes #109 · Part of epic #106 · **Stacked on #105** (base `feat/release-targets`, which contains #107, #108 and v0.7.1)

## Summary

`uninstall.sh` no longer aborts on opencode, and it now removes the #108 hook plugin — using the **same Go removal code as install** (owner decision O1), so uninstall can never delete something install would refuse to.

- **Hidden `devexp uninstall --target opencode [--dry-run] [--yes]`** (O4, target map ready for Kimi's #115) calls new `hooks.UninstallOpencode`, which shares `removeWholeOpencodePlugin` with `InstallOpencode` plus the existing `CleanLegacyOpencode`. Parity is pinned by `TestUninstallOpencode_MatchesInstall` (uninstall vs install-with-every-hook-disabled on the same trees). No `repo.Resolve` call, so the asset cache `uninstall.sh` may run from is never deleted.
- **Same guarantees as #108**: only devexp-owned paths (manifest or recognised on disk); never removes through a symlinked `plugins/`; symlinked `devexp/` / dangling `plugins/` refused; entry and `devexp/` removed together or not at all; legacy rule and exact-match `config.json` byte splice; symlinked `config.json` untouched.
- **Manifest (O3)**: only the `plugins` key is rewritten to what had to stay; only if it loaded cleanly; never on `--dry-run`.
- **`uninstall.sh`**: top-level `local` crashes gone; opencode MCP block tolerates unreadable/malformed/odd `config.json`; plugin-only installs detected; plugin step runs before the MCP block; dry-run preview before the confirm prompt; `find_devexp_bin` probe captures help output (no `grep -q` under `pipefail`). No usable binary → warning, plugin untouched, rebuild hint, exit 0 (O2). `runRemove` passes `DEVEXP_BIN`.
- **Docs + CHANGELOG** `[Unreleased] ### Fixed`.

## Release note
Completes epic #106 (#107 → #108 → #109). The `v0.8.0` cut should include all three.

## Out-of-scope findings (filed)
- https://github.com/alexandrocuma/devexp-toolkit/issues/123
- https://github.com/alexandrocuma/devexp-toolkit/issues/124

## Test plan
- [x] Go `go vet` + `go test ./... -race -cover` — 479 tests/subtests; `hooks` 91.8%, `cmd` 67.6%
- [x] `uninstall.test.sh` — 59 passed (55 + SKIP without go on PATH)
- [x] Claude Code hook suites (25/10/11/34/12/37), opencode (25/130/38)
- [x] Temp-HOME real-binary round trip: symlinked `plugins/` (nothing removed, exit 0) · real dir + legacy + foreign (only foreign remains, config bytes preserved, manifest cleaned) · foreign `utils.js`/`package.json` kept · plugin-only detection with old binary (warn, exit 0) then new binary (removed) · run from embedded-asset cache
- [x] Mutation checks: 18/18 caught
- [ ] Manual: real opencode session after uninstall (no plugin load errors)


uninstall.sh
```diff
@@ -213,19 +249,28 @@ PYEOF
...
-    mcps = json.load(f)
 config_path = sys.argv[2]
-with open(config_path) as f:
-    config = json.load(f)
+try:
+    with open(sys.argv[1]) as f:
+        mcps = json.load(f)
+    with open(config_path) as f:
+        config = json.load(f)
+except (OSError, ValueError) as e:
```
alexandrocuma, line 268: **Should-fix: the MCP step still exits 1 when `config.json` is readable but not writable.** This breaks AC 1 ("completes with exit 0") and the new line in `docs/guides/install.md`.

The guard covers both reads, but the write at L280-283 (`open(config_path, 'w')`) is not covered. A `PermissionError` there gives a traceback, python exits 1, and `set -e` aborts the script. By then the agents and the plugin are already gone, and for choice [3] the Claude Code hook removal never runs.

**Repro** (temp HOME, binary built from 0843fb4):
1. Run `devexp install --reinstall-mcps`. `config.json` gets `mcp.context7`.
2. Make `config.json` read-only (mode 444).
3. Run `DEVEXP_BIN=<bin> bash uninstall.sh --yes </dev/null`.

Result: `PermissionError: [Errno 13] Permission denied: …/config.json`, `uninstall.sh exit=1`, and no "Uninstall complete".

**Fix:** wrap the save in `try/except OSError as e: print(f"  [warn] could not save {config_path}: {e}")`. Add a `{"mcp":{"context7":{}}}` case with a read-only file to the heredoc loop in `uninstall.test.sh`.

uninstall.sh
```diff
@@ -213,19 +249,28 @@ PYEOF
 fi
 
 if $REMOVE_OPENCODE && [[ -f "$REPO_DIR/mcps/registry.json" ]]; then
-    local config_path="$HOME/.config/opencode/config.json"
+    config_path="$HOME/.config/opencode/config.json"
```
alexandrocuma, line 252: **Should-fix: this block now runs for opencode, and it rewrites a symlinked `config.json` through the link, non-atomically.** Before this PR the `local` on this line aborted the script first, so this python has never executed for an opencode uninstall. Fixing the crash makes it live.

In the same run, the Go step leaves a symlinked `config.json` untouched on purpose ("rewriting would either follow the link or replace it … change someone's dotfiles setup", `removeLegacyConfigEntry`). This block then does exactly that:
- it follows the link;
- it reformats the whole file with `json.dump(indent=2)`;
- it truncates the file before writing, so an interrupted write leaves a truncated file.

**Repro** (temp HOME):
1. Install.
2. Move `config.json` into `dotfiles/` and symlink it back.
3. Run `uninstall.sh --yes`.

Result: the dotfiles copy's hash changes (`c4a37aa…` → `f8e35d8…`), and the output has `Saved: …/config.json`.

The plan put moving MCP removal to Go out of scope, and that still makes sense. The two guards that keep this consistent with the Go step are small, though:
```python
if os.path.islink(config_path):
    print(f"  [skip] {config_path} is a symlink — remove the devexp MCP servers from it by hand"); sys.exit(0)
```
For the save, write to a temp file in the same directory and `os.replace` it. At minimum, note the new reachability in the Known gaps in `docs/architecture/overview.md`.

cli/internal/hooks/opencode.go
```diff
@@ -298,17 +300,51 @@ func InstallOpencode(registry Registry, repoDir, pluginsDir string, disabled, re
...
+	return append(dests, kept...), nil
+}
+
+// removeWholeOpencodePlugin removes stale, the files of a plugin that is no
+// longer installed at all, once the roots have been checked. It is all-or-
+// nothing on the entry: if devexp.js has to stay, everything stays, because
+// an entry without its hooks.json blocks every opencode tool call. devexp/ is
+// pruned once nothing was kept. It returns the paths it kept, which the
+// manifest should go on recording. reason labels the dry-run lines.
+func removeWholeOpencodePlugin(pluginsDir string, stale []string, reason string, dryRun bool) []string {
```
alexandrocuma, line 315: **Should-fix: uninstall can leave a half-removed plugin that blocks every opencode tool call, and it prints no warning.**

`removeWholeOpencodePlugin` is all-or-nothing only when `devexp.js` is stale or when `entryKeepReason` fires (symlink, not a regular file, unreadable). A **regular** `devexp.js` that is neither recorded nor header-matching falls through:
- every owned `devexp/` file is removed, `hooks.json` included;
- the entry stays.

`devexp-plugin.js` then returns `blockAll(misconfigured('devexp/hooks.json unreadable'))`, and its advice ("re-run devexp install") is refused by `checkPluginDest`.

**Repro** (temp HOME, binary built from 0843fb4):
1. Run `devexp install --reinstall-mcps`.
2. Delete `.devexp-manifest.json`.
3. Prepend `// @ts-check` to `plugins/devexp.js`. An editor adding a BOM, or a future change to the header text, has the same effect.
4. Run `uninstall.sh --yes`.

Result: exit 0, and `plugins/` contains only `devexp.js`. Importing it with node and calling `tool.execute.before` gives `BLOCKED: [devexp] opencode plugin is misconfigured (devexp/hooks.json unreadable)`.

Why keeping everything is always safe here:
- Owned `devexp/` files next to an unrecognised `devexp.js` almost always mean devexp's own entry was edited.
- Install refuses to write next to a foreign entry, so a real foreign plugin can't have been installed alongside devexp.
- If the entry really is foreign, keeping `devexp/` costs only clutter. Removing it can lock the user out.

**Fix:** in this helper, treat "`devexp.js` exists but isn't in `stale`" as a keep reason, with its own warning: "…isn't recognised as devexp's, so devexp/ is kept with it; move devexp.js aside and re-run". Install and uninstall share the helper, so parity holds. Update the pinned case `an unrecorded foreign devexp.js is kept, devexp's own files go` (opencode_test.go:1706) to expect `keptAll`, and fix the `entryKeepReason` comment to match.

alexandrocuma (commented):
## PR review: #125 (uninstall: remove the opencode hook plugin through shared Go code)

**Recommendation: Request changes.** Nothing is blocking, but three should-fix findings are inside this diff.

The core design holds up well under attack.
- **Paths:** removal goes through #108's helpers, and every root-symlink, hostile-manifest and foreign-file case I tried behaved as the ACs say.
- **Refactor:** `removeWholeOpencodePlugin` leaves install unchanged (traced branch by branch).
- **Parity test:** it is real. A mutation that makes `UninstallOpencode` ignore `recorded` fails `TestUninstallOpencode_MatchesInstall`.

The problems are at the edges: the newly reachable python MCP step, one half-removed state inherited from the shared rules, and the manifest guard.

### Should-fix
1. **The opencode MCP step can still fail `uninstall.sh` with exit 1** when `config.json` is read-only. The try/except covers the reads but not the write. (uninstall.sh:263)
2. **The MCP step writes through a symlinked `config.json`, non-atomically.** The Go step deliberately leaves that same file alone. This code never ran for opencode before, because the `local` crash stopped the script first. (uninstall.sh:252)
3. **Uninstall can leave `devexp.js` without `devexp/`, and opencode then blocks every tool call.** This needs a lost manifest plus an entry whose header no longer matches. No warning is printed. (hooks/opencode.go:314)

### Nits
4. **A dangling manifest symlink passes the "manifest existed" guard,** so `manifest.Save` creates the link target. (cmd/uninstall.go:93)
5. **With `HOME` unset or empty, the hidden command works on `./.config/opencode/plugins` under the current directory.** (cmd/uninstall.go:80)
6. **Detecting a plugin-only install sends a Claude-plus-leftover-plugin user into the 1/2/3 menu.** Under `--yes </dev/null` that menu exits 1 and removes nothing. (uninstall.sh:47)

### Verified
- `go vet` and `go test ./...` are green.
- `uninstall.test.sh` gives 59/59, including with `PATH=/usr/bin:/bin`, where python is 3.9.6.
- **Temp-HOME runs with a binary built from 0843fb4:**
  - read-only config, symlinked config, dangling manifest link, and a lost manifest with an edited entry (findings 1-4);
  - `HOME` unset (finding 5), and the Claude-plus-plugin menu (finding 6);
  - running `uninstall.sh` from the embedded asset cache: an old binary on `PATH` warns and exits 0 with the plugin kept, then with `DEVEXP_BIN` the plugin is removed and the cache is byte-identical.
- **`find_devexp_bin`:** a binary built from `feat/release-targets` exits 1 on `uninstall --help` with no `supported: ` text, so it is correctly skipped. The probe output is captured, so there is no SIGPIPE under pipefail.
- **Dry run:** the command writes nothing, and the manifest is not saved.
- **Legacy step:** the byte splice runs before the python step.

### Done well
- Keeping one implementation for install and uninstall (O1), and pinning it with a differential test on every fixture.
- Never calling `repo.Resolve`, with a test that protects the cache.
- The stub harness in `uninstall.test.sh`: `env -i`, a temp `HOME`, a SIGPIPE-sized help output, and old-binary fall-through.

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,137 @@
...
+// legacy flat install left behind. Agents, commands and MCP servers are
+// uninstall.sh's job.
+//
+// Warnings (a kept entry, a symlinked plugins/) don't fail the run: what is
+// kept stays recorded in the manifest. An error means the plugin roots were
+// refused and nothing was removed, legacy files and config.json included.
+func doUninstallOpencode(home string, dryRun bool) error {
+	p := opencodeTargetPaths(home)
+
+	_, statErr := os.Lstat(p.manifest)
```
alexandrocuma: **Nit: a dangling manifest symlink creates a manifest,** which breaks O3's "never create a manifest" and the docs line "A manifest that is missing … is left as it is".

- `os.Lstat` succeeds on the dangling link, so `statErr == nil`.
- `manifest.Load` sees `IsNotExist` and returns an empty manifest with no error.
- When anything is kept, for example behind a symlinked `plugins/`, `manifest.Save` (`os.WriteFile`) follows the link and creates its target.

**Repro:** after an install:
1. Replace `plugins/` with a symlink to a copy.
2. Replace `.devexp-manifest.json` with a symlink to `dotfiles/manifest-target.json`, which doesn't exist.
3. Run `uninstall.sh --yes`.

Result: `dotfiles/manifest-target.json` is created with `"agents": null, "skills": null, "plugins": [14 paths]`.

**Fix:** use `os.Stat` here, so a dangling link counts as missing. Add the case next to "a missing manifest is not created even when files had to stay".

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,137 @@
...
+	sort.Strings(names)
+	return "supported: " + strings.Join(names, ", ")
+}
+
+func runUninstall(cmd *cobra.Command, args []string) error {
+	handler, ok := uninstallTargets[flagUninstallTarget]
+	if !ok {
+		return fmt.Errorf("unsupported target %q (%s)", flagUninstallTarget, supportedUninstallTargets())
+	}
+	return handler(os.Getenv("HOME"), flagUninstallDryRun)
```
alexandrocuma: **Nit: with `HOME` unset or empty, this deletes relative to the current directory.** `opencodeTargetPaths("")` yields `.config/opencode/plugins`.

**Repro:** in a directory that holds a dotfiles-style `.config/opencode/plugins/devexp.js` (real header) and `devexp/utils.js`, run `env -i PATH=/usr/bin:/bin devexp uninstall --target opencode`. Result: `Removing hooks (opencode plugin → .config/opencode/plugins)...` and both files are removed.

`uninstall.sh` can't hit this, because `set -u` stops it on `$HOME` first, and install has the same pattern. Still, this is a new deleting entry point, and the guard is cheap: refuse `home == ""` or a relative `home` with an error (or use `os.UserHomeDir()`).

uninstall.sh
```diff
@@ -35,14 +42,19 @@ for f in "$REPO_DIR/agents/"*.md; do
     [[ -f "$CLAUDE_AGENTS/$(basename "$f")"   ]] && HAS_CLAUDE_INSTALL=true
     [[ -f "$OPENCODE_AGENTS/$(basename "$f")" ]] && HAS_OPENCODE_INSTALL=true
 done
+# The hook plugin alone is an install too: every agent can be deselected.
+# devexp-plugin.js is the legacy flat entry.
+for p in "$OPENCODE_PLUGINS/devexp.js" "$OPENCODE_PLUGINS/devexp" "$OPENCODE_PLUGINS/devexp-plugin.js"; do
```
alexandrocuma, line 47: **Nit (pre-existing menu, newly widened): `--yes` in a script now fails for some Claude-only users.** Take a user with Claude Code agents plus a leftover opencode plugin, or a `plugins/devexp/` directory, and no opencode agents. This loop marks opencode as installed, so the script reaches the `read` menu at L72. That menu ignores `--yes`, and with stdin closed `read` fails and `set -e` exits.

**Repro:** a temp HOME with `~/.claude/agents/a.md` plus `plugins/devexp.js`, then `bash uninstall.sh --yes </dev/null`. Result: `exit=1`, and the Claude agent is not removed. Without the plugin file the same run exits 0.

#115 (K9) owns the menu and `--yes`, so a tracking note there is enough. Mention it in the PR if it isn't already covered.

uninstall.sh
alexandrocuma: Fixed in 0577051.

- When a devexp MCP has to be removed, the step now checks `os.access(config_path, os.W_OK)` first. A read-only file is left as it was, with `[warn] … is not writable, so it was left untouched — remove the devexp MCP servers from it by hand`.
- Any `OSError` during the save (for example a read-only directory, where no temp file can be created) is caught and warned about as well. Python exits 0 either way, so `set -e` no longer aborts the script.
- Your repro (`repro.sh` R1) now gives `uninstall.sh exit=0` and "Uninstall complete.", with no traceback.
- Tests in `uninstall.test.sh`:
  - At block level: a read-only file and a read-only directory each exit 0, leave `config.json` unchanged, print `[warn]` and leave no temp file.
  - A full choice-[3] run with a read-only `config.json` (file and directory variants) exits 0, still removes the Claude Code hook from `settings.json` and both CLIs' agents, and reaches "Uninstall complete."
- Mutation checks: dropping the writability check, or letting the save's `OSError` escape, each fail these tests.

uninstall.sh
alexandrocuma: Fixed in 0577051. Scope is limited to the symlink skip and the atomic replace; byte preservation stays in #124.

- `if os.path.islink(config_path)` skips the step with `[skip] … is a symlink, so it was left untouched — remove the devexp MCP servers from it by hand`, matching the Go step's rule.
- The save now works like this: `tempfile.mkstemp` in the same directory, `json.dump`, `fsync`, `chmod` to the old mode, then `os.replace`. The temp file is removed on any failure. It is Python 3.9 compatible; the tests also run with `/usr/bin/python3` 3.9.6.
- Your repro (R5): the dotfiles hash stays `c4a37aa…` before and after, the file is still a link, and there is no `Saved:` line.
- Tests (block level):
  - symlinked config: the link and its target bytes are unchanged, and the warning is printed;
  - atomic save: a hard link keeps the old bytes (so the file was replaced, not written in place), a 0640 mode is kept, and only `config.json` is left in the directory.
- Mutation checks: removing the `islink` guard, saving in place, or dropping the `chmod` each fail a test.
- `overview.md` Known gaps now notes that the step still reformats `config.json` (#124).

cli/internal/hooks/opencode.go
alexandrocuma: Fixed in 7e8f58c.

- `entryKeepReason(pluginsDir, stale)` now also returns "not recognised as devexp's (not recorded in the manifest, no devexp header)" for a regular `devexp.js` that isn't in `stale`. `removeWholeOpencodePlugin` then keeps everything and warns: "…so devexp.js and devexp/ are both kept (removing only devexp/ would block every tool call). Move devexp.js aside and re-run to remove the plugin."
- Install with every hook disabled goes through the same helper, so it changed identically, and `TestUninstallOpencode_MatchesInstall` still passes on every fixture.
- Tests:
  - The pinned case is now "an unrecorded foreign devexp.js keeps devexp/ with it" (expects everything kept).
  - A new "lost manifest + `// @ts-check`" fixture is in the parity table.
  - `TestWholePluginRemoval_EditedEntryStaysLoadable` runs your repro for both install (every hook disabled) and uninstall with the real registry. It checks the tree is unchanged and the warning is printed, then loads `devexp.js` with node and asserts a harmless read is **allowed**. Without the fix, that probe prints your `blocked: [devexp] opencode plugin is misconfigured (devexp/hooks.json unreadable)`.
- Your repro (R4) now leaves `devexp` and `devexp.js`, and the node probe prints "tool call allowed".
- Mutation check: disabling the new case fails `TestUninstallOpencode` and the node test.
- The `entryKeepReason` comment, `install.md` and the CHANGELOG are updated to match.

cli/cmd/uninstall.go
alexandrocuma: Fixed in b7250d6.

- The save now happens only when `.devexp-manifest.json` is a **regular file**, checked with `Lstat` and `IsRegular`. A symlink, dangling or not, is never written through. It is left untouched with a warning: "manifest … is a symlink, so it was left untouched — its plugins list no longer matches what is on disk".
- A dangling link loads as "no manifest", so nothing is created.
- Your repro (R2): `dotfiles/manifest-target.json` is no longer created, and the warning is printed.
- Tests: "a manifest symlink is never written through" has two subtests.
  - Dangling link with files kept behind a symlinked `plugins/`: the target is not created.
  - Link to a real manifest: the target bytes are unchanged, the plugin is still removed, and the link itself stays.
- Mutation checks: disabling the regular-file guard, or dropping the existence check, each fail `TestDoUninstallOpencode_Manifest`.

cli/cmd/uninstall.go
alexandrocuma: Fixed in b7250d6.

- `runUninstall` resolves HOME through `uninstallHome`, before any target handler runs. An unset, empty or non-absolute HOME returns `HOME is "…", not an absolute path — refusing to remove anything; set HOME and re-run`, and the command exits 1.
- Your repro: with the install fixed so it actually populates the checkout (`install --reinstall-mcps`; the original V ran the wizard without a TTY and installed nothing), the run now prints the refusal and the checkout still has all 14 files. Before the fix it removed them.
- Tests:
  - `TestUninstallHome` (table): absolute, trailing slash, empty, relative, `.`, and a literal `~`.
  - `TestUninstallCmd_RefusesBadHome` runs the real command with HOME unset, empty and relative. Its current directory holds a dotfiles-style `.config/opencode/plugins` with a real entry, and the test checks that nothing there changes.
- Mutation checks: accepting a relative HOME, or accepting an empty one, each fail both tests.
- Install's identical pattern is out of scope here; the orchestrator is filing it.

Issue #128, opened 2026-09-16
Stale agent/skill removal follows a symlinked target directory


alexandrocuma:
Found during #117 (out of its scope).

## Problem
When `~/.claude/agents`, `~/.claude/skills`, `~/.config/opencode/agents` or `~/.config/opencode/commands` is itself a symlink (e.g. managed from a dotfiles or source checkout), stale-entry removal on re-install deletes through the link. A disabled or removed devexp agent can then be deleted from the linked source tree. #108 established the rule for opencode `plugins/`: write through a symlinked directory, never remove through it (#117 already refuses symlinked *entries*).

## Acceptance Criteria
- [ ] When a target agents/skills/commands directory is a symlink, installs still write through it, but no stale entry is removed through it; a warning lists what was left to remove by hand
- [ ] Manifest keeps entries that couldn't be removed so a later run (after the link is replaced) can clean them up
- [ ] Behaviour for real directories unchanged
- [ ] Tests for both CLIs, mutation-checked

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
## Groomed plan — #128

**Groomed**: 2026-09-16
**Validated against**: commit `10ad67c39707e1f8734ba44e8f7f7e64f0db824d` (`main`)
**Verdict**: CORRECTED AND PROCEEDING — the ticket is accurate for the Go installer; the plan adds `uninstall.sh`, which has the same problem, and the manifest change AC 2 needs.

### Health report (summary)
- Confirmed: stale removal in `removeStale` (`cli/cmd/backup.go`) checks the entry but not the target directory itself, so a symlinked `agents/`, `skills/` or `commands/` is removed through. Confirmed in a scratch HOME on `main`.
- Confirmed: #108's rule for opencode `plugins/` (write through, never remove through, left-behind files stay recorded) and #117's refusal of symlinked entries.
- Added: `uninstall.sh` removes agents and skills through a symlinked target directory too, and removes a symlinked entry, unlike the Go rule.
- Added: the installers record only this run's names in the manifest, so anything `removeStale` keeps is forgotten. AC 2 needs `removeStale` to return what it kept.
- Scope: only the target directory itself is checked, as `checkPluginRoots` does for `plugins/`. #123 (opencode skills in uninstall) and #124 (config/manifest writes) are out of scope.

### Policy
[...]

PR #130, opened 2026-09-16
fix(cli): only remove stale agents and skills with installed-name shapes (#117)

### LINKED ISSUE #117: Stale agent/skill removal trusts manifest names without filtering

alexandrocuma:
Closes #117

## Summary
Stale-entry removal on re-install joined names from the previous install manifest onto the target directory without checks. Claude Code skills are removed with `os.RemoveAll`, so a corrupted or hand-edited manifest entry such as `""`, `.`, `..` or `../..` could delete all of `~/.claude/skills`, `~/.claude`, or `$HOME`.

- `removeStale` (`cli/cmd/backup.go`) now takes the expected entry shape per target (Claude agent file `<name>.md`, Claude skill dir `<name>`, opencode agent `<name>.md`, opencode command recorded as `<name>` with `.md` added after validation) and removes only entries that are a bare installed-name shape **and** match that shape on disk (`Lstat`).
- Rejected entries (empty, `.`, anything with `..`, `/` or `\`, wrong suffix) and symlinked or wrong-kind entries are never removed; a warning names the entry exactly as recorded.
- Valid stale entries are removed exactly as before, for both Claude Code and opencode.
- Docs (`install.md`, `overview.md` install flow incl. a step-order correction, `conventions.md`) and CHANGELOG `[Unreleased] ### Fixed`.

## Out-of-scope (filed)
- #128 — stale removal still follows a symlinked target *directory* (e.g. `~/.claude/agents` linked into dotfiles)

## Test plan
- [x] Go `go vet` + `go test ./... -race -cover` — 539 tests/subtests, `cmd` 69.6%
- [x] New: `TestIsInstalledName`, `TestRemoveStale_TamperedEntries` (real + dry-run), `TestRemoveStale_EntryShape`, `TestDoInstall_TamperedAgentSkillManifest` (both CLIs, protected files survive, exact warning per entry)
- [x] Mutation checks: dropping the filter, the on-disk shape check, or any clause of the name rule fails a test; `.md` warning regression caught
- [x] Hook suites and `uninstall.test.sh` (77) unchanged and passing


cli/cmd/backup.go
```diff
@@ -66,12 +67,58 @@ func loadOldManifest(path string) *manifest.Manifest {
...
+		if !isInstalledName(name, shape) {
+			ui.Warn(fmt.Sprintf("%q left untouched: listed in the manifest but not a name devexp installs in %s", name, dir))
+			continue
+		}
+		file := name
+		if shape == staleCommand {
+			file += ".md"
+		}
+		path := filepath.Join(dir, file)
+		if fi, err := os.Lstat(path); err == nil && !hasStaleShape(fi, shape) {
```
alexandrocuma: 🟡 **Important (older than this PR; fine as a follow-up): on a case-insensitive filesystem (default APFS), a stale entry that differs only in case removes the agent or skill this same run just installed.**

`manifest.Stale` compares names case-sensitively, but `Lstat` here and the remove that follows match without regard to case on macOS. So an old entry `DEV-AGENT.md` passes `isInstalledName`, `Lstat` finds the regular file `dev-agent.md` that was just written, and it gets deleted. The new manifest still lists it as installed.

Reproduced with a binary built from `7fabed2`, temp `HOME` on APFS:
- Claude agents: manifest `{"agents":["DEV-AGENT.md"]}` → `devexp install --agents-only` prints `- DEV-AGENT.md`. `~/.claude/agents/dev-agent.md` is **gone** and the manifest lists `dev-agent.md`.
- Claude skills: `{"skills":["GRAPHIFY"]}` → `- GRAPHIFY`. `os.RemoveAll` deletes the freshly installed `~/.claude/skills/graphify/`.
- opencode: `{"skills":["DELIVER"]}` → `- DELIVER.md`. `~/.config/opencode/commands/deliver.md` is gone.

This happens without anyone editing the manifest: a case-only rename of an agent or skill between toolkit releases triggers it. It fixes itself on the next install. APFS also ignores Unicode normalization (NFC vs NFD), so the same thing can happen with accented names.

**Suggested fix:** skip any stale entry that `strings.EqualFold`s a name in this run's install set, with a warning. `removeStale` doesn't have the install set, so either filter in the callers before calling it, or pass the install set in. A test using `DEV-AGENT.md` against an installed `dev-agent.md` would pin it down, and it fails on macOS today.

alexandrocuma: 🔵 **Minor: when `Lstat` fails for any reason other than "not exist", removal goes ahead without the type check.**

`if fi, err := os.Lstat(path); err == nil && !hasStaleShape(fi, shape)` only runs the type check when `Lstat` succeeds. Any other error (EACCES, EIO, ENAMETOOLONG, EINVAL) falls through to `removeFn`, including `os.RemoveAll` for skills. #108's `removePluginFiles` does the opposite: `err == nil && !IsRegular` → keep, and any other error keeps the file and warns.

I couldn't get a delete out of this. For example, `a\0.md` fails `lstat` and then fails `unlinkat` with EINVAL. But it contradicts the doc comment ("removed only when … what is on disk has that shape") and differs from #108.

**Suggested fix:**
```go
fi, err := os.Lstat(path)
switch {
case os.IsNotExist(err):
    continue // or keep today's report
case err != nil:
    ui.Warn(fmt.Sprintf("%s left untouched: %v", path, err)); continue
case !hasStaleShape(fi, shape):
    ...
}
```
If a missing entry still needs to print `- name` (which is what happens today), keep that branch explicitly.

cli/cmd/backup.go
```diff
@@ -66,12 +67,58 @@ func loadOldManifest(path string) *manifest.Manifest {
...
+			ui.Warn(fmt.Sprintf("%s left untouched: no longer in this release, but %s", path, describeMode(fi)))
```
alexandrocuma: 🔵 **Minor: control characters in an accepted name reach the terminal as-is.**

Rejected entries are printed with `%q`, but a name that passes `isInstalledName` can still contain ESC, CR/LF or NUL. That name is then printed raw in this warning, in the `ui.DryRun` preview (L123), in the `remove %s: %v` warning and in `ui.Removed(file)` (L130). This is the same threat model as the PR, a hand-edited or corrupted manifest.

Reproduced (real run):
- skill entry `"x.md\n  - everything"` prints `- x.md` followed by a second line `- everything` that looks like another removal;
- agent entry `"\u001b[2J\u001b[31mFAKE.md"` clears the screen;
- `"a\u0000.md"` prints a raw NUL in the `remove … invalid argument` warning.

Nothing is deleted outside the target, so this only matters for output spoofing.

**Suggested fix:** reject names containing control characters in `isInstalledName`. `strings.ContainsFunc(name, unicode.IsControl)` does it, and no installed name has one. Or print `path`/`file` with `%q` on these lines too.

cli/cmd/install_test.go
```diff
@@ -505,10 +506,303 @@ func TestRemoveStale(t *testing.T) {
...
+		"a trailing separator":                      {name: "graphify/", shape: staleDir},
+		"an absolute file path":                     {name: "/etc/agent.md", shape: staleFile},
+		"an absolute directory path":                {name: "/etc", shape: staleDir},
+		"an opencode command, recorded without .md": {name: "deliver", shape: staleCommand, want: true},
+		"an opencode command recorded with .md":     {name: "deliver.md", shape: staleCommand, want: true},
+		"an empty opencode command":                 {name: "", shape: staleCommand},
+		"a parent-relative opencode command":        {name: "../precious", shape: staleCommand},
+		"an opencode command that gains a dot-dot":  {name: "x.", shape: staleCommand},
+		"an absolute opencode command":              {name: "/etc/precious", shape: staleCommand},
+		"a backslash separator":                     {name: `..\agent.md`, shape: staleFile},
```
alexandrocuma, line 555: 🔵 **Minor: no test covers the `\\` clause, so the PR's claim that dropping "any clause of the name rule fails a test" isn't true for it.**

The only backslash case is `..\agent.md`, which the `..` check already rejects. On a copy, I changed `` strings.ContainsAny(name, `/\`) `` to `` `/` ``. All `RemoveStale|IsInstalledName|TamperedAgentSkill` tests still pass, and 0 fail. Removing `/` instead makes 16 tests fail.

On darwin/linux `\\` isn't a separator, so no data is at risk. It's a guard with no test behind it.

**Suggested fix:** add a case with a backslash but no `..`, e.g. `"a backslash separator": {name: `sub\agent.md`, shape: staleFile}` and one for `staleDir` (`sub\skill`).

alexandrocuma (commented):
## PR Review: fix(cli): only remove stale agents and skills with installed-name shapes (#117)

**Recommendation**: Request changes (small ones). #117's ACs are met and I found no way to delete anything outside the target directory. The three Minor items are inside the diff. The Important one is older than this PR and can be a follow-up.

**Summary**: Every stale agent/skill removal now goes through `removeStale`, which checks the name (`isInstalledName`) and then the file type on disk (`Lstat` + `hasStaleShape`). I used a binary built from `7fabed2` with a temp `HOME` and a fake `claude`/`opencode` on `PATH`, and tried to make it delete something it shouldn't. Nothing outside `agents/`, `skills/` or `commands/` was ever removed, and neither was the directory itself.

### What I checked
- **Name shapes, on both CLIs, real runs:** `""`, `.`, `..`, ` ..`, `.. `, `..md`, `.md`, `a/.md`, `C:\x.md`, `..\precious.md`, `../../precious`, `x.`, NUL (`a\0.md`, `\0`, `.\0.`, `\0/../../precious.md`), whitespace (` `, `\t`, `. `), lookalikes (U+2024 `․․`, U+2025, fullwidth `．．／`, U+2215, `..\u200b`), plus `~`, `$HOME`, `%2e%2e`, `*`, `*.md`. After each run I compared a before/after listing of `$HOME`. The only things removed were direct children of the target directory whose names matched the entry byte for byte. NUL names are either rejected or fail at `lstat`/`unlinkat` with EINVAL.
- **Symlinks:** a symlinked agent file, skill directory or opencode command is kept, with a warning, and its target is untouched. `os.RemoveAll` on a real stale skill directory that contains symlinks (to a directory and to a file) unlinks the links and never follows them. The targets survived.
- **Hard links:** a stale agent that is a hard link, and a hard link inside a stale skill directory, lose only that name. The data survives through the other link.
- **Race between `Lstat` and remove:** in a small Go program I swapped a real directory for a symlink to a directory after the `Lstat`. `os.RemoveAll` unlinked only the link and the target survived. `os.Remove` behaves the same way. At worst the race removes a symlink.
- **Dry run:** a before/after listing of the whole `$HOME` (mode, size, mtime) was identical. The preview matched what the real run then did.
- **Valid names:** valid stale entries are still removed on both CLIs. No shipped agent or skill name contains `..`, `\\` or `/`.
- **Tests:** `go vet`, `gofmt -l` and `go test ./... -race` all pass. I mutated a copy of `backup.go` ten ways. Dropping the `""`, `.`, `..`, `/` or `!= ".md"` clause, the on-disk type check, the command `.md` check or the `.md` append each makes 2–16 tests fail. **One mutation survived:** dropping `\\` from the separator check (see the inline comment).
- **Docs:** `install.md`, `overview.md` (including the corrected opencode step order and "skipped in dry-run") and `conventions.md` match the code.

### What's done well
- The checks sit in `removeStale`, so every current and future caller gets them. The opencode `.md` handling moved in there too, and warnings name the entry exactly as the manifest records it.
- It follows #108's rule that a symlink is never removed. Unlike #108's `removePluginFiles`, the checks also run in dry-run, so the preview is accurate.
- The end-to-end test puts every file inside a temp root and compares the whole tree, so a removal anywhere shows up.

No guard hooks blocked anything during this review.

cli/cmd/backup.go
alexandrocuma: Fixed in 1e99fe3.

- `removeStale` now receives the previous manifest's list and this run's install set, and runs `manifest.Stale` itself. It keeps a stale entry, with a warning, in two cases:
  - it equals an installed name apart from case (`strings.EqualFold`), e.g. `"DEV-AGENT.md" left untouched: the same name as "dev-agent.md", installed by this run, apart from case`. This check runs before `Lstat`, so it works on case-sensitive filesystems and in `--dry-run` too.
  - it is the same file on disk as an installed item (`os.SameFile` against each installed path in the same dir). This also covers Unicode NFC/NFD variants on APFS. The stdlib has no normalization (`golang.org/x/text` is only an indirect dependency), so a dry run, which installs nothing, may still preview removing a variant that differs only in normalization. `CHANGELOG.md` and `docs/guides/install.md` say so.
- Tests:
  - `TestRemoveStale_InstalledTwin`: case variants for Claude agents, Claude skills, opencode agents and opencode commands, in real and dry runs. It has hard-link cases so `os.SameFile` is exercised on case-sensitive Linux CI.
  - `TestDoInstall_CaseVariantManifest`: runs the real Claude and opencode installs with `{"agents":["DEV-AGENT.md"],"skills":["GRAPHIFY"]}`.
- Mutation results:
  - Case check dropped: the case-variant unit cases and the install test fail.
  - Same-file check dropped: the hard-link cases fail.
  - Both dropped: the install test fails on APFS because the installed files are gone.

alexandrocuma: Fixed in 1e99fe3. `removeStale` now switches on the `Lstat` result:
- not exist: nothing to do. It's skipped silently, and no longer printed as `- name`.
- any other error: the entry is kept, with a warning (`"<path>" left untouched: permission denied`).
- wrong type: the entry is kept, with a warning.
- only a matching type goes on to the same-file check and the removal.

`lstat` is now a package variable, so `TestRemoveStale_LstatError` can inject `fs.ErrPermission` (agents, skill dirs, commands; real and dry runs) and `fs.ErrNotExist` (asserts no output and no remove). Mutation results: dropping the error branch (the old `err == nil && …` form) makes the permission cases fail. Letting not-exist fall through to removal makes the already-gone cases fail.

cli/cmd/backup.go
alexandrocuma: Fixed in 1e99fe3.
- `isInstalledName` now rejects any name with a control character (`strings.ContainsFunc(name, unicode.IsControl)`).
- Every line `removeStale` prints now quotes names and paths with `%q`: the rejected-name, case, same-file, `Lstat` error and wrong-type warnings, the dry-run preview, the `remove … :` error, and `ui.Removed`, which now prints e.g. `- "old.md"`. Errors print the underlying cause, without the `*PathError`'s raw path.

Tests:
- `TestIsInstalledName` has newline, ESC, NUL and CR names for all three shapes.
- `TestRemoveStale_TamperedEntries` adds `"evil\n  - everything.md"`, `"\x1b[2J\x1b[31mFAKE.md"` and similar entries. It asserts the exact quoted warning line, and that neither `\x1b[2J` nor `\n  - everything` shows up raw in the output.

Mutation results: dropping the control-character clause fails those tests. Printing without `%q` fails a test for each checked line: rejected, wrong type, same file, `Lstat` error, dry-run preview, remove error, and the removal line.

cli/cmd/install_test.go
alexandrocuma: Fixed in 1e99fe3. Added `sub\agent.md` (agent file) and `sub\skill` (skill dir) to `TestIsInstalledName`. Neither contains `..`. Changing the check to `` strings.ContainsAny(name, `/`) `` now fails `TestIsInstalledName/a_backslash_in_an_agent_file` and `/a_backslash_in_a_skill_directory`.

Issue #129, opened 2026-09-16
devexp install: errors print twice with the usage text


alexandrocuma:
Found during #126 (out of its scope).

## Problem
`installCmd` doesn't set cobra's `SilenceUsage` / `SilenceErrors`, so every install error (for example the HOME refusal added in #126) prints cobra's `Error: …` plus the full usage text, and then the message again from the caller.

## Acceptance Criteria
- [ ] An install error prints its message once, without the usage text; usage still prints for actual flag/argument misuse
- [ ] Exit codes unchanged
- [ ] Consistent across `install` and `uninstall` commands
- [ ] Test asserting the output of an install error

## Release Targets
- `cli`

alexandrocuma:
Closing — verified satisfied at `46d8a07` (v0.10.2) during an `/improve` cycle. Both acceptance criteria check out against a binary built from current `main`.

**AC1 — "An install error prints its message once, without the usage text; usage still prints for actual flag/argument misuse."** ✅

The fix is in place at two points in `cli/cmd/install.go`:

```
31:	SilenceErrors: true,          // on installCmd
90:	cmd.SilenceUsage = true       // first line of runInstall
```

`SilenceUsage` is set as the *first* statement of `runInstall`, after flag parsing — so a bad flag still gets usage (the line hasn't run yet) while anything failing later does not. The code comment says exactly that.

Runtime error (`HOME=` unset), once, no usage block:

```
$ HOME= devexp install --dry-run
HOME is "", not an absolute path — refusing to install anything; set HOME and re-run

error lines:  1
usage blocks: 0
```

Flag misuse still gets usage:

```
$ devexp install --bogus-flag
Usage:
  devexp install [flags]

Flags:
  …
```

**AC2 — "Exit codes unchanged."** ✅ Both paths exit `1`:

```
runtime error (HOME unset): exit 1
bad flag:                   exit 1
```

Note `go run . install --dry-run` with `HOME=` is **not** a valid way to reproduce this — the Go toolchain needs `HOME` for its build cache and fails before the binary runs. Build first, then run.

PR #131, opened 2026-09-16
fix(cli): refuse an unset, empty or relative HOME in install, uninstall and remote-install (#126)

### LINKED ISSUE #126: install/uninstall: refuse an empty or relative HOME
Found during review of #125 (out of its scope for install).

## Problem
The Go CLI resolves install targets from `HOME`. When `HOME` is unset or empty, target paths become relative and resolve under the current working directory, so an install (or any command using the same path helpers) can write into or remove files from a dotfiles-style tree in the cwd. #109 fixes this for `devexp uninstall`; `devexp install` has the same pattern.

## Acceptance Criteria
- [ ] `devexp install` refuses (non-zero, clear message) when `HOME` is unset, empty or not absolute, before touching anything
- [ ] Shared path helpers can't return relative target paths
- [ ] Tests for unset, empty and relative `HOME`

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
Closes #126

## Summary
With `HOME` unset, empty or relative, target paths resolved under `/` or the current working directory, so an install or uninstall could write into or delete from whatever tree the user was standing in.

- **`devexp install`** — the check is the first statement of `runInstall`, before `repo.Resolve` (which otherwise deletes and re-extracts the asset cache under a relative `os.UserCacheDir()`), MCP registration and any write. Covers every flag path, the wizard and its Remove action. Shared `targetHome` helper (from #109's `uninstallHome`); `claudeTargetPaths` / `opencodeTargetPaths` now return an error and can't produce relative paths.
- **`devexp uninstall`** — same helper, same message as before.
- **`uninstall.sh`** — refuses right after `REPO_DIR` is set, before detection or removal (bash 3.2, `set -euo pipefail`, no top-level `local`).
- **`scripts/remote-install.sh`** — refuses before OS detection/download when `DEVEXP_INSTALL_DIR` is relative, or unset with a bad `HOME`; an absolute `DEVEXP_INSTALL_DIR` works without `HOME`.
- Docs: `install.md` (incl. fixing env var examples that set the variable for `curl` instead of `bash`), `setup.md`, `testing.md`; `CLAUDE.md` line reference; CHANGELOG `[Unreleased] ### Fixed`.

## Out-of-scope (filed)
- #129 — install errors print twice with cobra's usage text

## Test plan
- [x] Go `go vet` + `go test ./... -race -cover`: `TestInstallCmd_RefusesBadHome` (15 subtests × flag paths, standalone, wizard — cwd tree byte-identical, nothing under `TMPDIR`), `TestDoInstall_RefusesBadHome`, `TestTargetHome`, path helper cases, uninstall message
- [x] `uninstall.test.sh` 89 passed (12 new: unset/empty/relative HOME, tree checksum unchanged)
- [x] New `remote-install.test.sh` 24 passed (stub `curl`, `env -i`, no network), picked up by CI's `./*.test.sh`
- [x] Mutation checks: each guard removed → tests fail (incl. files deleted under cwd / download attempted)
- [x] Built binary with `HOME=""`, `HOME=home`, `HOME` unset → exit 1, nothing written
- [x] Hook suites unchanged and passing


cli/cmd/install.go
```diff
@@ -70,6 +71,14 @@ type wizardResult struct {
 // ── Entry point ───────────────────────────────────────────────────────────────
 
 func runInstall(cmd *cobra.Command, args []string) error {
+	// Before anything else, flags and wizard alike: repo.Resolve may already
+	// write (a standalone binary extracts its assets under the user cache dir,
+	// which a relative HOME puts under the current directory), and MCP
+	// registration runs before any target path is built.
+	if _, err := targetHome(os.Getenv("HOME")); err != nil {
```
alexandrocuma, line 78: 🟡 **Important: a relative `XDG_CACHE_HOME` still gets asset extraction under the cwd (Linux)**

The CHANGELOG and this comment say the check comes before "asset extraction". That only holds when the problem is HOME itself. On Linux, `os.UserCacheDir()` reads `XDG_CACHE_HOME` first and returns an error when it's relative. `extractEmbedded` (`cli/internal/repo/repo.go:89-92`) then silently falls back to `os.TempDir()`, which is `$TMPDIR` unchecked, or `/tmp`.

**Repro** (linux/arm64 build of `fa5bb63`, `FROM scratch` container, valid absolute HOME):
```
cwd/tmprel/devexp/assets/precious.txt   # pre-existing
docker run -w /work/cwd -e HOME=/work/home -e XDG_CACHE_HOME=relcache -e TMPDIR=tmprel \
  -e PATH=/work/stubs devexp /devexp install --agents-only
# → precious.txt deleted; cwd/tmprel/devexp/assets/{agents,hooks,mcps,…} extracted
```
With `TMPDIR` unset, the fallback is the shared `/tmp/devexp/assets`, and it is reused whenever `.devexp-version` matches. I pre-planted `/tmp/devexp/assets` with marker `dev` and an extra `agents/planted-by-other-user.md`. `devexp install --agents-only` then installed that agent into `~/.config/opencode/agents/`. For Claude Code, hook commands get registered pointing into that directory, so another local user's scripts would run on every tool call.

The code is pre-existing, but it's the bug class this PR sets out to close.

**Suggested fix:** never fall back to a relative or shared directory. For example, in `extractEmbedded`:
```go
base, err := userCacheDir()
if err != nil || !filepath.IsAbs(base) {
    return "", fmt.Errorf("no usable user cache dir (%v) — set XDG_CACHE_HOME to an absolute path or unset it", err)
}
```
Also add a `XDG_CACHE_HOME=rel` case to `TestInstallCmd_RefusesBadHome` (Linux) or a unit test on `extractEmbedded` with `userCacheDir` stubbed to return an error. If you'd rather not fix it here, please file a follow-up issue.

cli/cmd/install_test.go
```diff
@@ -1515,3 +1576,150 @@ func TestDoInstallOpencode_SymlinkedPluginsDir(t *testing.T) {
...
+// HOME puts under the current directory), before the wizard, and before MCP
+// registration — so a dotfiles tree in the current directory is left byte for
+// byte as it was.
+func TestInstallCmd_RefusesBadHome(t *testing.T) {
+	runs := map[string]struct {
+		clis       []string
+		args       []string
+		standalone bool // no DEVEXP_DIR: assets come from the binary
+	}{
+		"flags, claude":     {clis: []string{"claude"}, args: []string{"install", "--reinstall-mcps"}},
```
alexandrocuma, line 1694: 🟡 **Important: the tests can't observe MCP registration, so "refuses before MCP registration" isn't proven**

These cases run with `DEVEXP_DIR=writeOpencodeHookRepo(t)`, whose `mcps/registry.json` is `[]`. `fakeCLI` writes `#!/bin/sh` stubs that do nothing. So `--reinstall-mcps` and `--mcps-only` make **zero** `claude` calls whether or not the guard works, and a call would leave no trace anyway. The standalone case uses the real registry but runs with `--agents-only`.

**Mutation that survives** (scratch copy): I added `installMCPsClaude(opts)` *before* `claudeTargetPaths` in `doInstallClaude`. `TestDoInstall_RefusesBadHome` and `TestInstallCmd_RefusesBadHome` both still pass. A real `claude mcp add` with `HOME=home` writes `home/.claude.json` under the cwd.

The mutation is caught once both of these hold: (a) the fixture registry has one stdio MCP, e.g. `[{"name":"probe","command":"echo","args":["hi"],"scope":"user"}]`, and (b) the fake `claude` logs its calls. With both, `TestDoInstall_RefusesBadHome/claude,_HOME_relative` fails.

**Suggested fix:** give these HOME-refusal tests a registry with one MCP. Use a fake `claude`/`opencode` that appends `$0 $*` to an **absolute** log file outside the cwd (e.g. `CALLS=<t.TempDir()>/calls`, so unset/empty HOME can't hide the call), and assert the log is empty. Apply the same change to the `fakeCLI(t, "claude", "opencode")` call in `TestDoInstall_RefusesBadHome` (line 1707).

cli/cmd/install_test.go
```diff
@@ -1515,3 +1576,150 @@ func TestDoInstallOpencode_SymlinkedPluginsDir(t *testing.T) {
...
+
+// writeDotfilesTree fills dir with what a bad HOME would make devexp read,
+// write or remove there: an existing Claude Code and opencode install at the
+// top (HOME empty) and under home/ (HOME=home), each with a stale agent the
+// manifest lists, plus an embedded-asset cache from another version under
+// home/'s user cache dir (darwin and linux), which repo.Resolve would wipe.
+func writeDotfilesTree(t *testing.T, dir string) {
+	t.Helper()
+	files := map[string]string{
+		"Library/Caches/devexp/assets/.devexp-version": "some-other",
```
alexandrocuma: 🔵 **Minor: the cache fixture is never under `home/`, so the destructive path isn't tested**

The doc comment says there's "an embedded-asset cache from another version under home/'s user cache dir, which repo.Resolve would wipe". But `Library/Caches/…` and `.cache/…` are only added to `files` at the top level. The loop that adds the `home/` copies runs over the *other* map. With `HOME=home` the relevant cache is `home/Library/Caches/devexp/assets`, which doesn't exist, so a missing guard shows up only as newly *created* files, never as the `os.RemoveAll` of an existing tree. The top-level copies can't be reached either: with an empty HOME, `UserCacheDir` returns an error.

**Suggested fix:** move the two cache entries into the map the loop copies, or add `home/`-prefixed keys, so the relative-HOME case covers the wipe.

uninstall.sh
```diff
@@ -24,6 +24,13 @@ find_devexp_bin() { for c in "${DEVEXP_BIN:-}" "$REPO_DIR/bin/devexp" "$(command
 
 REPO_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
+# ── HOME ──────────────────────────────────────────────────────────────────────
+# Every path below is built from HOME. Unset, empty or relative, it would point
+# at / or at the current directory (a dotfiles checkout, say), so refuse before
+# looking at anything — the same rule `devexp install` / `devexp uninstall`
+# apply (#126).
+[[ "${HOME:-}" == /* ]] || die "HOME is \"${HOME:-}\", not an absolute path — refusing to remove anything; set HOME and re-run"
```
alexandrocuma, line 32: 🔵 **Minor: `install.sh` needs the same guard. In a fresh clone it builds before `devexp install` can refuse**

`install.sh` isn't in this diff. When `bin/devexp` is missing it runs `stage-assets.sh` and `(cd cli && go build …)` with whatever HOME it was given. Go's default cache dirs come from HOME.

**Repro** (copy of the worktree without `bin/`, cwd elsewhere):
```
env -i HOME=home GOPATH=$(go env GOPATH) GOMODCACHE=$(go env GOMODCACHE) PATH=$GOBIN:/usr/bin:/bin \
  GOPROXY=off /bin/bash <copy>/install.sh --dry-run
# → build succeeds, writes <copy>/cli/home/Library/Caches/go-build/** and
#   <copy>/cli/home/Library/Application Support/go/telemetry, THEN "refusing to install anything"
```
Without `GOPATH`/`GOMODCACHE` in the env, the build fails with `GOMODCACHE entry is relative`, and `install.sh` prints the misleading `Build failed. Ensure Go is installed`. `cli/home/` is not gitignored. `docs/guides/install.md` now says `./install.sh` "Refuses to run … before it resolves assets…", which isn't true for a fresh clone.

**Suggested fix:** add the same one-liner at the top of `install.sh`, before the build:
```bash
[[ "${HOME:-}" == /* ]] || { echo "HOME is \"${HOME:-}\", not an absolute path — refusing to install anything; set HOME and re-run" >&2; exit 1; }
```

cli/cmd/install_claude.go
```diff
@@ -17,7 +17,10 @@ import (
 // ── Claude Code ───────────────────────────────────────────────────────────────
 
 func doInstallClaude(opts *installOpts) error {
-	p := claudeTargetPaths(os.Getenv("HOME"), time.Now())
+	p, err := claudeTargetPaths(os.Getenv("HOME"), time.Now())
```
alexandrocuma, line 20: 🔵 **Minor (follow-up candidate): a relative `CLAUDE_CONFIG_DIR` still sends `claude mcp` writes to the cwd**

The `claude` subprocess ignores HOME when `CLAUDE_CONFIG_DIR` is set. HOME is validated now, but a relative `CLAUDE_CONFIG_DIR` is passed straight through, both here (`mcp.InstallClaude` → `claude mcp list/remove/add`) and in `uninstall.sh`'s `claude mcp remove` block.

**Repro** (real `claude` 2.1.273, temp HOME and cwd):
```
cd $tmp/cwd && env -i HOME=$tmp/home PATH=/usr/bin:/bin CLAUDE_CONFIG_DIR=relcfg claude mcp add --scope user probe -- echo hi
# → "File modified: relcfg/.claude.json"; cwd now has relcfg/.claude.json and relcfg/backups/
```
**Suggested fix:** in `runInstall` (and `uninstall.sh`), refuse a set but non-absolute `CLAUDE_CONFIG_DIR` with the same message style. A related pre-existing issue could go in the same follow-up: when `CLAUDE_CONFIG_DIR` is set to an absolute path, agents, skills and `settings.json` still go to `$HOME/.claude`, where Claude Code won't read them.

scripts/remote-install.sh
```diff
@@ -13,11 +13,22 @@
...
 
+# ── Resolve the install directory ─────────────────────────────────────────────
+# Before any download. The default is built from HOME: unset, empty or relative,
+# it would be /.local/bin or a directory under wherever this runs (#126). An
+# explicit DEVEXP_INSTALL_DIR doesn't need HOME, but must be absolute too.
+if [[ -n "${DEVEXP_INSTALL_DIR:-}" ]]; then
+    INSTALL_DIR="$DEVEXP_INSTALL_DIR"
+    [[ "$INSTALL_DIR" == /* ]] || die "DEVEXP_INSTALL_DIR is \"$INSTALL_DIR\", not an absolute path — refusing to install; set it to an absolute path and re-run"
+else
+    [[ "${HOME:-}" == /* ]] || die "HOME is \"${HOME:-}\", not an absolute path — refusing to install; set HOME (or an absolute DEVEXP_INSTALL_DIR) and re-run"
```
alexandrocuma, line 28: 🔵 **Minor (docs): the 11 new lines shift `docs/guides/release.md`'s line references**

`setup.md` was updated for the shift, but `docs/guides/release.md` wasn't, and all of its `remote-install.sh` references are now 11 lines off:
- `:40-44` (releases/latest lookup) → `:51-55`
- `:41` and `:42` (`releases/latest` curl) → `:53`
- `:38` (`DEVEXP_VERSION`) → `:50`. This one was already one line off before this PR.
- `:50-72` (download → `--version`) → `:61-83`

**Suggested fix:** update those five references in `release.md` (lines 51, 57, 58, 77, 78). The `DEVEXP_INSTALL_DIR="$(mktemp -d)"` release check there still passes the new guard, since `mktemp -d` returns an absolute path.

alexandrocuma (commented):
## PR Review: fix(cli): refuse an unset, empty or relative HOME (#126)

**Recommendation**: Request changes (posted as a comment). Two of the findings are in this diff (test coverage). The rest are gaps in the same class of bug, outside the diff.

**Summary**: The HOME guard is in the right place. It is the first statement of `runInstall`, before `repo.Resolve`, the wizard (including Remove) and MCP registration. The path helpers can no longer return a relative path, and the `uninstall.sh` and `remote-install.sh` guards work under bash 3.2 with `set -euo pipefail`. All three ACs of #126 are met. I tested a binary built from `fa5bb63` in temp dirs, and a linux build in a `FROM scratch` container. Unset, empty, relative, `.`, `~` and leading-space HOME all exit 1 with nothing written, on flag paths, `--dry-run`, the wizard and `uninstall`. Legitimate setups still work: HOME with a trailing slash, HOME as a symlink, and a non-ASCII path with a space. Other ways to reach the same cwd write remain: a relative `XDG_CACHE_HOME` on Linux, `./install.sh` in a fresh clone, and a relative `CLAUDE_CONFIG_DIR`. The tests also can't see MCP registration.

### Severity index
| # | Sev | Where | Finding |
|---|-----|-------|---------|
| 1 | 🟡 Important | `cli/cmd/install.go:78` → `cli/internal/repo/repo.go:89-92` | Relative `XDG_CACHE_HOME` (Linux): asset cache is wiped/extracted under cwd, or reused from shared `/tmp` |
| 2 | 🟡 Important | `cli/cmd/install_test.go:1654,1707` | Tests never run `claude mcp`: empty MCP registry + silent fake. Moving MCP registration before the HOME check still passes |
| 3 | 🔵 Minor | `install.sh` (anchored on `uninstall.sh:32`) | `./install.sh` in a fresh clone runs `go build` before any guard. It writes a Go build cache into `<repo>/cli/home/…` |
| 4 | 🔵 Minor | `cli/cmd/install_claude.go:20` | Relative `CLAUDE_CONFIG_DIR`: `claude mcp add/remove` writes `.claude.json` under cwd |
| 5 | 🔵 Minor | `docs/guides/release.md:51,57,58,77,78` (anchored on `scripts/remote-install.sh:28`) | `remote-install.sh` line references are now 11 lines off |
| 6 | 🔵 Minor | `cli/cmd/install_test.go:1598` | Cache fixture is never under `home/`, so the destructive `RemoveAll` of an existing cache is not tested |

### Checked, no action needed
- `HOME=/tmp/../x`-style values: `targetHome` cleans the path lexically, and `filepath.Join` already did that before this PR. The only divergence is when a component before `..` is a symlink. Go writes to the lexical path, while the `claude` subprocess and `uninstall.sh` (raw `$HOME`) follow the kernel's resolution. I reproduced this with `HOME=<a>/lnk/../h`: agents and cache went to `<a>/h`. This is pre-existing and very low risk.
- `HOME="/abs/me "` (trailing space) is accepted and creates a `me ` directory. The path is absolute and it's what the user set, so I'd leave it.
- `HOME=/` is accepted, which is correct: Docker sets it for users with no passwd entry.
- `remote-install.sh` handles a `DEVEXP_INSTALL_DIR` with a trailing slash, spaces, `..` or non-ASCII characters. I ran it with a stub `curl` and a fake tarball. A trailing slash only makes the "not on your PATH" note misfire, which predates this PR. A bad HOME with an absolute `DEVEXP_INSTALL_DIR` downloads, then `devexp install` refuses, as documented.
- Nothing is read before the guard in `uninstall.sh` (only `REPO_DIR` is set) or in `remote-install.sh`.
- The wizard's Remove action inherits the validated HOME, and `uninstall.sh` has its own guard.
- Nothing in `cli/` reads `XDG_CONFIG_HOME`/`UserConfigDir`, and `opencode` is never executed.
- Test runs: `/bin/bash` 3.2.57 `remote-install.test.sh` 24/24, `uninstall.test.sh` 89/89. `go vet` and `go test ./... -race` are green.
- The corrected `curl … | VAR=x bash` examples in `install.md` are right. The `bash <(curl …)` form in `release.md` was already correct.

### ✅ What's done well
- The guard runs before `repo.Resolve`, which catches the wipe-and-re-extract that a check inside the path helpers alone would miss.
- `claudeTargetPaths`/`opencodeTargetPaths` now return an error, so a relative target path can't be represented.
- The tests compare the whole directory tree (`treeState` includes directories) and also check `TMPDIR`. The shell tests use `env -i` and a stub `curl`, so they never touch the network or the real HOME.
- The shell guards are bash 3.2-safe (no top-level `local`; `${HOME:-}` works under `set -u`).

cli/cmd/install.go
alexandrocuma: Fixed in 2e6042f. `extractEmbedded` (`cli/internal/repo/repo.go:96-102`) no longer falls back to `os.TempDir()`. If the cache dir lookup fails, or returns a relative path (for example a relative `XDG_CACHE_HOME` on Linux, or a relative HOME), it returns an error asking for an absolute `HOME` (or `XDG_CACHE_HOME`). Nothing is extracted to a relative `$TMPDIR` or to the shared `/tmp/devexp/assets` anymore, and that was the only `TempDir` use in `repo.Resolve`.

Test: `TestExtractEmbedded_NoUsableCacheDir` goes through the `userCacheDir` seam, so it doesn't depend on the OS. It covers a lookup error and a relative dir, each with a relative and an absolute `TMPDIR`, and checks that nothing is written under the cwd or `TMPDIR`. Mutations: putting back the original fallback fails all 4 subtests (files under the cwd or `TMPDIR`). Removing only the error branch fails the lookup-error cases. Removing only the absolute-path check fails the relative-dir cases. CHANGELOG has a Fixed bullet for this.

cli/cmd/install_test.go
alexandrocuma: Fixed in 2e6042f. Both refusal tests now use `refusalRepo` and `loggingCLI` (`cli/cmd/install_test.go`). `refusalRepo` is the hook repo plus one stdio MCP, `probe` (`echo hi`, user scope). The `loggingCLI` stubs for `claude` / `opencode` append `$0 $*` to a log file in a separate `t.TempDir()`, with the absolute path written into the script, so HOME and cwd can't redirect it. `noCalls` then checks the log was never created, in every bad-HOME case, including the `fakeCLI(t, "claude", "opencode")` call in `TestDoInstall_RefusesBadHome`.

Mutations: your one (`installMCPsClaude(opts)` before `claudeTargetPaths` in `doInstallClaude`) now fails `TestDoInstall_RefusesBadHome/claude` for unset, empty and relative HOME with "CLI called despite the HOME refusal". Adding that and also removing the `runInstall` guard fails all 15 `TestInstallCmd_RefusesBadHome` subtests. The claude and mcps-only cases fail on the call log.

cli/cmd/install_test.go
alexandrocuma: Fixed in 2e6042f. The two cache entries now go through the same loop as the rest of the fixture, so they exist at the top level and under `home/`. Each also has an extra `precious.txt` besides `.devexp-version`.

Mutation: removing the `runInstall` guard and restoring the old `extractEmbedded` fails `TestInstallCmd_RefusesBadHome/flags,_standalone,_HOME_relative`, and its tree diff shows `home/Library/Caches/devexp/assets/precious.txt` gone, so the `RemoveAll` is detected, not just new files. With the new `extractEmbedded`, removing only the guard is still caught, because extraction refuses the relative cache dir and the message check fails.

uninstall.sh
alexandrocuma: Fixed in 2e6042f. `install.sh:6` is the one-line guard from your comment (same message, exit 1), before `mkdir bin`, `stage-assets.sh` and `go build`. It takes the place of the blank line before the build, so no `install.sh:N` reference in the docs moves.

Test: new root `install.test.sh`, which CI's `./*.test.sh` glob picks up, 17 checks. For unset, empty and relative HOME it runs a copy of `install.sh` in a temp repo with no `bin/`, with a stub `scripts/stage-assets.sh` and a stub `go` on PATH that log to an absolute file. It checks exit 1, the message, no build attempted, and that the repo and a dotfiles-style cwd are unchanged. An absolute-HOME case checks the build does run. Mutation: removing the guard gives 9 failures (message, build attempted, repo changed, for each of the three). I also rewrote the `docs/guides/install.md` bullet: `install.sh` checks before the build, and `devexp install` checks again first, before it resolves assets.

scripts/remote-install.sh
alexandrocuma: Fixed in 2e6042f. I checked each reference against the current `scripts/remote-install.sh` and your mapping is right. The `releases/latest` block is `:40-44` → `:51-55`. The pre-release fallback and the Latest-resolution row pointed at the lookup curl (`:41` / `:42`), now `:53`. `DEVEXP_VERSION` is `:38` → `:50`. The download through `--version` is `:50-72` → `:61-83` (61 = `url=`, 83 = `--version`). Line 7 (the one-liner in the usage comment) and the `DEVEXP_INSTALL_DIR="$(mktemp -d)"` release check didn't need changes.

cli/cmd/install_claude.go
alexandrocuma: Out of scope for #126; tracked in #132.

cli/internal/hooks/installer.go
```diff
@@ -131,6 +131,12 @@ func ParseRegistry(data []byte) (Registry, error) {
 }
 
 func InstallClaude(registry Registry, repoDir, settingsPath string, disabled []string, dryRun bool) error {
+	// Every command is registered as repoDir/<script> and run later from
+	// whatever directory Claude Code is in, so it must be absolute (#126).
+	if !filepath.IsAbs(repoDir) {
+		return fmt.Errorf("hooks: repo dir %q is not an absolute path, so hook commands would be relative — refusing to register them", repoDir)
```
alexandrocuma, line 138: 🟡 **Important (pre-existing, not introduced here): an absolute command still doesn't run the intended script if the path needs shell quoting**

This check makes the commands absolute, but that alone doesn't make them run. Claude Code runs a hook `command` through a shell, and `filepath.Join(repoDir, cc.Script)` (line 176) is written in unquoted. If the repo or cache path contains whitespace or shell metacharacters, the shell splits the command or interprets it.

Checked locally: a clone under a directory with a space in its name registers `…/My Proj/clone/hooks/claude-code/secret-guard.sh`. Then `sh -c "$cmd"` fails with `…/My: No such file or directory`, exit 127. Claude Code treats that as a non-blocking hook error, so the guard fails open. If a program exists at the truncated path, that program runs instead.

Fix, pick one:
- Refuse a `repoDir` that needs quoting. This is the smallest change and matches this PR's refuse-don't-guess approach.
- Or register a POSIX single-quoted command. Then every matcher (the `alreadyIn` comparison, `isStaleDevexpHook`, `isForeignDevexpHook`, `isRelativeDevexpHook`, and `uninstall.sh`'s `is_devexp_hook`, e.g. via `shlex.split`) has to unquote the first word before comparing.

Either way, add a test case with a space in the repo dir.

alexandrocuma (commented):
Focused re-review of 2e6042f and 7448047.

What I checked: `go test ./...` and all `./*.test.sh` pass (install 17/17, remote-install 24/24, uninstall 91/91). I ran end-to-end installs with a built binary in temp HOMEs: from a clone, standalone from the embedded cache, and with an absolute, relative (`../clone`) or `~/` `DEVEXP_DIR`. Every hook command registered is absolute and points at the intended root. `~/…` is refused with a clear message. `uninstall.sh` removes all 7 registrations for both the clone and the cache installs. Two mutations of the `install.sh` guard are caught by `install.test.sh`: deleting the guard fails 9 checks, and weakening it to `-n` fails the 3 relative-HOME checks.

The `extractEmbedded` fix is correct. It has no remaining fallback to `os.TempDir()` or any shared or relative location, and its error says what to set. Putting `absoluteSource` on both `Resolve` arms together with the `IsAbs` check in `InstallClaude` is good defence in depth.

Three findings below. The first two are older than this PR, but they fall under the same rule this PR sets up: "a hook command must run the intended script." They could be follow-up issues. The third is new in 7448047.

cli/internal/repo/repo.go
```diff
@@ -62,7 +94,7 @@ func findRepoDir() (string, error) {
 		}
 		dir = parent
 	}
-	return "", fmt.Errorf("not found")
+	return "", errRepoNotFound
```
alexandrocuma, line 97: 🟡 **Important (pre-existing, follow-up candidate): the walk-up picks any ancestor that has `agents/`, `skills/` and `mcps/` as the asset root, release binaries included**

The binary-adjacent check and the cwd walk-up only test for those three directories. So `devexp install` run anywhere inside a tree that has them uses that tree as `RepoDir`, even from a release binary on `PATH`. The tree's agents and skills get copied into the user config. Every script in its `hooks/registry.json` gets registered in the user-level `settings.json` as an absolute command, and those commands then run in every session. When the root isn't embedded, nothing tells the user which root was used (only the standalone case prints a notice).

Checked locally: I ran a standalone binary from a subdirectory of a scratch tree that had those three dirs plus a one-entry hooks registry. It registered that tree's script under `PreToolUse` without printing any notice.

Fix:
- Require a toolkit-specific marker in `isRepoDir`, not just directory names. For example, `hooks/registry.json` plus `cli/go.mod` declaring `module devexp`, or a dedicated marker file.
- Always print the resolved asset root before writing anything.
- Consider skipping the walk-up (and possibly the binary-adjacent lookup) for tagged builds unless `DEVEXP_DIR` is set.

cli/internal/hooks/installer.go
```diff
@@ -314,11 +320,21 @@ func isForeignDevexpHook(cmd string, managed map[string]bool, repoDir string) bo
...
 }
 
+// isRelativeDevexpHook reports whether cmd is a devexp-managed hook registered
+// with a relative path — what an install from a relative DEVEXP_DIR wrote
+// before repo dirs were made absolute (#126). It is matched like
+// isForeignDevexpHook, by registry basename plus the script directory, and
+// replaced by the absolute registration on the same run.
+func isRelativeDevexpHook(cmd string, managed map[string]bool) bool {
+	return cmd != "" && !filepath.IsAbs(cmd) && managed[filepath.Base(cmd)] &&
+		strings.Contains(filepath.ToSlash(cmd), scriptDir)
```
alexandrocuma: 🔵 **Minor (new in 7448047): `isRelativeDevexpHook` matches any command string that isn't an absolute path, and the script-dir check is a substring match**

An install from a relative `DEVEXP_DIR` only ever wrote a plain `filepath.Join` path. But `!filepath.IsAbs(cmd)` is also true for shell-expanded or wrapped commands. Each of these is now pruned on every install with the reason "relative path":
- `$CLAUDE_PROJECT_DIR/hooks/claude-code/secret-guard.sh` and `\"$CLAUDE_PROJECT_DIR\"/hooks/claude-code/…`
- `~/vendor/hooks/claude-code/…` and `$HOME/…/hooks/claude-code/…`
- `FOO=1 /abs/hooks/claude-code/…` and `bash /abs/hooks/claude-code/…`

Before this commit the installer kept all of these. The first two point at per-project copies, so replacing them with the global registration changes what runs.

Separately, `strings.Contains(…, scriptDir)` isn't anchored to a path component. `my-hooks/claude-code/secret-guard.sh` matches, while the new "relative user hook sharing a basename" case only covers `my-hooks/secret-guard.sh`. The same unanchored match already exists in `isForeignDevexpHook` (`/home/me/my-hooks/claude-code/secret-guard.sh` is judged foreign) and in `uninstall.sh`. In the other direction, a devexp command with arguments or quotes is never matched by either function (pre-existing).

Fix:
- Treat a command as a relative devexp hook only if it is a plain path: no whitespace, `$`, `~`, quotes or backticks.
- Anchor the directory: after `ToSlash`, require a `hooks/claude-code/` prefix or a `/hooks/claude-code/` substring. Apply the same anchoring in `isForeignDevexpHook` and `is_devexp_hook`.
- Add the cases above to `TestIsRelativeDevexpHook`.

alexandrocuma: Fixed in 39f4ef6. All three matchers now use one rule. A command counts as devexp's only if:
- it is a plain path, with no whitespace, `$`, `~`, quotes, backticks or other shell syntax;
- its basename is a registry script;
- `hooks/claude-code/` sits directly above that basename, at the start of the command or right after a `/`.

The Go side is `isManagedScriptPath`, used by both `isRelativeDevexpHook` and `isForeignDevexpHook`. `uninstall.sh`'s `is_devexp_hook` applies the same rule. So `$CLAUDE_PROJECT_DIR/…`, `"$CLAUDE_PROJECT_DIR"/…`, `~/…`, `$HOME/…`, `FOO=1 /abs/…`, `bash /abs/…`, `my-hooks/claude-code/…` and `/home/me/my-hooks/claude-code/…` are all left alone. I didn't widen matching for devexp commands that have arguments or quotes; those are still not matched, as before.

Tests: every case above is in `TestIsRelativeDevexpHook`, plus `hooks/claude-code/sub/…`, `;` and a trailing argument. `TestIsForeignDevexpHook` has the segment-boundary and wrapped/quoted absolute negatives. `TestPruneForeignDevexpHooks` checks that all of them are kept. `uninstall.test.sh` has three new cases showing these hooks survive uninstall.

Mutations: dropping the plain-path check fails 6 `TestIsRelativeDevexpHook` cases and the prune case in Go, and 2 `uninstall.test.sh` cases. Dropping the segment anchor (a substring match instead) fails 2 cases each in `TestIsForeignDevexpHook` and `TestIsRelativeDevexpHook` plus the prune case, and 1 `uninstall.test.sh` case.

cli/internal/hooks/installer.go
alexandrocuma: Out of scope for #126; tracked in #135

cli/internal/repo/repo.go
alexandrocuma: Out of scope for #126; tracked in #134

PR #133, opened 2026-09-16
fix(hooks): resolve on-save paths to absolute and run jest 29/30 alike (#121)

### LINKED ISSUE #121: on-save hooks: pass file paths after an end-of-options separator
Found during the #119 audit (low risk, out of its scope).

## Problem
The format-on-save and lint-on-save hooks (Claude Code and opencode) pass the edited file path to formatters/linters as a positional argument without an end-of-options separator. A path beginning with `-` could be read as an option by the tool. Current CLIs pass absolute paths, so this is defense in depth.

## Acceptance Criteria
- [ ] Paths are passed after `--` (or made unambiguous, e.g. `./`-prefixed) for every tool that supports it; tools that don't are handled per their conventions
- [ ] Behaviour for normal absolute paths unchanged
- [ ] Tests cover a leading-dash relative path for both CLIs

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
Closes #121

## Summary
The format-, lint- and test-on-save hooks (Claude Code and opencode) passed the edited path to tools as a bare positional argument. A path starting with `-` was read as an option by ruff, black, flake8, prettier, eslint, gofmt and rubocop, and jest's `--testPathPattern <path>` could run the wrong tests. Relative paths were also resolved against the hook's own working directory while the tool ran from the project root.

- **Absolute paths first** — the edited path is resolved (and normalised) against the payload `cwd` (Claude Code) / `ctx.directory` (opencode), else the hook's working directory, before the project root, tool cwd and argv are chosen. Every tool now receives an absolute path; absolute-path argv is byte-identical to before for all non-jest tools.
- **Jest** — `jest --passWithNoTests --no-coverage --runTestsByPath -- <path relative to the project root>`: runs exactly that file (a plain pattern is a regex that could match other tests or none) and works on Jest 29 and Jest 30 (where `--testPathPattern` was removed, so test-on-save was broken). vitest keeps its absolute path (it drops file filters after `--`).
- **`--` wasn't used generally**: ruff expands `@file` arguments even after `--`, and vitest ignores filters after it.
- **Docs** — hook authoring guide: resolve to absolute first; opencode modules use the async `runCommand` (the section previously recommended `execFileSync`, contradicting the guide's own rule); `editedPath` utility; `conventions.md`, `testing.md`; CHANGELOG `[Unreleased] ### Fixed`.

## Verification with real tools
Leading-dash files handled correctly through the new hooks by ruff (format, check), black, prettier, eslint, gofmt; jest 29.7.0 and 30.5.0 run exactly the intended test (normal path, `a+b.test.js`, `c(1).test.js`, `mod.test.js` next to `mod.test.jsx`/`sub/mod.test.js`, `-foo.test.js`, `-dir/x.test.js`), `--passWithNoTests` honoured. With the previous hooks these fail (option errors; Jest 30 fails on every run).

Also: `agents/onboarding.md` no longer recommends `--testPathPattern` (rejected by Jest 30).

## Test plan
- [x] `hooks/claude-code/on-save-path.test.sh` 73 · `hooks/opencode/on-save-path.test.js` 61 (new) — exact argv and cwd per tool branch via stub tools; leading-dash, nested project, payload `cwd` / `ctx.directory`, no-`cwd` fallback. 26 of them fail against the previous hooks.
- [x] Mutation checks: dropping path resolution, ignoring payload `cwd` / `ctx.directory`, reverting jest to `--testPathPattern` or dropping `--` each fail tests
- [x] All other hook suites, Go suite, `uninstall.test.sh` pass


hooks/claude-code/test-on-save.sh
```diff
@@ -90,17 +100,19 @@ if ext in ('.js', '.jsx', '.ts', '.tsx', '.mjs', '.cjs'):
 
     local_jest    = os.path.join(root, 'node_modules', '.bin', 'jest')
     local_vitest  = os.path.join(root, 'node_modules', '.bin', 'vitest')
+    # jest 29 and 30 both read a pattern given after '--' (30 renamed
+    # --testPathPattern), and after '--' a pattern starting with '-' is never
+    # read as options (#121). vitest gets no '--': it would drop the file filter.
+    jest_args     = ['--passWithNoTests', '--no-coverage', '--', os.path.relpath(test_file, root)]
```
alexandrocuma: **Important: the jest pattern is a regex, so the wrong test can run, or none with a false PASS** (same at `hooks/opencode/test-on-save.js:67`)

After `--`, jest reads the relative path as a regex and searches for it anywhere in each test file's absolute path. I tested this with real Jest 29.7.0 and 30.5.0, both with `--passWithNoTests --no-coverage -- <pattern>`:

| test file the hook picks | what jest ran |
|---|---|
| `a+b.test.js` | `ab.test.js` (a different file; its failure is reported) |
| `c(1).test.js` | nothing: `No tests found`, exit 0, so the hook prints **PASS** |
| `mod.test.js` | `mod.test.js` plus `sub/mod.test.js` and `mod.test.jsx` |

`--testPathPattern` behaved the same on Jest 29, so part of this predates the PR. But this PR rewrites this call and describes it as running exactly the intended test, and on Jest 30 the call is only reachable now.

**Fix**: `jest --passWithNoTests --no-coverage --runTestsByPath -- <path relative to root>`. It matches the exact path, not a regex. Verified on 29.7.0 and 30.5.0 (cwd = root): `a+b.test.js`, `c(1).test.js`, `mod.test.js` and `-foo.test.js` each run exactly that one file. Keep the path **relative**: an absolute path through a symlinked root (for example `/tmp` → `/private/tmp`) found no tests. Then update the jest argv in both suites and the `--` sentence in `hook-authoring-guide.md:99`. A regression case with a metacharacter file name (e.g. `a+b.test.js`, with a stub for argv) would pin it.

hooks/claude-code/format-on-save.sh
```diff
@@ -15,9 +15,19 @@ set -euo pipefail
...
 # The trailing "x" keeps $(...) from trimming newlines that belong to the path.
-file_path=$(echo "$input" | python3 -I -c \
-    "import sys,json; d=json.load(sys.stdin); sys.stdout.write(str(d.get('tool_input',{}).get('file_path','')) + 'x')") || {
+file_path=$(echo "$input" | python3 -I -c '
+import sys, json, os
+d = json.load(sys.stdin)
+p = str(d.get("tool_input", {}).get("file_path", ""))
+c = d.get("cwd")
+if p and not os.path.isabs(p):
+    p = os.path.join(c if isinstance(c, str) and os.path.isabs(c) else os.getcwd(), p)
```
alexandrocuma: **Minor: sh does not normalise the resolved path, but opencode does** (same line in `lint-on-save.sh` and `test-on-save.sh`)

`os.path.join` keeps `.` and `..` segments. `path.join` in `editedPath` removes them, which matches how opencode's edit tool resolves the path. So the same relative input gives different argv in the two CLIs. With `cwd=/p` and `./-m.py`, sh passes `/p/./-m.py` and js passes `/p/-m.py`. It also changes behaviour. `find_root` uses `abspath`, which works on the text of the path, but the tool follows the real filesystem. When a `..` segment comes after a symlinked directory, the sh hook points the tool at a different file than the one js picks, from a root that does not contain that file.

**Fix**: in the relative branch only, `p = os.path.normpath(os.path.join(...))`. Absolute input stays byte-identical. I applied this to all three hooks: `on-save-path.test.sh` (66/0), `fail-closed.test.sh` and `interpreter-isolation.test.sh` still pass. Add a `./-mod.py` (or `a/../-mod.py`) case to both suites so the two CLIs stay pinned to the same argv.

alexandrocuma (commented):
## PR Review: fix(hooks): resolve on-save paths to absolute and run jest 29/30 alike (#121)

**Recommendation**: Request changes (two in-diff findings, one Important, one Minor)

**Summary**: The design update holds up. Every format and lint tool now gets an absolute path. Argv for absolute input is byte-identical to v0.8.0 for every non-jest tool, root detection is unchanged, and jest runs on both 29 and 30. The v0.7.1 hardening is intact: `python3 -I`, no tool input in interpreter source, the quoted heredoc, the `x` suffix, argument arrays with no shell, and async `runCommand`. Malformed input still exits 0 with `internal error`. Two problems remain. The jest pattern is still a regex, so the claim that jest runs exactly the intended test is not true in general. And the Claude Code resolution does not normalise the path the way the opencode side does.

### What I verified
- `on-save-path.test.sh` 66/0 and `on-save-path.test.js` 54/0. All other `hooks/claude-code/*.test.sh` and `hooks/opencode/*.test.js` suites pass.
- Mutations, each caught by the suites: sh ignores the input `cwd` (1 fail), sh drops resolution (2), sh drops the jest `--` (6), js ignores `ctx.directory` (4), js drops resolution (21), js reverts jest to `--testPathPattern` (6), js format uses the raw path (10).
- Real jest 29.7.0 and 30.5.0: `-- --help` is taken as a pattern, not an option. No new path lets a leading `-` or `@` reach a tool as an option or argfile. Every tool now gets a path starting with `/`, and go gets `./<dir>`.
- Trust: the input `cwd` comes from the harness, not the model. Resolving against it gives the model no reach it lacked before, since it could already send an absolute path.

### Outside this diff
- `agents/onboarding.md:319` still recommends `--testPathPattern=orders`, which Jest 30 rejects.

hooks/claude-code/test-on-save.sh
alexandrocuma: Fixed in 4d37bc9, following your suggestion.

- **Change:** both hooks now run `jest --passWithNoTests --no-coverage --runTestsByPath -- <path relative to root>`. The same change is in `hooks/opencode/test-on-save.js`, and the path is still relative.
- **Real jest, run directly** (jest 29.7.0 and 30.5.0, results from jest's `--json` output):
  - `a+b.test.js`, `c(1).test.js`, `mod.test.js`, `-foo.test.js` and `-dir/x.test.js` each run exactly that one file, even with failing `ab`, `sub/mod.test.js` and `mod.test.jsx` decoys present.
  - A file excluded by `testPathIgnorePatterns` still reports `No tests found` and exits 0.
- **Through both hooks** (run from another directory, with the input `cwd` / `ctx.directory` set): `a+b`, `c(1)`, `mod`, `-foo` and `a/../-foo` each report `[PASS] Tests: 1 passed, 1 total` on 29 and 30.
- **Tests:** the jest argv in both `on-save-path` suites now includes `--runTestsByPath`. New `a+b.test.js` and `c(1).test.js` cases were added to both suites.
- **Mutation:** dropping `--runTestsByPath` fails 8 cases in each suite.
- **Docs:** the `--` sentence in `hook-authoring-guide.md` now says jest reads a bare path as a regex, which is why the hook adds `--runTestsByPath`. The CHANGELOG bullet is updated.
- **Same topic:** `agents/onboarding.md` recommended `--testPathPattern=orders`, which Jest 30 rejects. It now shows the positional `npm test -- src/orders`.

hooks/claude-code/format-on-save.sh
alexandrocuma: Fixed in 4d37bc9, following your suggestion.

- **Change:** `format-on-save.sh`, `lint-on-save.sh` and `test-on-save.sh` now apply `os.path.normpath(os.path.join(...))` in the relative branch only. Absolute input is untouched, and its argv is still byte-identical.
- **Tests:** both suites now pin the same argv for `./-mod.py` and `a/../-mod.py`:
  - format-on-save and lint-on-save with ruff give `[$P/-mod.py]`;
  - test-on-save with pytest gives `[$P/test_-mod.py]`.
  `a/` exists in the fixture, so a path that isn't normalised would still reach the tool, with different bytes.
- **Mutations:**
  - dropping `normpath` in `format-on-save.sh` fails 2 cases, and in `lint-on-save.sh` 2 cases;
  - replacing `join` with string concatenation in opencode's `editedPath` fails 4 cases.
  - The test-on-save `..` case can't catch a missing `normpath`, because the pytest path is rebuilt from `abspath` either way. It only pins that sh and js match.
- **Other suites:** `fail-closed.test.sh` (10) and `interpreter-isolation.test.sh` (11) still pass, as do all four CI suites.

PR #136, opened 2026-09-16
fix(cli): only accept a devexp-toolkit checkout as the asset root (#134)

### LINKED ISSUE #134: security: repo detection should only accept a devexp-toolkit checkout
Found during review of #131.

## Problem
When no `DEVEXP_DIR` is set and the binary isn't next to a clone, repo detection can accept a directory that is not a devexp-toolkit checkout, and install assets from it without telling the user which root was used.

Details are intentionally kept brief in this public issue.

## Acceptance Criteria
- [ ] Repo detection only accepts a devexp-toolkit checkout, identified by a toolkit-specific marker — not by generic directory names alone
- [ ] Before writing anything, install prints the resolved asset root (clone, `DEVEXP_DIR`, or embedded cache)
- [ ] Tagged release builds don't adopt an arbitrary checkout from the working directory unless `DEVEXP_DIR` is set (or the rule is documented and justified)
- [ ] Tests: a directory with the generic shape but no marker is not used; the real checkout still is; mutation-checked

## Release Targets
- `cli`, `toolkit-clone` — include in the next patch release

alexandrocuma:
Closes #134 · Security

## Summary
Repo detection accepted directories that merely had the same directory names as the toolkit, and install only named its asset root when it used the bundled assets.

- **Marker** — a checkout is identified by the committed `.devexp-toolkit` file (regular file, first line exactly `devexp-toolkit`, ≤ 4 KiB) plus `agents/`, `skills/`, `mcps/`. The marker is staged and embedded, so the extracted bundled assets still qualify; `DEVEXP_DIR` gets the same check.
- **Tagged builds** (release binaries) use `DEVEXP_DIR` or their bundled assets only. **Dev builds** (`./install.sh`, `go run`, `go test`) keep the next-to-binary and walk-up lookups, both requiring the marker. No supported flow places a tagged binary inside a clone, so nothing is lost.
- **Announcement** — install decides the asset root first and prints `Asset root: <dir> (<origin>)` before anything is extracted or written.
- Docs (`setup.md`, `install.md`, `overview.md`, line refs in `release.md` / `testing.md`); CHANGELOG `[Unreleased] ### Security`.

**Upgrade note:** clones and forks created before this change need to pull (to get `.devexp-toolkit`); until then a dev binary falls back to the bundled assets and says so in the `Asset root:` line.

## Out of scope (noted)
- `devexp uninstall` reads `$DEVEXP_DIR/hooks/registry.json` without the marker check; it only narrows which files are removed inside devexp's own opencode plugin directory.

## Test plan
- [x] Go `go vet` + `go test ./... -race -cover` — repo 88.8%, cmd 78.6%; new marker, tagged/dev lookup, announcement-order and end-to-end `install --reinstall-mcps` tests (non-toolkit directory: nothing registered or written, tree byte-identical; real checkout still used)
- [x] Mutation checks: 15/15 caught
- [x] Hook suites; `install.test.sh` 17, `remote-install.test.sh` 24, `uninstall.test.sh` 94 (real-binary round trip ran)
- [x] Built binaries: dev binary via `install.sh` uses the checkout; tagged binary inside the worktree uses bundled assets; `DEVEXP_DIR` to the worktree used; unmarked `DEVEXP_DIR` refused


alexandrocuma (commented):
## PR Review: only accept a devexp-toolkit checkout as the asset root

**Recommendation**: Request changes (one in-diff item, scoped to dev builds)

Release builds are closed off properly: with `DEVEXP_DIR` unset, a tagged binary uses only its bundled assets, wherever it runs from. The asset root is printed before anything is extracted or written, and a `DEVEXP_DIR` without the marker is refused. The open point is the dev-build lookup, which still makes its trust decision based on the cwd (inline).

**Verified** (scratch copy, temporary HOMEs, stub CLI on PATH):
- `go vet`, `go test ./... -race`: green. `install.test.sh` 17, `remote-install.test.sh` 24, `uninstall.test.sh` 94 (real-binary round trip ran).
- Mutations: ignoring the build kind in `locate`, and announcing after extraction, are both caught.
- Tagged binary (`-X devexp/cmd.version=…`): uses the bundled assets and registers nothing from the surrounding tree. `DEVEXP_DIR` works for a clone and for the extracted cache. An unmarked `DEVEXP_DIR` is refused before any write.
- `./install.sh` from a clone (fresh build), run from inside another tree, still uses the clone. `uninstall.sh --yes` round trip is clean.
- goreleaser: `before` runs `stage-assets.sh`, which now stages the marker. The `//go:embed` pattern includes it, and the extracted cache passes `isRepoDir`.

**Done well**
- Splitting `locate` (no writes) from extraction makes the announce-before-write guarantee structural, and the test checks it by asserting the cache is still empty at announce time.
- A refused `DEVEXP_DIR` fails outright with no silent fallback, and its message is clear.
- The end-to-end `TestInstallCmd_AssetRoot` asserts the right things: hook command prefixes, no MCP calls, and a byte-identical tree.

cli/internal/repo/repo.go
```diff
@@ -65,39 +108,62 @@ func absoluteSource(src Source) (Source, error) {
...
+		return "", "", errRepoNotFound
+	}
+	if exe, err := executable(); err == nil {
 		if candidate := filepath.Dir(filepath.Dir(exe)); isRepoDir(candidate) {
-			return candidate, nil
+			return candidate, OriginBinaryDir, nil
 		}
 	}
 	cwd, _ := os.Getwd()
 	for dir := cwd; ; {
```
alexandrocuma: 🟡 **Important — dev builds still decide based on the cwd; the marker identifies a checkout but does not authenticate one**

`.devexp-toolkit` is an ordinary committed file, so any repository can carry it. For tagged builds that doesn't matter, because they never search the filesystem. Dev builds still do: when the next-to-binary lookup misses, the walk-up from the cwd adopts any marked directory as the asset root. The next-to-binary lookup misses whenever a `dev` binary isn't sitting in its own clone's `bin/`, for example when it is copied or installed onto PATH. On macOS it also misses for a symlink onto PATH, because `os.Executable` doesn't resolve symlinks there. From then on, where the user runs `devexp install` decides which hooks and scripts get registered globally. So the property the CHANGELOG and `install.md` describe ("only accepts a devexp-toolkit checkout") holds for release binaries, but for dev builds it only holds for a directory without the marker.

**Suggested fix:** anchor dev builds to the checkout they were compiled from, not to the cwd:
- Derive the source root from `runtime.Caller(0)` in this file. The path is absolute for `go build`, `go run` and `go test`; under `-trimpath` it is module-relative, so fall back to the bundled assets. Accept that root only if `isRepoDir` passes.
- Drop the cwd walk-up. With a compile-time root it isn't needed for `go run` or `go test` inside the clone.
- Optionally `filepath.EvalSymlinks` the executable in the next-to-binary lookup, or drop that lookup too, since the compile-time root covers `./install.sh`'s `bin/devexp`.

Then update the `"dev build, a devexp-toolkit checkout"` case in `TestInstallCmd_AssetRoot`, which currently asserts that a marked temp tree around the cwd is adopted. Also state the property in CHANGELOG, `install.md` and `setup.md` as: release builds never adopt a directory found on disk, and dev builds use only their own source checkout. Don't present the marker as the protection.

cli/cmd/install.go
```diff
@@ -200,3 +198,13 @@ func runInstall(cmd *cobra.Command, args []string) error {
 	fmt.Printf("\033[0;32m\033[1mAll done.\033[0m\n\n")
 	return nil
 }
+
+// announceAssetRoot tells the user which directory devexp installs from and
+// how it was chosen (#134).
+func announceAssetRoot(src repo.Source) {
```
alexandrocuma, line 205: 🔵 **Minor — a clone or fork without the marker falls back silently**

When a dev binary's own clone lacks `.devexp-toolkit` (an old fork, or the file was dropped), install prints `Running standalone — using assets bundled in this binary.` but doesn't say why. A contributor editing assets in that clone gets the bundled copies instead. Those can also be stale, because the extraction cache is reused while `.devexp-version` equals the binary version, and every dev build's version is `dev`. That part predates this PR, but this PR adds a new way for a dev build to end up in that cache.

**Suggested fix:** when the directory the dev build would have used has `agents/`, `skills/` and `mcps/` but no valid marker, `ui.Warn` that it was skipped for lacking `.devexp-toolkit` (pull, or restore the file). With the compile-time-root approach from the other comment, that check is on the source root. You could also treat `dev` as never matching the cache version, so dev builds always re-extract.

cli/internal/repo/repo.go
alexandrocuma: Fixed in 031fffa. Dev builds are now bound to the checkout they were compiled from, and no directory on disk is searched for any more:

- `Resolve` uses `DEVEXP_DIR`, then (dev builds only) the source root derived from `runtime.Caller(0)` in `repo.go` if it passes `isRepoDir`, then the bundled assets. Under `-trimpath` the recorded path is module-relative, so there is no source root and the bundled assets are used.
- The cwd walk-up and the next-to-binary lookup are removed. `./install.sh`'s `bin/devexp`, `go run` and `go test` are all covered by the compile-time root. The wizard's Remove runs `uninstall.sh` from that resolved root only.
- Tests: `TestSourceRootOf` (absolute, `-trimpath`, relative and other-path cases), `TestSourceRoot_RealCheckout`, and `TestLocate_NoDevexpDir`, which runs every build/source-root case from inside other checkouts (root, `bin/`, subdirectory) and from outside. `TestInstallCmd_AssetRoot` now asserts that a dev build uses its own source checkout and a tagged build its bundled assets, whatever tree it runs in. I mutation-checked re-adding the walk-up, accepting a marked cwd, treating a missing source root as the cwd, accepting a relative source path and skipping the marker check on the source root; each one is caught.
- I also built real dev binaries (copied, symlinked, placed in another tree's `bin/`, and `-trimpath`) and ran them from inside another marked tree. Each used its own source checkout or its bundled assets, and nothing from the tree was registered.
- CHANGELOG, `install.md`, `setup.md` and `overview.md` now state the rule: release builds use only `DEVEXP_DIR` or bundled assets, and dev builds use only `DEVEXP_DIR`, their own source checkout or bundled assets. The marker is described as identifying a checkout, not as the protection.

cli/cmd/install.go
alexandrocuma: Fixed in 031fffa.

- When a dev build's source checkout has `agents/`, `skills/` and `mcps/` but no valid `.devexp-toolkit`, install now warns that the checkout was skipped and says to pull (or restore the file). The warning is printed before the asset root line, ahead of any extraction. There is a matching warning when the source checkout no longer exists, which says to rebuild with `rm bin/devexp && ./install.sh`.
- `extractEmbedded` no longer reuses the cache for version `dev`, so dev builds re-extract on every run.
- Tests: `TestLocate_NoDevexpDir` (marker missing, invalid marker, checkout gone), `TestResolve` (the warning reaches `announce`), `TestAnnounceAssetRoot` (warning shown, and shown before the root line) and `TestExtractEmbedded` ("a dev build never reuses an extraction"). I mutation-checked dropping each warning, not passing the warning on, printing it after the root line and reusing the cache for dev; each one is caught.

alexandrocuma (commented):
## Focused re-review: 031fffa (how the asset root is chosen)

I built real binaries: a dev build, a `-trimpath` dev build, a tagged build (`-X devexp/cmd.version=…`) and a tagged `-trimpath` build. I also tried `go run`, `./install.sh` from a fresh clone, cgo, a `-gcflags=-trimpath` build and a build through a symlinked path. Each ran against a temp HOME with a stub `claude`. The rules hold: tagged builds only use `DEVEXP_DIR` or their bundled assets, and dev builds only use `DEVEXP_DIR`, the checkout they were compiled from, or their bundled assets. That stays true wherever the binary is run from or copied to. The warnings for a moved checkout and a checkout without the marker fire with clear text. I also deliberately broke each rule in the code (version gate, marker check, absolute-path check, path-boundary check, no cache reuse for dev builds, cwd walk-up, missing-checkout warning, warning display), and each break makes at least one test fail. Nice work.

### 🟡 Important

**1. The compile-time checkout is trusted by its path alone** (`cli/internal/repo/repo.go:171`)
`sourceCheckout` only checks what is at the recorded path when the binary runs: the layout and the public marker. If the original checkout is removed and something else later appears at the same path, a dev binary that outlived its checkout uses the new directory without any warning. The CHANGELOG supports keeping such binaries ("a copied or linked dev binary still uses its own checkout"). The marker can't tell the two apart because anyone can copy it.
*Suggested fix:* before using the checkout, also require that its root (and `hooks/`) is owned by the current user and is not writable by group or others. Otherwise warn and fall back to the bundled assets, the same way the missing-marker case does. Add a `TestLocate_NoDevexpDir` case for it.

### 🔵 Minor

**2. Tests assume builds without `-trimpath`** (`cli/internal/repo/repo_test.go:212`, `cli/cmd/install_test.go:2397`)
With `GOFLAGS=-trimpath go test ./...`, `TestSourceRoot_RealCheckout` and `TestInstallCmd_AssetRoot` fail, because the recorded path is module-relative. CI doesn't set the flag, but a developer who does gets failures that look like a regression. *Fix:* skip these tests, or assert the embedded fallback instead, when `runtime.Caller` returns a path that isn't absolute.

**3. Dev and tagged builds share one extraction dir, and a dev run replaces it in place, not atomically** (`cli/internal/repo/repo.go:243-247`)
Every dev run that falls back to the bundled assets runs `RemoveAll` and then re-extracts into the same `devexp/assets` dir that a release install's registered hooks point into. Until the extraction finishes, those hooks are missing, and afterwards they hold the dev binary's copies. Concurrent runs also race on the same dir (6 parallel runs didn't fail for me). Everything happens under the same user, so this isn't a trust issue. *Fix:* give dev builds their own dir (for example `devexp/assets-dev`), or extract to a temp sibling and rename it into place.

**4. The CHANGELOG security entry doesn't tell clone users to rebuild** (`CHANGELOG.md:134`)
`install.sh` never rebuilds an existing `bin/devexp`. A binary built before this change keeps the old lookup until it is rebuilt. Other entries in this release say "Clone users: run `rm bin/devexp && ./install.sh`", and this entry should say the same.

**5. Nit: goreleaser doesn't pass `-trimpath`** (`.goreleaser.yaml:22`)
Release binaries record the CI checkout path. I checked a tagged build without `-trimpath`: it ignores that path because of the version check, so this is not a bug. Adding `flags: [-trimpath]` would add a second layer of protection and make builds more reproducible.

alexandrocuma:
Re: [focused re-review](https://github.com/alexandrocuma/devexp-toolkit/pull/136#pullrequestreview-5229578718). All five points are fixed in df873bb, and CI is green.

**1. The compile-time checkout is trusted by its path alone.** A dev build now uses its recorded checkout only if `ownedCheckout` can verify it as the user's (Unix):
- With symlinks resolved, the checkout and its `agents/`, `skills/`, `mcps/`, `hooks/`, `hooks/claude-code/` and `hooks/opencode/` must be real directories (not symlinks), owned by the current user and not writable by group or others.
- The directory holding the checkout must be owned by the user or by root, and not writable by group or others unless it has the sticky bit. If the recorded path is itself a symlink, the link must be owned by the user or by root, and its directory must pass the same check.
- Otherwise install warns that "the checkout this binary was built from … can't be verified as yours", gives the reason and how to fix it, and uses the bundled assets. On platforms without file ownership the check always fails, so those builds use the bundled assets.
- `TestLocate_NoDevexpDir` covers each rule through an owner seam and real `chmod`s: another user owns the root, `hooks/claude-code/` or `agents/`; the root, `hooks/` or `mcps/` is writable by group or others; `hooks/` is a symlink; the parent is writable, owned by another user, or sticky but owned by another user; the recorded path is a symlink in a writable directory or owned by another user; ownership is unavailable. Positive cases: a normal checkout, one in a sticky directory you own, and one reached through your own symlink.
- A real dev binary whose checkout was replaced at the same path falls back with the warning when the parent, root or `hooks/` is writable by group or others.
- 11 mutations are all caught: dropping the check, each owner, permission, parent, sticky and symlink rule, only checking the root, and accepting an unknown uid.

**2. Tests assume builds without `-trimpath`.** `TestSourceRoot_RealCheckout` and `TestInstallCmd_AssetRoot` now expect the bundled assets when the recorded path isn't absolute. `GOFLAGS=-trimpath go test ./...` passes, and it fails again if either change is reverted. `TestInstallCmd_AssetRoot` also accepts the "can't be verified" fallback, so a checkout cloned with a group-writable umask doesn't cause spurious failures.

**3. Shared, non-atomic extraction dir.** Dev builds now extract to `devexp/assets-dev`, and release builds keep `devexp/assets`.
- For both, the directory is a symlink to a complete extraction. Each extraction goes into a fresh sibling, which a symlink rename swaps in. A failed extraction is removed and leaves the previous tree as it was. An in-place directory left by an older binary is moved aside and replaced.
- Leftovers from killed runs, or from runs that lost a race, are swept once they are an hour old. The tree currently in use is never swept.
- Tests cover separate dirs (a dev run leaves the release extraction alone), an interrupted extraction (the previous tree is intact and nothing is left behind), replacing an in-place directory, 8 concurrent runs for dev and tagged builds (a complete tree and no partial leftovers), and the sweep (old leftovers removed, recent ones and the current tree kept).
- 9 mutations are caught: a shared dir, extracting in place, no cleanup on failure, keeping the previous tree, a sweep that ignores age or the current tree, no sweep, and not moving an in-place dir aside. Writing the version file before the tree is not detectable, since only a successful extraction is ever swapped in.
- With real binaries, six parallel dev runs all succeeded and left no partial trees. Every hook a release install registered still resolved after a dev run.

**4. CHANGELOG.** The entry now says: "Clone users: run `rm bin/devexp && ./install.sh`", and explains that a binary built before this change keeps the old lookup until it is rebuilt.

**5. goreleaser `-trimpath`.** `flags: [-trimpath]` is added. `goreleaser check` (v2.18.1) validates the config, and a `goreleaser build --snapshot --single-target` binary reports `devexp version 0.8.0-SNAPSHOT-…` and contains no build-path strings. `release.md` now mentions the flag and its line references are updated.

Docs (`setup.md`, `install.md`, `overview.md`) describe the ownership rule and the extraction layout.

alexandrocuma (commented):
## Final focused review: `df873bb` together with #139 (head `af1209f`)

I built real binaries (a dev build, a `-trimpath` tagged build, and v0.8.0), ran installs in temporary HOMEs on macOS and in a Linux container as a non-root user, and ran each registered hook command through `sh -c` before and after every extraction. Nothing here blocks merge. The asset swap works, and the hook commands keep working through it.

### What checks out
- **Hook commands and the symlinked cache.** Commands are still registered as `<cache>/devexp/assets/hooks/claude-code/<script>`, so the link path is what gets stored. v0.8.0 → this build (in-place directory → link): `settings.json` is byte-identical, and 7/7 hooks run before and after, on macOS and Linux. That also holds for a downgrade back to v0.8.0 (its `RemoveAll` takes out only the link) and for upgrading again. A hook that was already running during a re-extraction finished normally (exit 0). Claude Code sessions that were already open keep working, since the stored path doesn't change.
- **#126/#139 matchers.** `isStaleDevexpHook`, the foreign/relative pruning and the quoting logic compare paths as strings and `os.Stat` through the link, so the link is transparent to them. Moving between a release install (`assets`) and a dev fallback (`assets-dev`) removes the other root's registrations, as intended. `uninstall.sh` run from the linked cache removes every hook.
- **Swap safety.** New directories get mode `0755` (`MkdirTemp` plus `Chmod`). The previous tree is removed only when the link target is a bare `.<name>.` sibling, so an absolute or `../` target is ignored. The `.assets.` and `.assets-dev.` prefixes can't match each other. Linux and macOS agree on rename-over-link and on the migration path. A 3-way concurrent stress test (120 runs) always left `assets` pointing at a complete tree with a matching version file.
- `go test ./...` passes normally and with `GOFLAGS=-trimpath`. `go vet` is clean. goreleaser `builds[].flags: [-trimpath]` is valid.

### 🟡 Important
**1. The ownership rule rejects clones made under a group-writable umask (`repo.go` ~260, ~277).** Several Linux distributions give users a private group and a default umask of `002`. A clone made there has directories with mode `775`, so the dev build warns and falls back to its bundled assets (confirmed in a container). The documented contributor loop ("edit `agents/` → `./install.sh`") then deploys the copy embedded when `bin/devexp` was built, not the edits. The warning does name the fix. Consider allowing group write when the directory's group is the user's own private group (gid equals the user's primary gid and the group name equals the user name). Otherwise, call out this case in the install and setup docs.

### 🔵 Minor
**2. Removing the previous tree immediately can fail hook invocations that start at that moment (`repo.go` ~403–405).** In a stress run on macOS, hook invocations that started while the old tree was being deleted were killed (exit 137): 30 of 623. In a minimal reproduction, keeping the old tree for a while brought that to 0. A PreToolUse guard that dies this way counts as a non-blocking error. This only happens during a re-extraction, and the old in-place extraction had a wider window of the same kind, so this is better than before. Suggestion: don't delete the previous tree inline. Refresh its mtime (`os.Chtimes`) and let `sweepStale` remove it after a short grace period, and run the sweep on the reuse path too.

**3. Only directories are checked for ownership, not the files inside them (`repo.go` ~198–263).** A checkout whose hook scripts or `hooks/registry.json` are writable by others, or owned by another user, is still accepted (confirmed). Directories that are the user's and not group/other-writable already cover the case the rule targets: a path that someone else re-created. The check also runs only at install time, while hooks keep running from the checkout afterwards. A file-level walk isn't required. If you want the rule to be consistent, one pass over `agents/ skills/ hooks/ mcps/` that rejects entries not owned by the user, group/other-writable, or symlinks is cheap.

**4. A swept in-progress extraction can be recreated partially and swapped in (`repo.go` ~428, ~471/482).** If a run is paused for more than `staleAfter` in the middle of extracting (for example, a laptop that sleeps), another run's sweep deletes its tree. `extractFS` then recreates the tree root and parents through `MkdirAll`, writes the remaining files and the version file, and swaps in an incomplete tree, which a tagged build keeps reusing. This is unlikely. A fix is to create directories with `os.Mkdir` (WalkDir visits parents first) so extraction fails once the root is gone.

**5. The migration failure path deletes the only complete tree (`repo.go` ~389–400).** If the final `Rename(link, dest)` fails after the in-place directory was moved aside, the deferred `RemoveAll(old)` deletes it and `dest` stays missing until the next run. On failure, rename it back before returning.

**Nit.** `sweepStale` runs only after an extraction, so trees left by runs that lost a race (about 1 MB each) stay until the next version change. These are hidden `.assets.*` directories, and `uninstall.sh` doesn't clean the cache. That's unchanged behaviour, but they're easier to miss now.

alexandrocuma:
Re: [final review](https://github.com/alexandrocuma/devexp-toolkit/pull/136#pullrequestreview-5229788359). All five points are fixed in 18a7bef, and CI is green.

**1. Group-writable umask.** Group write is now accepted on a checked entry, and on the directory holding the checkout, when its gid is the user's primary gid and that group is named like the user (the user-private-group convention, looked up with `os/user`). If a lookup fails, the entry is rejected. Other-write is always rejected, and the sticky rule for the parent is unchanged.
- Tests use a private-group seam: a checkout, hook script or parent that is group-writable through the private group is used. Group write through a shared group, group write when the lookup fails, and other-write even with a private group all fall back with the warning. `TestLookupPrivateGroup` compares the lookup against `os/user`.
- In an Alpine container with a private-group user and umask 002, a 775 clone is used (7/7 hooks registered from it). The same clone owned through a shared primary group falls back with the warning, and is used again once group write is removed.
- The repo tests now pin umask 022, so they pass under a caller's umask 002 (checked on Linux and macOS).
- `setup.md`, `install.md`, `overview.md` and the CHANGELOG describe the rule.

**2. Removing the previous tree immediately.** A swap no longer removes the tree it replaced. Its mtime is refreshed, and `sweepStale` removes trees the link doesn't point at once they are older than an hour. The sweep now also runs when an existing extraction is reused. An in-place directory moved aside during migration is retired the same way.
- I kept the one-hour grace period: on-save hooks can run test suites for minutes, and a retired tree costs about a megabyte.
- Tests: the retired tree survives the swap with a fresh mtime, survives a reuse within the grace period, and is swept by a reuse after it.
- Your `stress.sh` (40 rounds × 3 concurrent re-extracting installs, with a guard hook running in a loop): this build had **0 failures in 668 hook runs**. The previous head, built the same way, had 25 failures (rc=137) in 693. `runhooks.sh` shows 7/7 hooks before and after. After aging, one more install left only the current tree.

**3. Files not ownership-checked.** `ownedCheckout` now walks `agents/`, `skills/`, `mcps/` and `hooks/`, and also checks `.devexp-toolkit`, `devexp.config.json` and `uninstall.sh`. It rejects any entry that isn't owned by the user, is a symlink, is writable by others, or is group-writable other than through the private group. The tree is about 110 entries.
- Tests: a hook script writable by others, a hook script writable by a shared group, `hooks/registry.json` owned by another user, an agent file owned by another user, a symlinked skill file, an MCP registry writable by group, `uninstall.sh` owned by another user, `devexp.config.json` writable by others, and the marker owned by another user.

**4. Swept in-progress extraction rebuilt.** `extractFS` now creates directories with `os.Mkdir` (WalkDir visits parents first) and writes files without creating parents. Once its tree is removed, extraction fails. The deferred cleanup runs and nothing is swapped in.
- Test through the `extractTree` seam: the tree is removed mid-walk, before either a directory or a file comes next. The result is an error, the removed tree is not recreated, and the previous tree is intact.

**5. Failed migration.** If the final rename fails after the in-place directory was moved aside, it is renamed back and the temporary directory is removed.
- Test through a `rename` seam: the in-place tree is back with its content, and nothing is left behind.

**Mutations.** All 22 are caught:
- 7 for the private-group rule: any group write, gid not compared, lookup failure accepted, other-write allowed, private group refused, parent using the old rule, group-name check dropped.
- 6 for file checks: only directories checked, symlinks allowed, root files unchecked, `hooks/` not walked, only top-level directories walked, marker unchecked.
- 4 for retention: retired tree removed inline, mtime not refreshed, no sweep on reuse, moved-aside directory removed inline.
- 3 for `Mkdir`: `MkdirAll` for directories, file writes creating parents, root recreated.
- 2 for move-back: none, or the temporary directory left behind.

**Your nit.** Retired and leftover trees are now swept on reuse as well, so they no longer wait for a version change.

Issue #138, opened 2026-09-16
Stale hook pruning removes user commands under the repo dir


alexandrocuma:
Found during #135 (out of its scope).

## Problem
`isStaleDevexpHook` treats any registered command string under the devexp repo dir whose raw string isn't an existing file as a stale devexp hook, so a user's own hook such as `<repo>/hooks/x.sh --flag` is pruned on install.

## Acceptance Criteria
- [ ] Only commands matching devexp's own hook form (plain or canonically quoted path under `hooks/claude-code/`, per #126/#135 matchers) are considered stale
- [ ] User commands under the repo dir (with arguments, other dirs, shell syntax) are never pruned
- [ ] Tests incl. mutation check

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
Groomed together with #137 (both change `cli/internal/hooks/installer.go`; delivered on one branch, separate commits). Verified execution plan: https://github.com/alexandrocuma/devexp-toolkit/issues/137#issuecomment-5706992964 — see "Commit 2". Correction: `isManagedScriptPath` can't be reused as-is for staleness (a stale script's basename is no longer in the registry); the devexp form is a `commandPath`-form, clean, absolute `<repo>/hooks/claude-code/<name>`. `uninstall.sh` has no stale-pruning step, so nothing changes there for this ticket.

Issue #137, opened 2026-09-16
Rewriting settings.json drops extra fields from users' hooks


alexandrocuma:
Found during #135 (out of its scope).

## Problem
When devexp rewrites `~/.claude/settings.json` (install or uninstall), its hook entry types keep only `matcher`, `type` and `command`. Any other field on a user's own hook entries — e.g. `timeout`, `args`, `shell`, or fields added by future Claude Code versions — is silently dropped. A hook relying on `args` then switches to shell form. This breaks the guarantee that user hooks are left untouched.

## Acceptance Criteria
- [ ] Install and uninstall preserve every field of hook entries and hook commands devexp doesn't own, byte-for-byte where practical (unknown fields round-trip)
- [ ] devexp-owned entries still written as today
- [ ] Tests with user hooks carrying `timeout`, `args`, `shell` and an unknown field, through install, re-install and uninstall
- [ ] Mutation-checked

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
# Execution plan — #137 + #138 (one branch: `fix/137`)

## Ticket Health Report — #137 / #138

**Tickets**: #137 "Rewriting settings.json drops extra fields from users' hooks" · #138 "Stale hook pruning removes user commands under the repo dir"
**Type**: Bug fix (both)
**Groomed**: 2026-09-16
**Validated against**: commit `43dc1d0` (`fix/137`, from `origin/main`, v0.9.0)

### Schema evidence (Claude Code hooks reference, code.claude.com/docs/en/hooks)

- Matcher group (`hooks.<Event>[]`): `matcher` (optional — omitted = match all), `hooks` (required array).
- Every handler: `type` (`command` | `http` | `mcp_tool` | `prompt` | `agent`), `if`, `timeout`, `statusMessage`, `once`.
- `command`: `command`, `args` (array — when present `command` is **spawned directly, no shell**), `async`, `asyncRewake`, `shell` (`bash` | `powershell`, ignored with `args`).
- `http`: `url`, `headers`, `allowedEnvVars`. `mcp_tool`: `server`, `tool`, `input`. `prompt` / `agent`: `prompt`, `model`.
- Top-level hook-related keys: `disableAllHooks`, `allowedHttpHookUrls`, `httpHookAllowedEnvVars`, `allowManagedHooksOnly` — plus every other settings key (`permissions`, `env`, `model`, …) devexp never interprets.

### Confirmed claims
[...]

PR #139, opened 2026-09-16
fix(hooks): quote hook commands whose path needs it (#135)

### LINKED ISSUE #135: Hook commands with spaces or shell metacharacters in the path
Found during review of #131.

## Problem
Claude Code runs each registered hook command through a shell, and devexp writes the absolute script path unquoted. When the repo or asset-cache path contains a space or shell metacharacters, the command is split or interpreted: the hook fails to run (Claude Code treats that as non-blocking, so a guard fails open) or a different program runs.

## Acceptance Criteria
- [ ] Registered hook commands run the intended script for paths containing spaces and shell metacharacters (quote the path, or refuse such paths with a clear message — decide and document)
- [ ] If quoting: every matcher that recognises devexp hooks (install prune/foreign/relative, `uninstall.sh`) handles the quoted form, and existing unquoted entries are migrated
- [ ] Tests with a repo path containing a space and a metacharacter, executing the registered command through `sh -c`; mutation-checked

## Release Targets
- `cli`, `toolkit-clone` — include in the next patch release

alexandrocuma:
Closes #135

## Summary
Claude Code runs a command hook through `sh -c`. devexp wrote the absolute script path unquoted, so a repo or asset-cache path containing a space or shell syntax was split or interpreted: the hook didn't run (a non-blocking error, so a guard failed open) or a different program ran.

- **Quote only when needed** — a registered path is POSIX single-quoted exactly when it contains a character from the shell-syntax set introduced in #126. Paths that already worked are written byte-for-byte as before, so existing settings are unchanged and older binaries / `uninstall.sh` still recognise them.
- **One recognised form** — a devexp hook command is either a plain path without shell syntax, or exactly one single-quoted absolute path that re-quotes to the identical string, ending in `/hooks/claude-code/<registry script>`. Double quotes, arguments, chains, other escape spellings and quoted relative paths remain the user's. Applied consistently to install-time duplicate/stale/foreign/relative matching and `uninstall.sh`.
- **Migration** — an existing unquoted (or non-canonically quoted) entry for a registry script under the repo dir is rewritten in place on the next install, disabled hooks included; user commands untouched.
- Exec form (`args`) was not used: no documented minimum Claude Code version, and clients unaware of it fall back to the shell.
- Docs: `install.md` (new "Hook commands (Claude Code)" section), `hook-authoring-guide.md`; CHANGELOG `[Unreleased] ### Fixed`.

## Out of scope (filed)
- #137 — rewriting `settings.json` drops extra fields (`timeout`, `args`, `shell`, …) from users' hooks
- #138 — stale hook pruning can remove user commands under the repo dir
- An unquoted entry from a *different* install root whose path needs quoting can't be distinguished from a user command with arguments; an install or uninstall from its own root cleans it up (documented)

## Test plan
- [x] `TestInstallClaude_PathsNeedingQuotes` — repo dirs with space, `'`, `$`, `;`, `&` and combined; each registered command executed via `sh -c` from an unrelated directory with decoy programs at the truncated paths: only the intended scripts run (decoys run with quoting removed)
- [x] Matcher, migration, quoting-only-when-needed tests; `uninstall.test.sh` 101 (removes quoted and old unquoted devexp entries, keeps 11 user forms)
- [x] Mutation checks 13/13 (Go and `uninstall.sh`)
- [x] Manual e2e: install from `…/My Proj/it's $x;&clone` → registered secret-guard via `sh -c` blocks `.env` (exit 2); `uninstall.sh --yes` removes all 7 devexp hooks, keeps user hooks
- [x] Go `go test ./... -race -cover`, all hook suites, `install.test.sh`, `remote-install.test.sh`


cli/internal/hooks/installer.go
```diff
@@ -185,7 +186,7 @@ func InstallClaude(registry Registry, repoDir, settingsPath string, disabled []s
 		alreadyIn := false
 		for _, e := range hooksMap[cc.Event] {
 			for _, h := range e.Hooks {
-				if h.Command == scriptAbs {
+				if h.Command == command {
```
alexandrocuma, line 189: **Important: a user's own double-quoted form of this repo's script now gets a second copy that also runs**

`alreadyIn` and `requoteDevexpHooks` only accept the canonical single-quoted form. The natural manual fix for #135 is `"<repo>/hooks/claude-code/format-on-save.sh"`. For that entry, this install finds no match and appends `'<repo>/…/format-on-save.sh'`. The two strings differ, so Claude Code doesn't dedupe them and both run, in parallel.

The old binary appended the unquoted command in the same situation, which never ran. I checked both binaries against the same settings (with `set -- $cmd`): old gives `"…"` RUNS plus bare path no-run, this PR gives `"…"` RUNS plus `'…'` RUNS. Guards running twice is harmless. `format-on-save` and `test-on-save` running twice at once on the same file is not.

**Fix**: treat exactly `"` + `filepath.Join(repoDir, script)` + `"` as one more legacy spelling, but only for this repo dir and only when the path has none of `$`, backquote, `\`, `"`. With that restriction the double quotes are literal. Either rewrite it to `hookCommand(...)` like the unquoted form (and match it in `uninstall.sh`'s `legacy` set), or at least count it as `alreadyIn`. Add a migration test case for it.

cli/internal/hooks/installer.go
```diff
@@ -397,3 +460,46 @@ func pruneForeignDevexpHooks(hooksMap map[string][]hookEntry, registry Registry,
...
+	for event, entries := range hooksMap {
+		for i := range entries {
+			for j, h := range entries[i].Hooks {
+				p := h.Command
+				if _, legacy := want[p]; !legacy {
+					if p, _ = commandPath(h.Command); p == "" {
+						continue
+					}
+				}
+				cmd, ok := want[p]
```
alexandrocuma, line 510: **Minor: rewriting can leave two identical entries**

If an event already holds the canonical command and also a legacy spelling of it, the legacy one is rewritten in place and becomes an exact duplicate. After that, `alreadyIn` reports "already registered" and nothing ever merges them.

This happens when an older devexp runs against the same repo or cache dir after this one, for example `git checkout <older tag> && ./install.sh` and back again. The older binary doesn't recognise the quoted form and appends the unquoted one. I reproduced it with the `origin/main` and PR `InstallClaude` on one settings file: new, old, new, new leaves two identical `'…/secret-guard.sh'` entries. Claude Code dedupes identical commands at runtime, so the only effect is a messy `settings.json`.

**Fix**: when `cmd` already exists in `hooksMap[event]`, drop this hook instead of rewriting it (and drop the entry if it ends up empty).

cli/internal/hooks/installer_test.go
```diff
@@ -869,3 +871,309 @@ func TestLoadRegistry_RepoRegistry(t *testing.T) {
 		}
 	}
 }
+
+// ── Paths that need shell quoting (#135) ─────────────────────────────────────
+
+// needsQuoting are repo dir names with each kind of character Claude Code's
+// shell (sh -c) would split on or interpret.
+var needsQuoting = map[string]string{
```
alexandrocuma, line 881: **Minor: the tests pin only part of the character set, and nothing checks Go and `uninstall.sh` agree**

`needsQuoting` and `TestHookCommand` cover space, `'`, `$`, `;`, `&`, backquote, `(`, `)` and `\`. Mutation results:
- In Go, removing any of `| < > ( ) * ? [ ] { } ! # " \t \n \r` from `shellSyntax` still passes `go test ./internal/hooks/`. Removing `(` still passes because `)` remains in the set and `"everything"` contains both.
- In `uninstall.sh`, removing `*`, `|`, `>` or `#` from `SHELL_SYNTAX` still passes `uninstall.test.sh`.

A narrower set brings the bug back for those characters, and some of them are worse than a split word: a redirection or a pipeline instead of the script. The Python copy can also drift from the Go one without any test failing.

**Fix**: make `TestHookCommand` loop over every rune of `shellSyntax`: each path must be quoted and must run through `sh -c`. Assert that some near-misses stay plain (for example `%`, `=`, `,`, `@`, `+`). Add a parity check that the `SHELL_SYNTAX` literal in `uninstall.sh` equals `shellSyntax`, for example `uninstall.test.sh` extracting both, or a Go test reading `../../../uninstall.sh`.

docs/guides/install.md
```diff
@@ -66,7 +66,7 @@ Re-running the installer is how you update devexp — there's no separate "upgra
 - **Agents and skills** are overwritten in place with the versions shipped in the new release. Before overwriting, the Claude Code install backs up your existing `~/.claude/agents/*.md` and `~/.claude/skills/<name>/` directories into a timestamped `~/.claude/.devexp-backup-<timestamp>/` folder (`cli/cmd/backup.go`, `cli/cmd/paths.go`). The opencode install makes **no backup** of `~/.config/opencode/agents/` or `commands/` (`cli/cmd/install_opencode.go`).
 - **MCP server registrations** are *not* refreshed automatically — pass `--reinstall-mcps` if an MCP's config (command, args, env) changed in the new release.
 - **Hooks**:
-  - Claude Code: new hooks in `hooks/registry.json` are added to `settings.json`; hooks already registered are left as-is, including ones disabled since.
+  - Claude Code: new hooks in `hooks/registry.json` are added to `settings.json`; hooks already registered are left as-is, including ones disabled since. The one exception is the command string: an entry an earlier install wrote from this repo/cache dir unquoted, for a path that needs quoting, is rewritten in place to the quoted form (see [Hook commands](#hook-commands-claude-code)).
```
alexandrocuma: **Minor (docs)**: this line (and the CHANGELOG entry) says the only rewrite is "unquoted, for a path that needs quoting → quoted". `requoteDevexpHooks` also rewrites a single-quoted own-root path that needs no quoting to the plain form, and `TestInstallClaude_RequotesQuotedPlainPath` tests it. That user-visible rewrite should be mentioned too, e.g. "…rewritten in place to the form devexp writes (quoted only when the path needs it)".

alexandrocuma (commented):
## PR Review: fix(hooks): quote hook commands whose path needs it (#135)

**Recommendation**: Request changes (small ones; the core fix is sound)

The quoting is correct. `shellQuote` plus "quote only when the path has a `shellSyntax` character" ran the intended script, with the exact `$0`, for 53 repo dir names under `sh`, `bash`, `bash --posix`, `dash`, `zsh`, `zsh -o extendedglob` and `zsh -o rcquotes`. The names covered every character in the set, plus characters outside it that stay unquoted: `\v`, `\f`, NBSP, U+2003, U+3000, U+0085, `%`, `^`, `=`, `,`, `:`, `@`, `+`, `-`, leading and trailing space, and control bytes. The Go `commandPath`/`isManagedScriptPath`+legacy and the `uninstall.sh` `command_path`/`is_devexp_hook` gave identical results on about 286k fuzzed commands. Every command classified as devexp's parses to exactly one word, the decoded path, in all four shells. Install from a path with space, `'`, `$`, `;`, `&`, `|` and `>`: the real secret-guard blocks `.env` with exit 2 via sh, bash and zsh. A second install writes nothing, and `uninstall.sh --yes` removes all 7 hooks and keeps the user's hook.

Inline findings, most important first:
1. **Important**: a user's own double-quoted spelling of this repo's script now gets a second copy that also runs.
2. **Minor**: the rewrite can leave two identical entries.
3. **Minor**: the tests pin only part of `shellSyntax`, and nothing checks the Go and Python sets match.
4. **Minor (docs)**: the migration description leaves out one of the two rewrites.

Out of scope and already there before this PR (not a quoting issue): `json.Marshal` turns invalid UTF-8 into U+FFFD. A repo or cache dir whose name isn't valid UTF-8 (possible on Linux; macOS refuses to create one) is registered as a different path. The hook then doesn't run, and every install prunes it as foreign and adds it back. Worth its own ticket if Linux matters.

cli/internal/hooks/installer.go
alexandrocuma: Fixed in 3a68c1e.

`requoteDevexpHooks` now builds an exact legacy set for each registry script under this repo dir. It holds the bare path, plus `"` + path + `"` when the path has none of `$`, backquote, `\` or `"` (`doubleQuoted`), so the double quotes are literal. A matching entry is rewritten in place to `hookCommand(...)`, which also makes `alreadyIn` match. `uninstall.sh` adds the same double-quoted spelling to its `legacy` set.

Any other double-quoted command stays the user's: other directories, other scripts, arguments, or a path holding `$`, backquote or `\`.

Tests:
- `TestInstallClaude_DoubleQuotedHandFix` covers a path that needs quoting and one that doesn't, a disabled hook, idempotence on a second run, and running every command through `sh -c` (each script runs exactly once per event; the user's `"…" --strict` and the other root's copy run separately).
- `TestInstallClaude_DoubleQuotedWithShellSpecials` covers `$`, backquote and `\` (left as the user's).
- `uninstall.test.sh` removes the double-quoted spelling for a plain path and for one that needs quoting (disabled included), and keeps arguments, another root, another script, and paths with `$`/backquote/`\`.

Mutations killed: no double-quoted spelling (Go and Python), and dropping the `$`/backquote/`\`/`"` restriction (Go and Python).

cli/internal/hooks/installer.go
alexandrocuma: Fixed in 3a68c1e.

Before rewriting, `requoteDevexpHooks` now collects the commands already in the event. If the rewritten command is already there, the legacy spelling is dropped instead of rewritten. An entry left with no commands is removed, and other commands, entries and matchers are kept. A second legacy spelling of the same script in the same event is dropped as well. The same command under another event is a separate registration, so it is rewritten, not dropped.

Tests:
- `TestInstallClaude_CollapsesLegacyDuplicates` replays new → old (a stand-in for the pre-#135 append of the bare path) → new → new. It also puts a hand-quoted duplicate in an entry that holds a user command. The result is exactly one command per script, the user entry keeps its matcher and command, no entry is left empty, and `sh -c` runs each script once.
- `TestRequoteDevexpHooks_DropsEmptiedEntry` checks the function directly (`pruneStaleHooks` would otherwise hide an empty entry).

Mutations killed: no collapse; keeping emptied entries.

cli/internal/hooks/installer_test.go
alexandrocuma: Fixed in 3a68c1e.

- `TestShellSyntax_IsExactlyTheListedCharacters` compares `shellSyntax` with a separately written list of all 25 characters.
- `TestInstallClaude_EveryShellSyntaxCharacter` registers a repo dir `a<c>b` for each of the 25. The command must be single-quoted, and `sh -c` must run the intended script exactly once. Near-misses `- _ . + @ % , : = é ß` must stay plain and run the same way.
- `TestShellSyntax_MatchesUninstallScript` reads `uninstall.sh`, evaluates the `SHELL_SYNTAX` literal with python3 and compares it to `shellSyntax`.
- `uninstall.test.sh` has a new check: for every character, a bare path from another root holding it is kept (not a plain path) and its single-quoted form is removed.

Mutations killed: removing `|`, `>`, `*`, `#`, `\t`, `(` or `"` from Go's `shellSyntax`, and each of the same from `uninstall.sh`'s `SHELL_SYNTAX`, which `uninstall.test.sh` also catches on its own.

docs/guides/install.md
alexandrocuma: Fixed in 3234e30.

- `install.md`: the "What gets overwritten" line and the Hook commands section now list all three spellings rewritten to the registered form (quoted only when the path needs it). Those are the bare path, the hand double-quoted path (only when its quotes are literal) and a single-quoted path that needs no quoting, which becomes plain. They also say that a spelling duplicating a registered command is removed instead, and which spellings `uninstall.sh` removes.
- The CHANGELOG entry says the same.

Issue #141, opened 2026-09-17
CI: run govulncheck on every push and PR


alexandrocuma:
Follow-up from #104.

## Problem
Vulnerability scanning is manual (`/improve` found GO-2026-5037). A called standard-library or module vulnerability can ship unnoticed between scans, and toolchain drift (e.g. an end-of-life Go minor) isn't flagged.

## Acceptance Criteria
- [ ] A CI job runs `govulncheck ./...` from `cli/` (after `./scripts/stage-assets.sh`) with the same toolchain as the build (`go-version-file: cli/go.mod`)
- [ ] The job fails on **called** vulnerabilities; uncalled findings are reported but don't fail (or the policy is documented)
- [ ] The release workflow runs the same check before goreleaser, or the release guide documents why not
- [ ] `docs/development/testing.md` / `release.md` describe the check and how to respond to a finding

## Release Targets
- `toolkit-clone` (CI only)

alexandrocuma:
Approach:

- **`ci.yml`**: new `govulncheck` job next to `test`/`hooks` (same triggers: every PR, every push to `main`). It checks out, sets up Go from `cli/go.mod` (so it scans the go1.26.8 standard library the build uses), runs `./scripts/stage-assets.sh`, installs govulncheck pinned to `v1.8.0` (not `@latest`), and runs `govulncheck -show verbose ./...` in `cli/` (source mode).
- **Policy**: govulncheck's default exit code already matches the criteria: exit 3 (job fails) only when a vulnerable symbol is reachable from our code; `-show verbose` still prints uncalled package/module findings in the log without failing.
- **`release.yml`**: the same install + scan runs as steps before goreleaser, so a called finding blocks publishing a tag.
- **Docs**: `testing.md` (CI jobs, how to run locally, how to respond to a finding), `setup.md` command table, `release.md` (release gate + what to do when it blocks), CHANGELOG `[Unreleased]`.

PR #142, opened 2026-09-17
fix(build): pin Go toolchain go1.26.8; bump golang.org/x/text (#104)

### LINKED ISSUE #104: Bump toolchain for GO-2026-5037 — verify against the version CI builds with
Surfaced by `/improve` after the v0.7.0 release. **TD-004** in the tech debt register.

## Problem

`govulncheck` reports one vulnerability that this code **actually calls**, as distinct from merely importing:

```
Vulnerability #1: GO-2026-5037
    Inefficient candidate hostname parsing in crypto/x509
  Standard library
    Found in: crypto/x509@go1.26.3
    Fixed in: crypto/x509@go1.26.4
    Example traces found:
      #1: internal/config/config.go:27:30: config.Load calls viper.Viper.GetString,
          which eventually calls x509.HostnameError.Error
```

The scan also found 12 vulnerabilities in imported packages and 1 in a required module, **none of which this code calls**. This is the only reachable one.

## Honest severity

Stating this plainly rather than inflating it because a scanner produced output:

**The practical risk here is low.** This is a denial-of-service in certificate hostname parsing, reached through viper's error formatting, in a **local CLI that parses a JSON config file**. There is no network listener, no untrusted certificate input, no multi-tenant surface. The realistic exploitation path is close to nonexistent.

It is worth fixing anyway because the fix is a version bump, and because leaving a known-reachable finding in the scan output trains everyone to ignore the scan.

## The toolchain question needs answering first

There is a discrepancy worth resolving before anyone claims this is fixed:

| Where | Version |
|---|---|
| `cli/go.mod` | `go 1.25.11` |
| `.github/workflows/ci.yml` | `go-version: "1.25"` |
| `.github/workflows/release.yml` | `go-version: "1.25"` |
| The scan that produced this finding | `go1.26.3` (local dev machine) |

The finding was produced on a **1.26.3** toolchain. The **released binary is built by CI on 1.25**. So the shipped artifact's exposure depends on which 1.25 patch release CI resolves to, and whether GO-2026-5037 is fixed in that line — which is not the same question as `go1.26.4`.

Do not close this by bumping the local toolchain and re-running the scan on a developer machine. The artifact users install is the one that matters.

## Suggested direction

1. Determine whether the 1.25 line is affected, and which patch fixes it
2. Bump the toolchain accordingly — `go.mod`, plus the pinned `go-version` in **both** workflows, which currently pin a minor (`"1.25"`) rather than a patch
3. Re-run `govulncheck` and confirm the reachable finding is gone
4. Consider whether pinning a minor version in CI is the right call, given that it silently changes the compiler under the release binary between runs

## Acceptance Criteria

- [ ] GO-2026-5037 no longer appears as a **called** vulnerability in `govulncheck ./...`
- [ ] `go.mod` and both workflow pins agree on the toolchain
- [ ] The verification is performed against the version **CI builds with**, not only a local machine
- [ ] `go test ./... -race` green on the new toolchain across all 10 packages
- [ ] The 12 uncalled import vulnerabilities are noted as unchanged, not silently conflated with this fix

## Non-Goals

- Addressing the 12 vulnerabilities in imported packages that the code does not call
- Upgrading direct dependencies — cobra, viper and promptui are all current
- Adding `govulncheck` to CI (reasonable, but a separate decision with its own trade-offs)

## Estimate

S — a version bump plus verification. The investigation of which 1.25 patch applies is the only real work.

alexandrocuma:
Closes #104 · Security (dependency hygiene)

## Summary
- **Shipped binaries were not affected** by GO-2026-5037: CI and release built with Go 1.25.14 (fixed). The finding reproduces only on dev machines running Go 1.26.0–1.26.3.
- Go 1.25 is end-of-life, and the workflows' floating `go-version: "1.25"` would have kept release builds on it.
- `cli/go.mod`: `toolchain go1.26.8` (language minimum `go 1.25.11` unchanged); older local toolchains auto-switch.
- `ci.yml` / `release.yml`: `go-version-file: cli/go.mod` — one source of truth for the toolchain.
- `golang.org/x/text` v0.38.0 → v0.39.0 for GO-2026-5970 (not called by the CLI, but compiled into release binaries and flagged by binary scans); `go mod tidy` moves `spf13/pflag` to a direct requirement.
- Docs: `setup.md` prerequisites, `workflows.md` (how to bump the toolchain), `release.md` (which toolchain builds releases); CHANGELOG `[Unreleased] ### Security`.

Follow-up filed: #141 (govulncheck in CI).

## Test plan
- [x] `govulncheck` (v1.8.0) source scan on go1.26.8: no vulnerabilities (called or uncalled); also `GOOS=linux GOARCH=amd64`
- [x] Binary-mode scan of local `-trimpath` linux/amd64 builds (unstripped and `-s -w` like goreleaser): no vulnerabilities
- [x] `go vet`, `go test ./... -race -cover`, all hook suites and installer script tests on go1.26.8
- [ ] CI log shows `go version go1.26.8` (checked after push)


alexandrocuma (commented):
**Review: the build and dependency change is correct. Three minor doc fixes, all about the same point.**

The `toolchain` line only sets a minimum for local builds. A contributor's Go is used as-is if it's newer than go1.26.8, and also whenever `GOTOOLCHAIN=local` is set. Only CI and release are held to exactly go1.26.8. The docs currently say local builds always use go1.26.8. See the inline comments. Nothing here blocks the security fix.

What I checked:
- **setup-go v6:** `src/installer.ts` reads `^toolchain go(...)` from `go-version-file`, and skips it only when `GOTOOLCHAIN` is already `local` when the step runs. `main.ts` exports `GOTOOLCHAIN=local` after reading the version. CI run 35170839230 (head `685610b`) logged `Setup go version spec 1.26.8`, `go version go1.26.8 linux/amd64` and `GOTOOLCHAIN: local` in the test steps.
- **Release workflow:** its setup-go block matches `ci.yml`: same inputs, same `cache-dependency-path`, and it's the only setup-go step in the job. `goreleaser-action@v7` and GoReleaser don't set `GOTOOLCHAIN` (I grepped both sources). GoReleaser builds with `os.Environ()` plus `CGO_ENABLED=0` and runs `go` from `PATH`, so the release uses go1.26.8 with `GOTOOLCHAIN=local`. `-trimpath` has no effect on which toolchain is picked. A local `GOOS=linux GOARCH=amd64 -trimpath -s -w` build reports `go1.26.8`.
- **Modules:** `go mod tidy -diff` is clean on go1.26.8 and on go1.25.14, and `go mod verify` passes. `spf13/pflag` is imported directly by `cli/cmd/uninstall_test.go`, so a direct requirement is correct. The vuln DB lists GO-2026-5970 as fixed in x/text `0.39.0`, and that release's `go.mod` says `go 1.25.0`, so the `go` directive doesn't change. `govulncheck@v1.8.0 -show verbose ./...` reports no vulnerabilities. go1.26.8 is the newest 1.26 patch, according to go.dev/dl.
- **Local toolchain switching:** with `GOTOOLCHAIN=auto`, Go 1.26.3 and Go 1.25.14 both switch to go1.26.8, and Go 1.27.1 stays on go1.27.1. With `GOTOOLCHAIN=local`, Go 1.26.3 and Go 1.25.14 each build with their own version. Offline with an empty module cache, the build fails with `toolchain not available`.
- **Line references:** `ci.yml:15`, `release.yml:21` and `cli/go.mod:3-5` in the docs are accurate.

docs/development/setup.md
```diff
@@ -8,7 +8,7 @@ For **contributors** working on the toolkit from a clone: prerequisites, build,
 
 | Tool | Version | Source |
 |------|---------|--------|
-| Go | 1.25 (`go 1.25.11` in the module; CI pins `"1.25"`) | `cli/go.mod:3`, `.github/workflows/ci.yml:15` |
+| Go | any Go 1.21+ on `PATH`. Builds use `toolchain go1.26.8` (minimum language version `go 1.25.11`). With the default `GOTOOLCHAIN=auto`, an older local Go downloads that toolchain and switches to it. CI and release read the same line (`go-version-file: cli/go.mod`) | `cli/go.mod:3-5`, `.github/workflows/ci.yml:15`, `.github/workflows/release.yml:21` |
```
alexandrocuma: 🔵 **Minor: "Builds use `toolchain go1.26.8`" isn't true for every contributor.** The `toolchain` line only sets a minimum for local builds. I checked these cases in `cli/`:
- **Go 1.27.1 on `PATH` (the default `GOTOOLCHAIN=auto`):** builds with go1.27.1.
- **`GOTOOLCHAIN=local`** (set in `go env` or by some distro packages):
  - Go 1.26.3 builds with go1.26.3, the GO-2026-5037 build that #104 is about.
  - Go 1.25.14 builds with go1.25.14.
  - Anything older than 1.25.11 fails.
  
  So "any Go 1.21+" is only true with the default `auto` setting.
- **First build with an older Go needs network.** Offline (`GOPROXY=off`, empty module cache) it fails with `go: download go1.26.8 for darwin/arm64: toolchain not available`. `install.sh:13` then prints "Ensure Go is installed", which points contributors at the wrong problem.

Suggested wording: *Go 1.21+ on `PATH`. `cli/go.mod` sets `toolchain go1.26.8` as the minimum. With the default `GOTOOLCHAIN=auto`, an older Go downloads go1.26.8 on first build (needs network), and a newer Go builds with itself. With `GOTOOLCHAIN=local` you need Go ≥ 1.25.11, and the build uses that version. CI and release install exactly go1.26.8 (`go-version-file: cli/go.mod`).* It could also go in the "Install deps" row (line 44): the toolchain download happens alongside the module downloads.

docs/guides/workflows.md
```diff
@@ -106,7 +106,7 @@ Values for `mcps/.env` go in the gitignored file, never in the registry or in co
...
+- **Go module:** add or bump it in `cli/go.mod` with the Go module tooling (`go get`, `go mod tidy` in `cli/`), and commit `cli/go.mod` and `cli/go.sum` together. There's no written approval rule. In practice, bumps land in their own `fix:` commit whose body says why (drift or CVE) and confirms the upgrade had no impact (`e6f65e3` cobra/viper, `e147e67` Go toolchain `1.25.11` for stdlib CVEs). **Go toolchain:** the `toolchain` line in `cli/go.mod` is the one place the build's Go version is set. Local builds switch to it under `GOTOOLCHAIN=auto`, and both workflows install it through `go-version-file: cli/go.mod` (`.github/workflows/ci.yml`, `.github/workflows/release.yml`), so don't add a `go-version` there, and don't set `GOTOOLCHAIN` in workflow or job `env` (setup-go then ignores the `toolchain` line and falls back to the `go` directive). To pick up a stdlib fix, bump only the `toolchain` line to a supported, fixed patch and check with `govulncheck ./...` from `cli/`. Leave the `go` directive (the minimum language version) alone unless code needs a newer one.
```
alexandrocuma: 🔵 **Minor: "the one place the build's Go version is set. Local builds switch to it"** only holds when the local Go is older than go1.26.8 and `GOTOOLCHAIN` is `auto`. Newer Go, or `GOTOOLCHAIN=local`, builds with the local version (details in the `setup.md` comment). Suggested: *"sets the exact Go version for CI and release, and the minimum for local builds (an older Go switches up to it under `GOTOOLCHAIN=auto`)"*.

The `GOTOOLCHAIN` warning is sound. For precision: setup-go v6 (`src/installer.ts`) ignores the `toolchain` line only when `GOTOOLCHAIN` is already `local` when setup-go runs. A second setup-go step in the same job would hit this, because the first step exports `local`.

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **Pinned the Go toolchain to a patched release (#104).** Building `devexp`
+  with Go 1.26.0–1.26.3 left a standard-library advisory (GO-2026-5037) in code
+  the CLI calls. `cli/go.mod` now pins `toolchain go1.26.8`, and CI and the
+  release workflow read the Go version from that file, so local builds, CI and
+  release binaries all use the same patched toolchain. Released binaries up to
```
alexandrocuma: 🔵 **Minor:** "so local builds, CI and release binaries all use the same patched toolchain" overstates what local builds get. Local builds use *at least* go1.26.8 under the default `GOTOOLCHAIN=auto`, and their own Go under `GOTOOLCHAIN=local`. Suggested: *"CI and the release workflow read the Go version from that file, and local builds with an older Go switch up to it."*

docs/development/setup.md
alexandrocuma: Fixed in 24f0514. The Go row in `setup.md` now describes the toolchain line the way you suggested:
- **Local builds:** `toolchain go1.26.8` is a minimum. Under `GOTOOLCHAIN=auto`, an older Go downloads go1.26.8 on the first build; that needs network, and offline it fails with `toolchain not available`. A newer Go builds with itself. With `GOTOOLCHAIN=local` you need Go ≥ 1.25.11, and the build uses that version.
- **CI and release:** they install exactly go1.26.8.

The "Install deps" row now mentions the toolchain download next to the module downloads.

`install.sh` also no longer stops at "Ensure Go is installed". When the build fails, it names the toolchain from `cli/go.mod`, says the first download needs network, and gives the offline options: install that version or newer, or use `GOTOOLCHAIN=local`. A new `install.test.sh` case covers the message; it fails 2 checks against the old `install.sh`. The `install.sh` line references in the docs moved by 2 lines, and I updated them.

docs/guides/workflows.md
alexandrocuma: Fixed in 24f0514. `workflows.md` now says the `toolchain` line "sets the exact Go version for CI and release, and the minimum for local builds": an older Go switches up to it under `GOTOOLCHAIN=auto`, and a newer Go or `GOTOOLCHAIN=local` builds with the installed Go. I kept the warning not to set `GOTOOLCHAIN` in workflow or job env. It now spells out the condition: setup-go ignores the `toolchain` line only when `GOTOOLCHAIN` is already `local` as it runs. It also warns against a second setup-go step in the same job, because the first step exports `GOTOOLCHAIN=local`.

CHANGELOG.md
alexandrocuma: Fixed in 24f0514. The entry now reads: "CI and the release workflow read the Go version from that file, and local builds with an older Go switch up to it." It also notes the new `install.sh` hint for a build that fails on the toolchain download.

Issue #144, opened 2026-09-17
large-file-guard: add an opencode behaviour test


alexandrocuma:
Found during #101.

## Problem
`hooks/opencode/large-file-guard.js` has no behaviour test (the Claude Code side has `large-file-guard.test.sh`). `docs/development/testing.md` records the gap.

## Acceptance Criteria
- [ ] `hooks/opencode/large-file-guard.test.js` mirrors the Claude Code test's decisions (size boundaries, missing file, directory, hostile paths) against the real module
- [ ] Mutation-checked; `testing.md` updated

## Release Targets
- `cli`, `toolkit-clone`

Issue #143, opened 2026-09-17
secret-in-write-guard: key-shaped placeholders and quoted PEM headers are blocked


alexandrocuma:
Found during #101 (out of its scope).

## Problem
Placeholder values shaped like real credentials still block writes: e.g. a Slack token prefix followed by a word like `your-token`, AWS's published documentation example access key ID, bodies made of repeated `x`, and a bare PEM private-key header quoted in documentation. This is the same false-positive class #89 fixed for `secret-guard`.

## Acceptance Criteria
- [ ] Obvious placeholders (documented example values, `your-…`/`<…>`/repeated-character bodies, header-only PEM lines without key material) are allowed in both CLIs
- [ ] Every real-shaped key still blocks (no new false negatives; mutation-checked)
- [ ] Mirrored tests in `secret-in-write-guard.test.sh` / `.test.js`

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
Related false positive, from the PR #145 review: the pre-existing `gh[posta]_` pattern allows `_` in the body, so long snake_case identifiers that happen to contain one of those prefixes (for example a test function name containing `ght_`) are blocked. Matching GitHub's real token shape (fixed alphanumeric body, no `_`) would fix it without reopening real tokens.

alexandrocuma:
## Grooming plan (#143, with #152)

**Validated against:** `10ad67c` (`main`) · **Verdict:** corrected and proceeding · **Targets:** `cli`, `toolkit-clone`

### Corrections to the ticket and follow-up comment
- The GitHub pattern is `gh[postaur]_` since #101, not `gh[posta]_`.
- "Fixed alphanumeric body, no `_`" doesn't hold for every GitHub token any more. Since 2026-04-27 GitHub App installation tokens (including Actions `GITHUB_TOKEN`) are rolling out in a longer stateless format that contains `_`, `.` and `-` ([changelog](https://github.blog/changelog/2026-05-15-github-app-installation-tokens-per-request-override-header/)). Dropping `_` everywhere would stop detecting those tokens, so that format gets its own shape match.
- Values wrapped in `<…>` are already allowed. They only need tests.

### Plan
1. **GitHub tokens:** match the classic tokens by their documented shape and the stateless installation tokens by theirs, so long snake_case names stop blocking and no real token format is dropped.
[...]

PR #145, opened 2026-09-17
fix(hooks): secret-in-write-guard blocks every secret it claims to, on every write tool (#101)

### LINKED ISSUE #101: secret-in-write-guard has no behavioral test — the one security guard nothing verifies
Surfaced by `/improve` after the v0.7.0 release. **TD-001** in the tech debt register — the only P1.

## Problem

Three security guards ship enabled by default. Two of them have behavioral test twins that assert they actually block:

| Guard | Fail-closed test | Behavioral test |
|---|---|---|
| `secret-guard` | ✅ `fail-closed.test.sh` | ✅ `secret-guard.test.sh` |
| `dangerous-cmd-guard` | ✅ `fail-closed.test.sh` | ✅ `dangerous-cmd-guard.test.sh` |
| **`secret-in-write-guard`** | ✅ `fail-closed.test.sh` | ❌ **none** |

`fail-closed.test.sh` (added in #88/#92) proves the guard *errors loudly* when its interpreter breaks. It does not prove the guard **blocks an API key from being written to disk** — which is the only thing it exists to do.

So the one guard standing between the model and a credential on disk is the one nothing verifies, in either direction.

## Why this is P1 rather than housekeeping

- **A regression here is silent and expensive.** A broken `secret-guard` announces itself — you get blocked reading files. A broken `secret-in-write-guard` does nothing at all, and the failure mode is a secret written to disk with no signal.
- **Its sibling had an entire undetected false-positive class.** #89 found `secret-guard` blocking committed templates and bare mentions — wrong in the *permissive-looking* direction while appearing to work. There is no reason to assume `secret-in-write-guard` is free of the same class; there is simply no test that would tell us.
- **The cost is near zero.** #88 made CI run every `*.test.sh` and `*.test.js`, so a twin executes the moment it lands. No pipeline work required.

## Suggested shape

Mirror `secret-guard.test.sh`, asserting both directions:

**Must block (red):** an API key, a bearer token, or private key material appearing in `Write`/`Edit` content.

**Must pass (green):** prose that merely *mentions* a token; a `.env.example` / `.sample` / `.template` file documenting which keys a project needs; a variable named `token` with no secret value. These are the exact cases #89 had to fix in the sibling guard — the twin should pin them here before they become a bug report.

Both implementations need the twin, as #89 did: `hooks/claude-code/secret-in-write-guard.sh` and `hooks/opencode/secret-in-write-guard.js`.

## Acceptance Criteria

- [ ] `hooks/claude-code/secret-in-write-guard.test.sh` asserts red cases block (exit 2) and green cases pass (exit 0)
- [ ] `hooks/opencode/secret-in-write-guard.test.js` covers the same cases for the JS twin
- [ ] Template files (`.example`, `.sample`, `.template`, `.dist`) are explicitly covered as green
- [ ] A bare mention of a secret-like word in prose is explicitly covered as green
- [ ] Both run in CI via the existing `hooks` job — no workflow changes needed
- [ ] Existing `fail-closed.test.sh` coverage stays green

## Non-Goals

- Changing what `secret-in-write-guard` blocks — this is a test-only ticket unless the twin exposes a real defect, in which case that fix folds in here
- Adding tests for the three `graphify-*` hooks (disabled by default; tracked as TD-008 in the register)

## Estimate

S — one test twin per implementation, reusing `secret-guard.test.sh`'s structure.

alexandrocuma:
Closes #101 · Security

## Summary
Adding the missing behaviour tests for `secret-in-write-guard` (the one enabled security guard nothing verified) showed that it let secrets be written without any block in several cases. All fixed, mirrored for Claude Code and opencode:

- **Claude Code: some secret kinds were never blocked** — pattern matching relied on `grep`, which could fail silently (a pattern read as an option; a pipeline error on large input), and a failure counted as "no match". Matching now happens in the guard's `python3 -I` step, which blocks on any internal error.
- **opencode: edits were never scanned** — the guard read the wrong argument name for the `edit` tool.
- **Write tools not covered** — opencode `apply_patch` (which some models use instead of write/edit) is now scanned (added lines only), and Claude Code `MultiEdit` / `NotebookEdit` are routed to the guard (registry matcher `Write|Edit|MultiEdit|NotebookEdit`).
- **Current token formats** — current OpenAI project/service/admin keys, GitHub user/refresh tokens and fine-grained PATs, and AWS temporary access keys are detected. Patterns were only added.
- Docs: `testing.md`, `conventions.md`, `docs/reference/hooks.md` (matcher semantics corrected); CHANGELOG `[Unreleased] ### Fixed`.

**Upgrade note:** existing Claude Code installs keep the previously registered matcher until the installer learns to update matchers — that change is in the #137/#138 delivery (same release).

## Out of scope (filed)
- #143 — key-shaped placeholders still blocked
- #144 — opencode `large-file-guard` behaviour test

## Test plan
- [x] New `hooks/claude-code/secret-in-write-guard.test.sh` (124) and `hooks/opencode/secret-in-write-guard.test.js` (107): every pattern blocked through every write tool, at any position and in large writes; placeholders, prose, templates, public keys, look-alikes, removals allowed; message names the kind without echoing the secret. Fake values are built at runtime.
- [x] Against the previous guard the new tests fail (48 sh / 29 js at the last step; 31 / 16 against the original)
- [x] Mutation checks: 47/47 caught
- [x] Go `go test ./... -race` (go1.26.8), all hook suites, installer script tests


hooks/claude-code/secret-in-write-guard.sh
```diff
@@ -1,43 +1,65 @@
...
+if isinstance(edits, list):
+    parts += [e.get('new_string') for e in edits if isinstance(e, dict)]
+content = '\n'.join(str(p) for p in parts if p)
+
+PATTERNS = [
+    ('an Anthropic API key (sk-ant-...)', r'sk-ant-[A-Za-z0-9_-]{40,}'),
+    ('an OpenAI API key (sk-...)',        r'sk-[A-Za-z0-9]{32,}'),
+    # Project, service-account and admin keys: the body has _ and -, so the
+    # prefix must start a word, or kebab-case ids like desk-admin-... match.
+    ('an OpenAI API key (sk-...)',        r'(^|[^A-Za-z0-9_-])sk-(proj|svcacct|admin)-[A-Za-z0-9_-]{40,}'),
```
alexandrocuma: **Medium: the left boundary on the new OpenAI project/service/admin pattern misses real keys and still blocks some kebab-case ids.** (The same applies to the JS twin at line 21.)

`(^|[^A-Za-z0-9_-])` treats the key as not starting a word whenever the character before it is a letter, digit, `_` or `-`. In serialized text that happens with a real key in two common cases:
- right after an escape sequence inside a quoted string (JSON, YAML, a code string literal)
- right after a percent-encoded character in a URL

A key after `_` or `-` is missed for the same reason. In the other direction, the boundary doesn't stop false positives when a kebab-case id *starts* with the prefix, such as a class name or i18n key of 40+ characters beginning `sk-admin-`. I checked both twins on each of these classes, and they agree.

**Fix:** drop the left boundary and make the body carry the signal. Real project keys are well over 100 characters, so a minimum like `{80,}` with no boundary catches every missed case in my corpus and allows all the look-alike ids, including the existing `desk-admin-…` / `task-proj-…` fixtures. Another option is the base64 marker that gitleaks' rule uses. Add block cases for the escaped and encoded contexts.

**Test gap:** replacing the class with `[^A-Za-z0-9]` in both twins leaves both suites green, so nothing pins the `_` and `-` members. The PR's "mutation 47/47" doesn't cover this. The minimum lengths (`{40,}`, `{36,}`) aren't pinned at the edge either.

hooks/claude-code/secret-in-write-guard.sh
```diff
@@ -1,43 +1,65 @@
...
+    ('an AWS Access Key ID (AKIA...)',    r'AKIA[0-9A-Z]{16}'),
+    # Temporary (STS) key IDs are exactly 20 characters; bounding both ends
+    # keeps words like ASIAPACIFICDATACENTER01 from matching.
+    ('an AWS temporary Access Key ID (ASIA...)', r'(^|[^A-Za-z0-9])ASIA[0-9A-Z]{16}([^A-Za-z0-9]|$)'),
```
alexandrocuma: **Low: the ASIA pattern has the same boundary problem on both sides.** (JS twin: line 25.)

- A temporary key ID right after an escape sequence or a percent-encoded character isn't matched, because the character before it is alphanumeric.
- The strict right boundary also drops a key followed directly by a letter or digit.

`AKIA` has no boundaries, so the two AWS rules now behave differently in the same contexts. The left boundary is needed for the `EURASIA…` fixture, so it can't just be removed. One option is to allow a boundary that is either non-alphanumeric or a common escape (backslash-letter, `%XX`). Whatever the choice, add one block case for an escaped context so the decision is pinned.

hooks/opencode/secret-in-write-guard.js
```diff
@@ -1,27 +1,59 @@
...
+  // Project, service-account and admin keys: the body has _ and -, so the
+  // prefix must start a word, or kebab-case ids like desk-admin-... match.
+  { re: /(^|[^A-Za-z0-9_-])sk-(proj|svcacct|admin)-[A-Za-z0-9_-]{40,}/m, label: 'OpenAI API key (sk-...)' },
   { re: /AKIA[0-9A-Z]{16}/m,                     label: 'AWS Access Key ID' },
-  { re: /gh[posta]_[A-Za-z0-9_]{36,}/m,          label: 'GitHub token' },
+  // Temporary (STS) key IDs are exactly 20 characters; bounding both ends
+  // keeps words like ASIAPACIFICDATACENTER01 from matching.
+  { re: /(^|[^A-Za-z0-9])ASIA[0-9A-Z]{16}([^A-Za-z0-9]|$)/m, label: 'AWS temporary Access Key ID' },
+  { re: /gh[postaur]_[A-Za-z0-9_]{36,}/m,        label: 'GitHub token' },
+  { re: /github_pat_[A-Za-z0-9_]{36,}/m,         label: 'GitHub token' },
```
alexandrocuma: **Low (false positive): the `github_pat_` body allows `_` without limit, so any long snake_case identifier that starts with `github_pat_` blocks.** A config attribute or vault-key name is enough (shell twin: line 47).

Fine-grained PATs have a fixed shape: `github_pat_`, 22 alphanumerics, `_`, then 59 alphanumerics. This pattern is new, so matching that shape (`github_pat_[A-Za-z0-9]{22}_[A-Za-z0-9]{59}`) doesn't reopen anything that was blocked before. Add an allow case for a long `github_pat_…` identifier.

alexandrocuma (commented):
## PR review: secret-in-write-guard (#101)

**Recommendation: Request changes.** Posted as a comment, not as a formal "request changes". The findings are small and all inside the diff.

The fix is solid. Moving the matching into the Python step removes both silent fail-open paths, and the guard now covers every tool that writes file content. I also checked the claims beyond the PR's own tests:

- **Parity:** I ran a 324-case corpus (every pattern in about 20 surrounding contexts, plus realistic false-positive probes) through both twins. They made the same decision on every case.
- **opencode `apply_patch`:** checked against upstream `packages/opencode/src/patch/index.ts`. The parser only writes `+` lines between the Begin and End markers (context lines count only after a fuzzy match, see the inline note). `patchAddedText` scans every `+` line in the text, so it covers more than the parser writes: heredoc wrappers, a missing End marker, several file sections, `*** Move to`, CRLF, `\ No newline` lines. Tool ids `write`/`edit`/`apply_patch` and the arg names `content`/`newString`/`patchText` match upstream.
- **Claude Code:** the matcher semantics in `hooks.md` match the current hooks docs (a plain `|` list is matched as exact names). `NotebookEdit` `new_source`, `edit_mode` and `cell_type` are handled. A delete-mode cell that still carries `new_source` gets scanned, which is a harmless false positive.
- **Error paths:** a null `tool_input`, a non-JSON body, an empty body and a top-level array all exit 2. Non-string or null fields and junk `edits` entries are ignored safely. A 20 MB write takes about 1 s in the shell twin. Adversarial near-miss inputs of 2–10 MB stay linear in both twins.
- **Suites:** all hook suites and the installer script tests pass.

Positives: runtime-built fixtures; tests that check the message never repeats the secret and that allowed writes leave stderr empty; the registry-routing test; the regression cases for the pipe-buffer and dash-pattern failures.

Questions: OpenAI has also issued user keys with a prefix other than proj/svcacct/admin (`sk-None-…`). Leaving those out, was that deliberate? Separately, the pre-existing `gh[posta]_` family allows `_` in the body, so long snake_case names such as test function names containing `…ght_…` get blocked. That's outside this diff, but worth a follow-up next to #143.

hooks/opencode/secret-in-write-guard.js
```diff
@@ -1,27 +1,59 @@
 /**
  * secret-in-write-guard.js — hard-blocks writing content that contains secret patterns
  *
- * Event: tool.execute.before (tool: write | edit)
+ * Event: tool.execute.before (tool: write | edit | apply_patch)
  *
- * Scans the content being written for high-signal secret patterns.
+ * Scans the text being written for high-signal secret patterns: write
+ * `content`, edit `newString`, and the lines apply_patch adds (opencode offers
+ * apply_patch instead of write/edit to GPT models). Text being replaced or
```
alexandrocuma, line 8: **Low: known limitation worth documenting.** Only the new text of an edit is scanned, never the text around it in the file. A secret whose prefix is already in the file (for example a template value with the prefix kept) can be completed by an edit that supplies only the body. The guard sees neither half. This applies to Claude Code Edit/MultiEdit and to opencode edit.

For `apply_patch`, the comment "unchanged context isn't written" is only mostly true. Upstream `seekSequence` matches context lines fuzzily (trimmed whitespace, Unicode dashes and quotes normalized to ASCII), and the replacement writes the *patch's* context text. So a near-key already in the file can come out in its ASCII form.

Both cases are narrow and adversarial rather than accidental, so I don't think they block merge. Say so in the header and in `hooks.md` so nobody reads the guard as covering them. If you want to close it, scan the new text together with a small window of the file around the replaced span.

CHANGELOG.md
```diff
@@ -7,6 +7,58 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    `apply_patch` instead of `write` and `edit`, so with those models nothing
+    was checked. The guard now scans the lines a patch adds (`+` lines, which
+    covers new-file bodies and added lines in update hunks). Removed lines,
+    unchanged context and `***`/`@@` headers aren't scanned, so a patch that
+    deletes a key is allowed.
+  - Claude Code: the matcher `Write|Edit` matches tool names exactly, so the
+    guard never ran for `NotebookEdit` (`new_source`) or for `MultiEdit`
+    (`edits[].new_string`, in older releases). The registry matcher is now
+    `Write|Edit|MultiEdit|NotebookEdit` and the guard scans those fields. Text
+    being replaced (`old_string`) is still not scanned. `devexp install` leaves
```
alexandrocuma: **Low (docs):**
1. This sentence says `devexp install` keeps an existing matcher until the hook is reinstalled or the matcher is edited by hand. The PR body says the #137/#138 delivery (same release) makes install update matchers. Whichever PR merges second needs to reconcile this entry, or users get conflicting instructions.
2. Keep the entry class-level. Lines 14–22 give the exact pipe-buffer size and the exact grep failure mode. For a public repo, and users who haven't upgraded yet, "some key kinds and large writes were never checked" says enough. The PR body already words it that way.
3. Line 35, "misses no current OpenAI or GitHub token format", overclaims given the boundary gaps noted on the guard.

docs/development/testing.md
```diff
@@ -18,7 +18,7 @@ CI (`.github/workflows/ci.yml`) runs on every pull request and every push to `ma
 | E2E | N/A — no automated end-to-end install test. `runInstall`, `doInstallClaude`, `doInstallOpencode` and `runWizard` have 0% coverage. Manual check: `./install.sh --dry-run` | — |
 | Agents / skills (Markdown) | N/A — no automated validation. Edit → `./install.sh` → try it in Claude Code or opencode | `docs/development/README.md` (Notes) |
 
-At this commit: 10 Go packages with tests; `dangerous-cmd-guard.test.sh` 25 cases, `fail-closed.test.sh` 10, `secret-guard.test.sh` 37; `dangerous-cmd-guard.test.js` 25, `devexp-plugin.test.js` 130, `secret-guard.test.js` 38; `uninstall.test.sh` 94, `install.test.sh` 17, `remote-install.test.sh` 24.
+At this commit: 10 Go packages with tests; `dangerous-cmd-guard.test.sh` 25 cases, `fail-closed.test.sh` 10, `interpreter-isolation.test.sh` 11, `large-file-guard.test.sh` 34, `on-save-path.test.sh` 73, `secret-guard.test.sh` 37, `secret-in-write-guard.test.sh` 124; `dangerous-cmd-guard.test.js` 25, `devexp-plugin.test.js` 130, `on-save-path.test.js` 61, `secret-guard.test.js` 38, `secret-in-write-guard.test.js` 107; `uninstall.test.sh` 107, `install.test.sh` 17, `remote-install.test.sh` 24.
```
alexandrocuma: **Nit:** this line was rewritten as "At this commit", but `install.test.sh` reports 21 cases on this branch, not 17. The other counts I ran match.

hooks/claude-code/secret-in-write-guard.sh
alexandrocuma: Fixed in 9312a37. The left boundary is gone and the body length carries the signal: the minimum sits well below real key lengths and above look-alike ids. New block cases cover a key after an escape sequence, a percent-encoded character, `_`, `-` and a digit; new allow cases cover long kebab-case ids that start with the prefix. The Anthropic and OpenAI minimums, and the other length thresholds this PR introduced, are now pinned at the edge (one short allows, exact blocks). Each boundary-class and threshold mutation now fails both suites.

hooks/claude-code/secret-in-write-guard.sh
alexandrocuma: Fixed in 2ef2165. A boundary is now any character that cannot be part of an ID, plus, on the left, an encoded character (`%XX`, `\uXXXX`, `\xXX`). The choice and its trade-off are documented next to the pattern in both twins. Block cases cover escaped and encoded left contexts and a key followed by a lowercase letter; the `EURASIA…` and `ASIAPACIFIC…` fixtures still allow, and the body length is pinned at the edge.

hooks/opencode/secret-in-write-guard.js
alexandrocuma: Fixed in dcad85e: both twins match the exact fine-grained shape. Allow cases cover long `github_pat_…` snake_case identifiers, and tests pin each segment length at the edge, the charset and the separator.

hooks/opencode/secret-in-write-guard.js
alexandrocuma: Documented in 179c8d6, class-level, in both twins' headers and in `docs/reference/hooks.md`: only the new text is scanned, and in opencode `apply_patch` context lines are matched loosely and written from the patch. Not closing the gap in this PR.

CHANGELOG.md
alexandrocuma: Done in d56d881. The first entry is class-level now, the overclaim is gone, and the matcher sentence says to re-run `devexp install`, which #147 (#137/#138) makes enough in the same release.

docs/development/testing.md
alexandrocuma: Fixed in 016e73f. I re-checked every count on that line against this branch: `install.test.sh` 21, and `secret-in-write-guard` 159 (sh) / 141 (js) after this round.

alexandrocuma:
Answers to the two review questions:

1. **`sk-None-…` user keys:** leaving them out wasn't deliberate. Public detector rules and write-ups (Trivy's legacy OpenAI user-key rule, for example) describe `sk-None-…` as a legacy user-key variant. OpenAI no longer issues it, but existing keys can still be live, and the legacy pattern didn't match it. addf809 adds it in both twins, with block cases through every tool, the minimum length pinned at the edge and a short placeholder allow case.
2. **`gh[posta]_` false positive on long snake_case names:** this is outside this diff's scope. It's recorded as a follow-up on #143: https://github.com/alexandrocuma/devexp-toolkit/issues/143#issuecomment-5707425040

I also ran a new parity corpus through the real shell hook and the opencode handler: runtime-built fake keys in many contexts, plus look-alike probes. The twins made the same decision on every case. Each boundary, threshold and shape mutation from this round fails both suites.

PR #147, opened 2026-09-17
fix(hooks): keep users' hook fields, prune only devexp's stale hooks, sync matchers (#137, #138)

### LINKED ISSUE #137: Rewriting settings.json drops extra fields from users' hooks

### LINKED ISSUE #138: Stale hook pruning removes user commands under the repo dir

alexandrocuma:
Closes #137
Closes #138

## What changed

**#137: install and uninstall keep every field of users' hooks.**
- `devexp install` rewrote `~/.claude/settings.json` from types holding only `matcher`, `type` and `command`. That dropped every other field on users' hooks (`timeout`, `async`, `shell`, `if`, `statusMessage`, `args`, and the fields of `http`/`mcp_tool`/`prompt` handlers). It also re-sorted and re-escaped the whole file.
- New `cli/internal/hooks/settings.go` keeps each handler's and entry's fields, their order and their values. Only the top-level `hooks` value is spliced back; the result must decode to the original with only `hooks` changed, or nothing is written.
- What is kept, precisely:
  - **Outside `hooks`:** every byte, including line endings (LF or CRLF) and numbers a float can't hold (`1e400`).
  - **Inside `hooks`:** fields, key order and values. When devexp writes, the whole `hooks` value is re-indented to the file's indentation (a one-line handler is spread over several) and written with the file's line ending, and escaped keys are re-encoded (`"comm\u0061nd"` becomes `"command"`). Install keeps string escapes and number spellings in values; `uninstall.sh` writes them as python's `json` does.
- devexp only treats a handler as its own when it has `type: "command"` and no `args`. A handler with `args` is spawned without a shell, so re-quoting it would break it.
- A `settings.json` that isn't valid JSON, or has an unexpected shape, is left untouched and install stops with an error naming it. Before, it was replaced by a file holding only the hooks.
- `uninstall.sh` applies the same ownership rule and splice, reads and writes with `newline=''` (CRLF stays CRLF), and leaves a file holding `NaN`, `Infinity` or `1e400` untouched with a message (python would write `1e400` back as `Infinity`, which isn't JSON).

**#138: pruning only removes devexp's own stale commands.**
- `isStaleDevexpHook` requires the exact form devexp writes: a plain or single-quoted `<repo>/hooks/claude-code/<name>`, or the bare legacy path, with nothing on disk.
- Commands with arguments, chaining, wrappers, other directories, variables or globs are the user's and are never pruned.

**Matcher sync.** Install used to skip a hook that was already registered, whatever its matcher, so existing installs never picked up a registry matcher change (e.g. `secret-guard` `Read` → `Read|Bash`, `secret-in-write-guard` gaining `MultiEdit|NotebookEdit` in #145). Now an enabled devexp hook moves to the registry's current matcher:
- If its entry holds only that hook, the matcher is updated in place (a `matcher` key the entry lacked is written before `hooks`).
- If the entry is shared with other commands, the devexp handler moves into its own entry and the other commands keep theirs.
- A second install writes nothing.

## Review fixes (#147 review)
- Go: the safety check decodes with `json.Number`, so `1e400` no longer aborts install; spliced lines use the file's line ending.
- `uninstall.sh`: `newline=''` on read and write; `parse_constant`/`parse_float` reject non-finite numbers, `allow_nan=False` as a backup, file left untouched with a `[skip]`.
- Matcher key goes before `hooks`; behaviour change documented (below, CHANGELOG, `docs/guides/install.md`).
- Docs: real examples; exact wording of what is kept.
- New pins for the five surviving mutations (requote `present` counting exec-form handlers, an entry gaining `"hooks": null`, moved handler taken from the last duplicate, read error ignored, python picking the first repeated `hooks` key). The decode checks (Go and python) can't fail by construction; code comments say so.

## Verification
- Go: `gofmt`/`go vet` clean; `go test -race ./...` passes; `internal/hooks` coverage 92.1%; 1034 test/subtest passes (417 in `internal/hooks`).
- Byte-exact tests: `TestInstallClaude_KeepsUserHookFields`, `TestInstallClaude_SettingsLayout` (2/4-space, tab, compact, `{}`, `null`, new file, CRLF ×3, `1e400`), `TestInstallClaude_SyncsRegistryMatcher` (7 cases × 2 runs), `TestIsStaleDevexpHook` (14 user-command cases), `TestInstallClaude_UnreadableSettings`.
- `uninstall.test.sh` 128, including CRLF, `NaN`/`Infinity`/`±1e400`, a repeated `hooks` key, and a real-binary round trip: install → re-install (identical bytes) → `uninstall.sh` ends byte-identical to the start.
- After merging `main` (#145, #149): all hook suites pass (dangerous-cmd-guard 169 sh / 170 js, secret-in-write-guard 159 / 141, others unchanged), `install.test.sh` 21, `remote-install.test.sh` 24. With the real registry, an install whose `secret-in-write-guard` entry was `Write|Edit` (CRLF file, `1e400` elsewhere) moves it to `Write|Edit|MultiEdit|NotebookEdit` and keeps the rest.
- Mutation checks: Go 44/45 killed; the survivor is the render decode check, which can't fail by construction. Python (`uninstall.test.sh`) 16/19 killed; the survivors are the decode check (same reason), `allow_nan=False` (a backup the parse hooks already cover) and `newline=''` on write (text mode writes LF on POSIX anyway).

## Behaviour changes
- `devexp install` now fails, leaving the file untouched, on a malformed `settings.json`.
- A devexp hook that a user converted to exec form (`args`) no longer counts as devexp's registration; install adds its own entry alongside it.
- **Install enforces the registry's matcher on devexp's own handlers.** A matcher you set by hand on one, or an entry with no `matcher` (matching every tool), is reset to the registry's on the next install. To use a matcher of your own, disable the hook (`hooks.disabled` in `devexp.config.json`), remove devexp's entry for it, and register your own command for it (a wrapper script, or the script with `"args": []`).
- `uninstall.sh` leaves a `settings.json` holding `NaN`, `Infinity` or a number a float can't hold untouched, instead of writing invalid JSON.

## Out of scope (follow-ups)
- `InstallClaude` and the `uninstall.sh` python step write `settings.json` non-atomically. An atomic rename would replace a symlinked file, so this needs its own design.
- Dry-run install prints "add" for hooks that are already registered and doesn't preview matcher updates.


uninstall.sh
```diff
@@ -401,48 +408,101 @@ def is_devexp_hook(cmd):
...
+
+# Only the hooks value is rewritten, in the layout of the line its key is on;
+# every other byte of settings.json stays as it was.
+key_start, value_start, value_end = top_level_member(text, 'hooks')
+lead = text[text.rfind('\n', 0, key_start) + 1:key_start]
+if lead.strip(' \t'):
+    value = json.dumps(hooks_section, ensure_ascii=False, separators=(',', ':'))
 else:
-    print("  [skip] no devexp hooks found in settings.json")
+    value = json.dumps(hooks_section, ensure_ascii=False, indent=lead or '  ').replace('\n', '\n' + lead)
```
alexandrocuma: Numbers outside the float range become invalid JSON here. `json.loads` reads `1e400` as `inf`, and `json.dumps` writes `Infinity`. `json.loads(new_text) != settings` passes (`inf == inf`), so the file gets written and Claude Code can't parse it.

Repro: a user handler `{"type":"command","command":"/u","timeout":1e400}` next to a devexp handler → uninstall writes `"timeout": Infinity`.

Suggested fix: `json.dumps(..., allow_nan=False)` on both branches, catching `ValueError` → `[skip]`. Also parse with `parse_constant` raising, so `NaN`/`Infinity` input is refused as Go refuses it. Add a test case.

uninstall.sh
```diff
@@ -401,48 +408,101 @@ def is_devexp_hook(cmd):
...
     return head == SCRIPT_DIR or head.endswith('/' + SCRIPT_DIR)
 
-with open(settings_path) as f:
-    try:
-        settings = json.load(f)
-    except json.JSONDecodeError:
-        print("  [skip] settings.json is not valid JSON")
-        sys.exit(0)
+try:
+    with open(settings_path, encoding='utf-8') as f:
```
alexandrocuma: Text-mode `open()` uses universal newlines: CRLF (and a lone CR) becomes `\n` on read, and the file is written back LF. Every line outside `hooks` changes, which contradicts install.md:126.

Repro: `{\r\n  "model": "x",\r\n  "hooks": {...devexp handler...},\r\n  "z": 1\r\n}\r\n` → written with `\n` only.

Suggested fix: `newline=''` on this open and on the write at :503, plus a CRLF test. Go already keeps CRLF outside `hooks`, though the spliced `hooks` value uses LF.

cli/internal/hooks/settings.go
```diff
@@ -0,0 +1,416 @@
...
+			insert = concat([]byte(",\n"+l.prefix+`"hooks": `), l.format(hooks))
+		} else {
+			insert = concat([]byte(`,"hooks":`), hooks)
+		}
+		out = concat(data[:last.end], insert, data[last.end:])
+	}
+
+	want := map[string]any{}
+	if len(bytes.TrimSpace(data)) > 0 {
+		if err := json.Unmarshal(data, &want); err != nil {
```
alexandrocuma: Decoding into `map[string]any` fails on valid JSON with a number outside float64 range. `{"a": 1e400}` makes install stop with `json: cannot unmarshal number 1e400 into Go value of type float64 — it was left untouched`, and the whole `devexp install` aborts (opencode, manifest).

The file stays safe, but the check itself causes the failure. Suggested fix: decode `want` and `got` with a `json.Decoder` using `UseNumber()`. Bytes outside `hooks` are verbatim and `hooks` comes from the same compact bytes, so `json.Number` string comparison is exact.

cli/internal/hooks/installer.go
```diff
@@ -232,25 +207,93 @@ func InstallClaude(registry Registry, repoDir, settingsPath string, disabled []s
...
+	}
+
+	var moved *hookCmd
+	updated := false
+	for i := range entries {
+		e := &entries[i]
+		switch n := countFunc(e.Hooks, ours); {
+		case n == 0:
+		case n == len(e.Hooks):
+			e.Matcher = matcher
```
alexandrocuma, line 241: This overwrites a matcher a user chose on purpose for a devexp handler (e.g. narrowing `secret-guard` to `Read`), on every install. Before this PR it was kept. It also narrows an entry with no `matcher` (matches all tools), and `withString` appends `"matcher"` after `"hooks"`.

That may be the right call, but it's a behaviour change. Please list it under "Behaviour changes" in the PR body and in the CHANGELOG, together with how to keep a custom matcher: disable the hook and register your own handler.

cli/internal/hooks/installer.go
```diff
@@ -491,62 +571,52 @@ func requoteDevexpHooks(hooksMap map[string][]hookEntry, registry Registry, repo
...
-					msg := fmt.Sprintf("%s: %s (duplicate of the registered command)", event, filepath.Base(p))
-					if dryRun {
-						ui.DryRun("remove " + msg)
-						keptCmds = append(keptCmds, h)
-						continue
-					}
-					ui.Removed(msg)
-					changed = true
-					continue
+				if h.devexpForm() {
```
alexandrocuma, line 579: Test gap: changing this to `if true {` survives the suite. With that mutation, an exec-form handler (`args`) whose command equals the quoted devexp command makes requote **drop** the legacy spelling instead of rewriting it, so its fields are lost. Please add a case: a legacy bare devexp command plus an `args` handler with the quoted command in the same event.

cli/internal/hooks/settings.go
```diff
@@ -0,0 +1,416 @@
...
+		return encodeJSON(struct {
+			Matcher string    `json:"matcher"`
+			Hooks   []hookCmd `json:"hooks"`
+		}{e.Matcher, e.Hooks})
+	}
+	obj, err := e.fields.withString("matcher", e.Matcher)
+	if err != nil {
+		return nil, err
+	}
+	if obj.has("hooks") || len(e.Hooks) > 0 {
```
alexandrocuma, line 211: Test gap: changing this to `if true {` survives. A user entry without `hooks` (e.g. `{"matcher":"X"}`) would then gain `"hooks": null` whenever install writes. Worth pinning in `TestInstallClaude_KeepsUserHookFields`.

CHANGELOG.md
```diff
@@ -7,6 +7,58 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    stops with an error naming it. Earlier releases replaced it with a file
+    holding only the hooks.
+  - `uninstall.sh` applies the same rules: it keeps pre-existing empty events,
+    no longer escapes non-ASCII text, splices the new `hooks` value into the
+    original text, and skips entries and handlers that aren't objects instead
+    of failing with a traceback.
+- **Existing installs never picked up a registry matcher change.** `devexp
+  install` skipped a hook whose command was already registered, whatever
+  matcher its entry had, so a hook whose registry matcher grew (for example
+  `secret-in-write-guard` gaining `MultiEdit|NotebookEdit`) kept firing only
```
alexandrocuma: This example didn't happen: `secret-in-write-guard` is still `Write|Edit` in `hooks/registry.json`. The real case is `secret-guard` going from `Read` to `Read|Bash` (89fae9b). Suggest using that one, here and in install.md:69.

alexandrocuma (commented):
## Review: PR #147 (#137, #138, matcher sync)

**Recommendation: Request changes.** The findings below are small, but they are inside this diff. The approach is sound: splicing only the `hooks` value, the `type == "command"` + no-`args` ownership rule, and the narrower `isStaleDevexpHook`. The tests are strong.

### Findings in this diff (fix now)
1. **`uninstall.sh` can write invalid JSON.** Take a user handler with `"timeout": 1e400` (valid JSON). Python reads it as `inf` and writes back `"timeout": Infinity`. The safety check passes because `inf == inf`, so Claude Code is left with a settings.json it can't parse. Python also accepts `NaN`/`Infinity` input, which Go rejects. Inline on uninstall.sh:498.
2. **`uninstall.sh` turns CRLF into LF across the whole file.** Text-mode `open()` translates newlines on read and write, so every line outside `hooks` changes. That contradicts install.md ("every other byte ... is kept"). Inline on uninstall.sh:412.
3. **Go install aborts on a valid file with an out-of-range number.** `{"a": 1e400}` anywhere makes `render` fail with `json: cannot unmarshal number 1e400 into Go value of type float64`. The file stays safe, but the whole `devexp install` stops (opencode target, manifest save). Inline on settings.go:351.
4. **Matcher sync overwrites users' deliberate matchers** on devexp handlers on every install. It also narrows an entry that has no `matcher` (matches all) and adds `"matcher"` after `"hooks"`. This is a behaviour change, but the PR's "Behaviour changes" section doesn't list it. Inline on installer.go:241.
5. **Docs/CHANGELOG.**
   - The CHANGELOG example (`secret-in-write-guard` gaining `MultiEdit|NotebookEdit`) never happened; the registry still says `Write|Edit`. The real case is `secret-guard` going from `Read` to `Read|Bash` (89fae9b).
   - install.md:126 and the PR body ("byte-for-byte") overstate what is kept inside `hooks`: when install writes, the whole value is re-indented and escaped keys are re-encoded.
6. **Test gaps (mutations that survive).** See the inline comments on installer.go:579 and settings.go:181. Also, ignoring the read error at installer.go:135 (`if false`) survives.

### Mutation results (my own, on a scratch copy)
- Go, `go test ./internal/hooks`: 22/27 killed.
- Survivors:
  - render decode check disabled (expected: the check can't fail by construction)
  - entry always writes a `hooks` key
  - requote `present` counts exec-form handlers
  - a moved handler keeps the last duplicate instead of the first
  - read error ignored
- python, `uninstall.test.sh`: 8/10 killed.
- Survivors:
  - decode check disabled
  - `top_level_member` picks the first duplicate `hooks` key instead of the last (the decode check still catches this at runtime)

### Pre-existing, outside this diff (follow-up candidates, not filed)
- Non-atomic writes of settings.json in Go and python (already noted in the PR).
- Dry-run doesn't preview matcher updates, and install.md "Previewing an update" says it shows every update (already noted).
- A stale registration from **another** install root whose script was removed from the registry is never pruned by install or removed by uninstall. `isStaleDevexpHook` requires `repoDir`, and the foreign/uninstall matchers require a basename the registry still lists.
- `InstallClaude` checks `hook.Enabled`, not `EnabledFor(TargetClaudeCode)`.
- The python step doesn't catch `RecursionError` on deeply nested JSON, so `set -e` aborts uninstall.

### Verified
- `go test -race ./...` passes, and `uninstall.test.sh` passes 115/0.
- Probes against the Go installer:
  - kept intact: CRLF outside `hooks`, escaped top-level `"hook\u0073"`, duplicate top-level `hooks` (last one edited), `1e3`/`-0` inside handlers, compact/indented/multiline insert layouts, `null` events, entries without `hooks`
  - refused: BOM and trailing content
  - idempotent: the second install writes nothing in every case
- A symlinked settings.json keeps its link and file mode (Go and python).
- `isStaleDevexpHook`:
  - pruned: quoted repo paths containing `'`, spaces and `"`, and the legacy bare form
  - not pruned: `//`, `/./`, a trailing space, the double-quoted form
- Go/python parity on ownership: `args: null` and a duplicate `type` key.

docs/guides/install.md
```diff
@@ -116,6 +116,16 @@ Each is rewritten in place (event, matcher and position kept). If the event alre
...
+- **A handler with `args`, or of a type other than `command`, is the user's**, whatever path it names. With `args`, Claude Code spawns `command` directly without a shell, so devexp never matches, re-quotes, prunes or removes it, and doesn't count it as the hook's registration.
+- **Removing devexp's handlers** drops an entry only when that leaves it with no handlers, and an event only when that leaves it with no entries. An entry with `"hooks": []` or an event with `[]` that was already empty stays.
+- **Only the value of the top-level `hooks` key is rewritten.** Every other byte of the file, other keys, their formatting and order, and the trailing newline, is kept. The `hooks` value is written in the indentation of the line its key is on (two spaces, four, tabs), or compact in a compact file. Install keeps string escapes and number spellings inside `hooks` as they were; `uninstall.sh` writes them as python's `json` does (`\u00e9` becomes `é`, `2.50` becomes `2.5`). A file without `hooks` gains the key after its last member; a new file is written as earlier releases wrote it. Events devexp adds go after the existing ones, sorted. Nothing is written when there's nothing to change.
```
alexandrocuma: Overstated in three ways:
- When install writes, the whole `hooks` value is re-indented, so a user's one-line handler is expanded.
- Escaped keys are re-encoded (`"comm\u0061nd"` → `"command"`), so only *values* keep their escapes.
- `uninstall.sh` rewrites CRLF files as LF (see the uninstall.sh:412 comment).

The PR body's "keeps handlers and entries byte-for-byte" has the same problem.

uninstall.sh
alexandrocuma: Fixed in 4366f42. The python step now parses with a `parse_constant` that raises and a `parse_float` that rejects non-finite values, dumps with `allow_nan=False` as a backup, and leaves the file untouched with a `[skip]` saying why. Tests in `uninstall.test.sh`: `1e400`, `-1e400`, `NaN`, `Infinity`, `-Infinity` files unchanged, with the reason checked. Removing either parse hook is killed.

uninstall.sh
alexandrocuma: Fixed in 4366f42 (python) and 5e496ab (Go). Python reads and writes with `newline=""`. Both write the rewritten `hooks` value with the file's own line ending (that of its first line), so a CRLF file stays CRLF throughout. Tests: a CRLF round trip in `uninstall.test.sh` (with a CR/LF count, since the fixture helper had been normalising newlines) and three CRLF cases in `TestInstallClaude_SettingsLayout`. Documented in install.md.

cli/internal/hooks/settings.go
alexandrocuma: Fixed in 5e496ab. `want`/`got` are decoded with `UseNumber()` (plus a trailing-data check), so numbers compare as text. New `TestInstallClaude_SettingsLayout` case: `1e400` at the top level and `-1e400` in a handler, install succeeds and both are kept. A comment says the check cannot fail by construction.

cli/internal/hooks/installer.go
alexandrocuma: Kept the sync. 9ace88f writes an added `matcher` before `hooks` (new `TestInstallClaude_SyncsRegistryMatcher` case for an entry without a matcher). The behaviour change is now in the CHANGELOG, install.md and the PR body's "Behaviour changes": install enforces the registry matcher on devexp's own handlers; to keep a custom one, disable the hook (`hooks.disabled`), remove devexp's entry, and register your own command.

cli/internal/hooks/installer.go
alexandrocuma: Pinned in d90e338: `TestInstallClaude_RequoteKeepsHandlerFields` now has an exec-form handler (`args`) whose command equals the quoted devexp command, in the same event as the legacy bare spelling. The `if true` mutation is killed.

cli/internal/hooks/settings.go
alexandrocuma: Pinned in d90e338: `TestInstallClaude_KeepsUserHookFields` now has an entry without `hooks` (`{"matcher": "idle"}`). The `if true` mutation is killed.

CHANGELOG.md
alexandrocuma: Fixed in 9ace88f: the CHANGELOG and install.md now use `secret-guard` going from `Read` to `Read|Bash`. (#145 has since made the `secret-in-write-guard` change real on main. The PR body mentions both, and I checked the sync against the real registry after merging.)

docs/guides/install.md
alexandrocuma: Fixed in c3a117a. install.md now says every byte outside `hooks` is kept, line endings included. Inside `hooks`, fields, order and values are kept, but the value is re-indented (one-line handlers expand), written with the file's line ending, and escaped keys are re-encoded. uninstall normalises escapes and numbers. The CHANGELOG and PR body use the same wording.

alexandrocuma:
Review findings that had no inline thread:
- **Moved handler keeps the last duplicate:** pinned in d90e338 (`TestInstallClaude_SyncsRegistryMatcher`, "shared, twice: the first copy moves out").
- **Read error ignored (installer.go:135):** pinned in d90e338 (`TestInstallClaude_UnreadableSettings`). The file is write-only, so a rewrite would succeed if the error were ignored.
- **`top_level_member` picks the first repeated `hooks` key:** pinned in d90e338 (`uninstall.test.sh`, "a repeated hooks key: the last one is edited").
- **Decode check disabled (Go and python):** still survives, because the check cannot fail by construction. Code comments in `render` and in the python step now say so.

Merged main again (#145, #149) in fcb82d9. All suites are green and the branch is pushed.

Issue #146, opened 2026-09-17
opencode dangerous-cmd-guard: regex matching can be slow on crafted long lines


alexandrocuma:
Found during #100 (pre-existing, out of its scope).

## Problem
The opencode `dangerous-cmd-guard` module evaluates its block patterns with JavaScript's backtracking regex engine. On crafted long lines some patterns can take a long time, delaying every guarded tool call (and, if the plugin host times out, affecting whether the guard runs). The Claude Code hook uses `grep` and isn't affected the same way.

## Acceptance Criteria
- [ ] Measure worst-case matching time per pattern on adversarial inputs (e.g. 1 MB lines)
- [ ] Rewrite or bound patterns (linear-time forms, input length caps that fail closed) so matching stays well under the hook budget
- [ ] Decisions unchanged on the existing test corpus; timing regression test

## Release Targets
- `cli`, `toolkit-clone`

alexandrocuma:
## Groomed plan (validated against `10ad67c`)

Verdict: needs a correction, then ready.

**Confirmed.** Several opencode patterns backtrack on crafted single lines: a repeated anchor followed by an unbounded scan, flag letters split two ways, and "X, then anything, then Y". The worst pattern grows faster than quadratically: seconds at about 10 KB and much worse beyond.

**Correction.** The Claude Code hook isn't immune. BSD `grep` stayed linear in local measurements. But the masking step both implementations share (#100) is quadratic in the number of pipeline stages: each stage rescans the rest of the pipeline. A 1 MB command takes tens of seconds in the sh hook.

**Plan (decisions unchanged):**
- **JS matching.** Decide one line at a time, with each rule written in a form that is linear in the line length:
  - flag letters split at the first occurrence;
  - "prefix, then gap, then suffix" resolved from the first prefix in each segment, with a cached suffix search;
  - the `.claude…/*` scan answered from a single per-line table.
- **Masking (py + js).** Compute whether a pipeline's later stages are safe once per pipeline instead of once per stage. Hoist the body-reader check out of the heredoc loop. Bound the JS character search the way Python's already is.
- **No input length cap.** A cap would block legitimate long heredocs and change decisions.
- **Measure** every rule before and after on crafted lines up to 1 MB, for JS and for the sh hook (python + `grep`, BSD and GNU).
- **Tests.** Add a timing regression test with a bound that the old JS exceeds. Existing suites must pass unchanged, and the sh/js parity fuzz must show 0 differences.

PR #149, opened 2026-09-17
fix(hooks): dangerous-cmd-guard stops blocking mentions and closes missed forms (#100)

### LINKED ISSUE #100: dangerous-cmd-guard blocks bare mentions of destructive commands, not just real ones
Found during the v0.7.0 release. `dangerous-cmd-guard` blocked a release step whose command contained **no destructive operation** — only an `echo` printing the text of one as a hint to the user:

```bash
echo "  LOCAL ONLY — remote untouched. Tag still revertible:"
echo "    git reset --hard origin/main"     # <- text inside a quoted echo
```

```
[devexp dangerous-cmd-guard] Blocked: git reset --hard will permanently
discard all uncommitted changes.
```

Nothing would have been discarded. The guard matched the literal string anywhere in the command line, including inside a quoted `echo` argument.

**It then blocked the report.** The first attempt to file this very issue was refused too, because the issue body quotes the command being described. A guard that cannot be documented — in a runbook, a `--help` string, or an error message that names the command it warns about — is the strongest form of this complaint.

## This is the same bug class as #89

#89 fixed exactly this shape in `secret-guard`: every shell token was treated as a path, so program text and prose that merely *contained* a secret-like string were refused — `jq -r '.key'` was blocked as a key file, and a heredoc naming a dotenv file blocked the whole command. The fix was to count a token as a path only when it plausibly is one: strip heredoc bodies, skip tokens carrying program syntax or spaces, require a stem for extension matches.

`dangerous-cmd-guard` never got the equivalent treatment. It still substring-matches against the whole command line.

## Why it matters beyond the annoyance

The block landed mid-release, between the merge and the tag — a guard firing on documentation at the least forgiving moment. The practical risk is not the interruption but the adaptation: a guard that fires on prose teaches the user to phrase around it, and a safety hook that trains evasion is worse than one that does not exist. This is the same argument #92 made for fail-closed behaviour, pointed the other way: a guard must be both hard to bypass *and* accurate enough to be worth not bypassing.

## Suggested direction

Match against the command actually being invoked rather than the raw string:

- Strip quoted string literals and heredoc bodies before matching — text inside `'...'` or `"..."` that is an argument to `echo`/`printf`/`cat` is not an invocation.
- Anchor patterns to a command position (start of line, or after `;`, `&&`, `||`, `|`) rather than matching anywhere in the line.
- Keep the existing patterns and severity — this is about *where* they match, not *what* they match.

The red cases must keep blocking: a real destructive reset, `rm -rf`, a force push — including when chained after `&&`, or nested inside `$(...)`. A test twin asserting both directions, in the style #89 added for `secret-guard`, is the right shape; `hooks/claude-code/*.test.sh` now runs in CI (#88), so the twin will actually execute.

Both implementations need it — the shell hook and the opencode JS twin — as #89 did.

## Acceptance Criteria

- [ ] A quoted mention of a destructive command inside `echo`/`printf`/a heredoc does not block
- [ ] A real invocation still blocks, including after `&&`, `;`, `|`, and inside `$(...)`
- [ ] Both the shell hook and the opencode JS twin are fixed together
- [ ] A `*.test.sh` twin covers the red and green cases, and runs in CI
- [ ] Existing blocked-command coverage stays green

## Non-Goals

- Relaxing which commands are considered dangerous
- Changing the opencode twin's severity model
- Making the guard configurable per-project

## Estimate

S — one predicate plus a test twin. Mirrors #89 closely enough to reuse its test structure.

alexandrocuma:
Closes #100

## What changed

**False positives.** `dangerous-cmd-guard` matched its patterns anywhere in a command line. That refused commands that only *mention* a destructive command, such as an `echo` recovery hint, a commit message or an issue body.
- Both the Claude Code hook and the opencode module now blank out text that provably never runs before matching: `echo`/`printf` arguments, `git commit|tag` messages, `gh` bodies and titles, heredoc bodies fed to a text filter, and comments.
- They only do this when that output can't reach a shell (no redirect into a file, no pipe into anything but a text filter).
- The parser fails closed. Anything it can't classify (unbalanced quotes, `case`, functions, aliases, a shell or interpreter anywhere in the command, a command word that is a path or variable) is scanned unmasked, as in v0.9.0.

**Missed commands (security).** The guard could also miss real destructive commands in some forms: some large commands, commands inside a quoted string handed to another shell or evaluator, and commands split across lines. Both implementations now match these forms, and a matching error blocks instead of allowing.

Matching rules are documented in `docs/reference/hooks.md`.

## Verification
- `dangerous-cmd-guard.test.sh` 160, `.test.js` 161 (from 25 each on `main`). They cover command-position and evaluator block cases, mention allow cases, unparseable input, large input and a grep failure.
- The new block cases fail on the v0.9.0 hook. Allow cases show the wider matching doesn't catch benign paths.
- Parity between the Claude Code hook and the opencode module:
  - Masking fuzz (Python vs JS): 0 differences over 20,000 random single-logical-line inputs.
  - Decision fuzz (the real bash/python/grep hook vs the JS module): 0 differences over 16,005 inputs, 12,019 of them multi-line, built from random LF, CR, CRLF, backslash continuations, tab, VT, FF and NUL between tokens, plus inputs that start a pattern on one line and end it on a later one. 555 of the inputs block on both sides. The corpus is ASCII only, because how grep treats non-ASCII whitespace depends on the locale.
- Mutations of the masking, boundary and continuation rules are each caught.
- After merging `main`: `go vet`, `go test -race ./...`, every hook suite and every installer script test pass.

## Out of scope (follow-ups)
- #146: opencode regex matching can be slow on crafted long lines.
- Text saved to a file and run later isn't seen by a command-line guard. This is documented as a limit.


hooks/opencode/dangerous-cmd-guard.js
```diff
@@ -34,7 +46,7 @@ export const BLOCK_PATTERNS = [
   {
     // Force flag must be an argument of the same push command (no intervening ; | & ), so an
     // unrelated `-f` elsewhere (e.g. `rm -f` in a commit message) no longer false-positives.
-    re: /git\s+push\b[^|&;]*\s(--force-with-lease|--force|-f)(\s|=|$)/m,
+    re: /git\s+push\b[^|&;]*\s(--force-with-lease|--force|-f)(\s|=|$|['"`);&|])/m,
```
alexandrocuma: **Parity / opencode-only false positive (multi-line).** This class (`[^|&;]*`) also matches a newline in a JS regex, so the pattern can span two separate logical command lines: a benign first line plus a later line that happens to complete `…\s(--force…)`. The Claude Code hook greps line-by-line and never joins them, so it allows the same input — a divergence, and an over-block here. Backslash-continued commands are already joined to one line by masking, so constraining this class to exclude newline (match line-by-line, like grep) closes the gap without losing any real single-line invocation. Concrete diverging inputs sent to the requester privately.

hooks/opencode/dangerous-cmd-guard.js
```diff
@@ -4,22 +4,34 @@
...
-    re: /rm\s+-[a-z]*f[a-z]*r\s+(\/\s*$|\/\s+|~\/?(\s|$)|\$HOME(\s|$))/m,
+    re: /rm\s+-[a-z]*f[a-z]*r\s+["']?(\/(\s|$|['"`);&|])|~\/?(\s|$|['"`);&|])|\$HOME(\s|$|['"`);&|]))/m,
     label: "'rm -rf /' or 'rm -rf ~' would wipe your filesystem or home directory",
   },
   {
     // Unanchored wildcard delete in a sensitive dir (/tmp/* , ~/.claude/.../* , or the dir
     // wholesale) — the blanket wipe an empty variable produces. Prefix-anchored globs like
     // /tmp/.deliver-PAY-123-* are allowed (no '*' right after the '/').
-    re: /rm\b[^|]*(\s["']?\/tmp["']?(\/\*|\/?(\s|$))|["']?(\$HOME|~)["']?\/\.claude(\S*\/\*|["']?\/?(\s|$))|\.claude\S*\/\*)/m,
+    re: /rm\b[^|]*(\s["']?\/tmp["']?(\/\*|\/?(\s|$|['"`);&|]))|["']?(\$HOME|~)["']?\/\.claude(\S*\/\*|["']?\/?(\s|$|['"`);&|]))|\.claude\S*\/\*)/m,
```
alexandrocuma: Same newline-crossing class as the force-push pattern: `[^|]*` (and `\S`) match across a newline in JS but not under line-oriented grep, so a multi-line command can be blocked here by the opencode module yet allowed by the Claude Code hook. Constrain to a single line for parity.

hooks/opencode/dangerous-cmd-guard.js
```diff
@@ -4,22 +4,34 @@
...
+ * Tests: node hooks/opencode/dangerous-cmd-guard.test.js
  */
 
+// A target ends at whitespace, end of line, or a character that closes the word in shell
+// syntax: ' " ) ` ; & | — so `sh -c 'rm -rf /'`, `$(rm -rf ~)` and `git push --force;` match.
+// A letter, digit, '/', '.', '-' or '*' continues it.
 export const BLOCK_PATTERNS = [
   {
-    re: /rm\s+-[a-z]*r[a-z]*f\s+(\/\s*$|\/\s+|~\/?(\s|$)|\$HOME(\s|$))/m,
+    re: /rm\s+-[a-z]*r[a-z]*f\s+["']?(\/(\s|$|['"`);&|])|~\/?(\s|$|['"`);&|])|\$HOME(\s|$|['"`);&|]))/m,
```
alexandrocuma: Same root cause via `\s+`: in JS `\s` includes a newline, so `rm -rf` on one line and a sensitive target on the next is matched here but not by the line-oriented shell hook. Use a horizontal-whitespace class (or otherwise forbid the newline) so both implementations agree. Applies to the mirrored pattern just below as well.

alexandrocuma (commented):
## Security review — dangerous-cmd-guard (#100 / PR #149)

**Recommendation: Request changes** (one in-diff parity finding; safe direction, but the PR's central claim is sh/js parity and it does not fully hold). No exploitable bypass of the guard was found.

This is a careful, well-tested change. I verified the core security invariant holds: masking only blanks provably-inert text (`echo`/`printf` args, `git commit|tag` messages, `gh` bodies/titles, heredoc bodies to text sinks, comments), and only when output can't reach a shell. A shell/interpreter, a pipe to a non-sink, or a redirect to a file anywhere causes the region to be scanned unmasked. `$(…)`/backticks inside a blanked argument are still walked and scanned. Nested `echo`, evaluator strings (`bash -c`, `sh -c`, `eval`, `ssh`, `psql -c`, `env -S`), `xargs`, `nohup`, `time`, `exec`, wrappers, process substitution, and heredocs-to-a-shell all still block. Fail-closed paths (malformed JSON, empty stdin, interpreter/grep error) all exit 2. Adversarial inputs (deep `$(` nesting, 100k words, 500k-char lines) stay well under a second on both parsers — nothing materially new beyond #146.

### What I ran
- **Masking parity** (`scan_text` vs `maskInert`) over **22,444 inputs** (structured + 20k random fuzz): **0 differences**.
- **Block-decision parity** over the same corpus: 0 differences on random fuzz — but see the finding below: a targeted multi-line class does diverge.
- **Mutation testing** on a copy: weakening `downstreamOk`, dropping the redirect/pipe context check, widening the force-push class, and removing the interpreter guard were each caught by the suites (4 / 8 / 1 / 3 failures). The tests genuinely pin the claims.

### 🟡 Finding — sh/js parity does not hold for multi-line commands (over-block, opencode-only)
See inline comments. Class: several JS `BLOCK_PATTERNS` use character classes that also match a newline, while the Claude Code hook greps line-by-line. A command whose first line is benign and whose *later* line coincidentally completes one of those patterns is blocked by the opencode module but allowed by the Claude Code hook. This is a false positive (over-block, safe direction — no destructive command slips through), but it is a real parity divergence, and the random-fuzz corpus behind the "0 differences over 20,000 inputs" claim does not contain this shape. Suggested fix: constrain those classes to a single line so both implementations match line-by-line.

### Follow-up candidates (pre-existing on `main`, not filed)
- **Target boundary vs. shell expansion.** The boundary set was widened to quotes/parens/operators, but a target immediately followed by a shell *expansion* (rather than a literal boundary char) is still not treated as a word end, so a sensitive-dir delete written with an expansion abutting the target can slip past the new wholesale-dir rule. Same on `main`. Worth a boundary follow-up; concrete cases sent to the requester privately.
- **Encoded/indirection payloads** (e.g. a decode piped into command substitution) are invisible to any command-line matcher — already acknowledged in the PR's "saved to a file and run later" limitation.

### Docs/CHANGELOG
Accurate and class-level; `docs/reference/hooks.md` documents the match/blank/scan rules clearly. Consider softening the CHANGELOG's "parity fuzz shows 0 differences" to note the corpus is single-logical-line-oriented, given the finding above.

hooks/opencode/dangerous-cmd-guard.js
alexandrocuma: Fixed in c5cc8c6. Every opencode pattern now stays on one line, like the Claude Code hook: negated classes exclude the newline. Backslash-continued commands are still joined first, and the continuation tests still block. Mirrored cases added to both suites.

hooks/opencode/dangerous-cmd-guard.js
alexandrocuma: Fixed in c5cc8c6, same change. This class and the "any character" match are now single-line, and every other pattern in the module was checked for the same issue. Covered by the mirrored parity cases and a multi-line decision fuzz that compares the real hook with the module.

hooks/opencode/dangerous-cmd-guard.js
alexandrocuma: Fixed in c5cc8c6. Whitespace is now a class that excludes the newline but still counts CR, tab, VT and FF, as grep does. The mirrored pattern below uses the same class.

PR #153, opened 2026-09-17
ci: run govulncheck on every push, PR and release (#141)

### LINKED ISSUE #141: CI: run govulncheck on every push and PR

alexandrocuma:
Closes #141

## What changed
- **`scripts/govulncheck.sh`** (new) is the single definition of the scan. It holds the pinned govulncheck version (**v1.8.0**), stages assets, installs govulncheck into a temporary `GOBIN` removed on exit, and runs `govulncheck -show verbose ./...` in `cli/` for every platform the release ships: each `goos` x `goarch` in `.goreleaser.yaml`, with its `CGO_ENABLED=0`.
  - Exit 0: nothing called (uncalled findings are printed).
  - Exit 3: a vulnerability in code the CLI calls, on any platform.
  - Any other failure exits 1 and is labelled an infrastructure failure (network, vuln.go.dev, install, config parsing).
- **`ci.yml`:** new `govulncheck` job on every PR and every push to `main` (`contents: read`, no persisted credentials). Sets up Go from `cli/go.mod` (go1.26.8, `GOTOOLCHAIN=local`) and runs the script.
- **`release.yml`:** workflow default is now `contents: read`. A separate `govulncheck` job runs the script; `goreleaser` has `needs: govulncheck` and job-level `contents: write`, steps unchanged from `main`.
- **`.goreleaser.yaml`:** `release.use_existing_draft: true`. `/release` creates the GitHub Release as a draft; goreleaser finds it by name (`{{.Tag}}`), keeps its notes, uploads, then publishes. A release is never public or Latest without assets, and a failed run leaves only a draft. Verified against goreleaser v2.18.2 source (`findDraftRelease`, `getReleaseNotes` keep-existing, `doPublish` → `PublishRelease`).
- **Docs:**
  - `testing.md`: "Vulnerability Scan" (platforms, exit codes, infra failures → `gh run rerun <id> --failed`, how to fix a finding, how to bump the version), checklist item.
  - `release.md`: draft-based Cut, Build section with the two-job flow and recovery (infra → re-run; finding or unrecoverable failure → delete the draft, fix, cut the next patch), Distribute/Promote/post-release verification (`isDraft` false, Latest).
  - `setup.md`, `workflows.md`, `overview.md`, `conventions.md`, `CLAUDE.md`: command row, script listing, line references.
  - CHANGELOG.

## Verification
- **Local, all platforms:** `./scripts/govulncheck.sh` → `No vulnerabilities found.` on darwin/amd64, darwin/arm64, linux/amd64, linux/arm64 (16 modules, go1.26.8 stdlib), exit 0.
- **Failure demos through the script (scratch copies):**
  - `x/text` v0.38.0, not called: exit 0, GO-2026-5970 listed under package results.
  - Same, plus a call into `unicode/norm`: **exit 3** on all four platforms.
  - Same call only in a `_darwin.go` file: **exit 3**, called on darwin/* only.
  - Toolchain go1.26.3, unmodified code: **exit 3** (GO-2026-5037 via `config.Load`; linux/* additionally GO-2026-5972 via `mcp.AddClaude`, darwin/* not).
  - vuln.go.dev unreachable: **exit 1**, `infrastructure failure, not a finding`.
- **Linting:** actionlint 1.7.12 passes on both workflows; `goreleaser check` (v2.18.2, checksum-verified) validates `.goreleaser.yaml`; `gofmt`, `go vet`, `go test -race ./...` pass.

## Out of scope (follow-ups being filed)
- A weekly `schedule:` run, to catch advisories published while `main` is quiet.
- Gating `release.yml` on the test suites via `workflow_call`.
- Stale `CHANGELOG.md` line references in the docs (`CHANGELOG.md:147`, `:60`).
- `skills/release/SKILL.md` Phase 6 should follow the release guide's Cut when it defines how the release object is created.


docs/guides/release.md
```diff
@@ -21,23 +21,31 @@ Asset edits reach the two targets at different times. Clone users get them on th
...
+# CI runs: ./scripts/stage-assets.sh → govulncheck -show verbose ./... (in cli/) → ./scripts/stage-assets.sh (goreleaser before-hook) → goreleaser release --clean   # source: .github/workflows/release.yml:24-43, .goreleaser.yaml:5-7
 ```
 
+The `govulncheck` step is the same check as the `govulncheck` job in `ci.yml`. If it finds a vulnerability in code the CLI calls, the job stops before goreleaser, so no GitHub Release assets are uploaded for the tag (uncalled findings are only printed). The tag stays pushed and, if `/release` already created it, so does the GitHub Release object, with no assets. `releases/latest` then points at a release `remote-install.sh` can't download from, so act straight away:
+
+1. Mark the empty release as a pre-release (`gh release edit v<version> --prerelease`) so `latest` falls back to the previous one.
+2. Fix the finding on a branch as described in [`../development/testing.md`](../development/testing.md#vulnerability-scan) and merge it.
+3. Cut `v<next patch>`. Don't move or re-push the failed tag (tags are never moved — see Rollback).
```
alexandrocuma: **Finding 1 (important, in scope).** The recovery here is manual, and it only starts once someone notices. Until then, `remote-install.sh` resolves `releases/latest` to this tag (`scripts/remote-install.sh:53`) and 404s on the asset for every new user. The asset-less window also exists on a *successful* release, and the scan adds about 20s to it (`go install` took about 17s in this PR's CI log).

This can be avoided instead of recovered from. goreleaser v2.18.2 always creates its own release as a draft, uploads, then publishes it. It only uploads into a live release because `/release` pre-created a published one, which `GetReleaseByTag` finds while keeping `Draft=false`. Suggested:

- `.goreleaser.yaml`:
  ```yaml
  release:
    use_existing_draft: true   # publish the draft /release created, keeping its notes (mode: keep-existing)
    github: …
  ```
- line 23: `gh release create v<version> --draft --title v<version> --notes-file <section> --verify-tag`
- these lines: a blocked scan leaves only an invisible draft and Latest doesn't move. `gh release delete v<version> --yes` (keeps the tag), fix, then cut `v<next patch>`.
- line 53: update "sets no draft".

Details and alternatives are in the review summary.

.github/workflows/ci.yml
```diff
@@ -43,3 +43,25 @@ jobs:
...
+      - name: stage assets
+        run: ./scripts/stage-assets.sh
+      - name: install govulncheck
+        run: go install "golang.org/x/vuln/cmd/govulncheck@${GOVULNCHECK_VERSION}"
+      # Fails (exit 3) only on vulnerabilities in code the CLI calls;
+      # -show verbose also lists uncalled findings without failing.
+      # setup-go puts $(go env GOPATH)/bin on PATH.
+      - name: govulncheck
+        working-directory: cli
+        run: govulncheck -show verbose ./...
```
alexandrocuma: **Finding 2 (important).** This scans only the runner's platform: linux/amd64 with `CGO_ENABLED=1` (from `go env` in the job log). goreleaser ships darwin and linux, amd64 and arm64, with `CGO_ENABLED=0` (`.goreleaser.yaml:14-21`). govulncheck's symbol reachability is per build configuration, and here the file sets really differ. `GOOS=darwin` vs `linux` `go list -deps` compiles `fsnotify/backend_kqueue.go` instead of `backend_inotify.go`, `crypto/x509/root_darwin.go` instead of `root_linux.go`/`root_unix.go`, readline `term_bsd.go` instead of `term_linux.go`, and different `x/sys/unix`, `os`, `net` and `syscall` files. A called vulnerability in darwin-only code would pass both this job and the release gate. I checked that govulncheck v1.8.0 honours `GOOS`: both `GOOS=linux` and `GOOS=darwin` `CGO_ENABLED=0` scans run clean on this branch, and each scan takes about 2s.

```yaml
      - name: govulncheck
        working-directory: cli
        env:
          CGO_ENABLED: "0"   # match .goreleaser.yaml
        run: |
          for goos in linux darwin; do
            echo "== GOOS=$goos"
            GOOS=$goos govulncheck -show verbose ./...
          done
```
(`bash -e` stops at the first non-zero exit and keeps exit 3.) Same in `release.yml`, and in the local command in `testing.md` and `setup.md` if you want them to match.

.github/workflows/release.yml
```diff
@@ -21,6 +21,21 @@ jobs:
...
+
+      - name: install govulncheck
+        run: go install "golang.org/x/vuln/cmd/govulncheck@${GOVULNCHECK_VERSION}"
+        env:
+          # Keep in sync with .github/workflows/ci.yml.
+          GOVULNCHECK_VERSION: v1.8.0
+
+      - name: govulncheck
+        working-directory: cli
+        run: govulncheck -show verbose ./...
```
alexandrocuma: **Finding 3 (important, least privilege).** These steps inherit the workflow-level `permissions: contents: write` (lines 8-9). `actions/checkout` persists that token for later steps by default. So `go install` (building x/vuln, x/tools, x/telemetry) and the govulncheck binary run beside a token that can push tags and edit releases. The actual risk is low, since it's Go-team code verified through sumdb, but nothing here needs write access. A separate job fixes that, and it also puts goreleaser's job back to its `main` shape, with no second staging:

```yaml
permissions:
  contents: read

jobs:
  govulncheck:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v6
        with:
          persist-credentials: false
      - uses: actions/setup-go@v6
        with:
          go-version-file: cli/go.mod
          cache-dependency-path: cli/go.sum
      - name: stage assets
        run: ./scripts/stage-assets.sh
      - name: install govulncheck
        run: go install "golang.org/x/vuln/cmd/govulncheck@v1.8.0"   # keep in sync with ci.yml
      - name: govulncheck
        working-directory: cli
        run: govulncheck -show verbose ./...   # + GOOS loop, see ci.yml comment

  goreleaser:
    needs: govulncheck
    runs-on: ubuntu-latest
    permissions:
      contents: write
    steps: # unchanged from main
```
Both jobs check out the same tag SHA, so there's no gap between scan and build. A failed release also shows up as a red `govulncheck` job rather than a failed step inside `goreleaser`. The line references in `release.md` (`release.yml:24-43, 35-37, 45`) and `setup.md` (`release.yml:24-37`) would need updating.

docs/development/testing.md
```diff
@@ -68,6 +68,30 @@ Gotchas when running tests:
...
+```bash
+./scripts/stage-assets.sh && (cd cli && go run golang.org/x/vuln/cmd/govulncheck@v1.8.0 -show verbose ./...)
+```
+
+When it fails:
+
+1. Read the finding: the advisory ID, `Found in` / `Fixed in`, and the trace showing which of our calls reaches it.
+2. **Standard library** (`Found in: <pkg>@go1.x.y`): bump only the `toolchain` line in `cli/go.mod` to a supported patch at or above `Fixed in` (see [`../guides/workflows.md`](../guides/workflows.md), Go toolchain). **Module**: `cd cli && go get <module>@<fixed version> && go mod tidy`, commit `go.mod` and `go.sum` together.
+3. Re-run the command above until the symbol results are clear, run the Go tests, and land the bump as its own `fix:` commit naming the advisory, with a `CHANGELOG.md` entry.
+4. No fixed version yet: don't disable the job. Record the advisory in an issue; if the called path can be avoided in our code, do that.
```
alexandrocuma: **Finding 4 (minor).** "When it fails" covers only findings (exit 3). The job also fails with exit 1 when vuln.go.dev or the module proxy is unreachable. In `release.yml` that blocks the tag with nothing to fix in the code. Suggest one line: *"Exit 1 with a fetch or load error (no `=== Symbol Results ===`) is infrastructure, not a finding. Re-run the failed job (`gh run rerun <run-id> --failed`)."* It's worth mirroring in `release.md`'s Build paragraph, where the guide otherwise says not to run the build again.

alexandrocuma (commented):
## Review: ci: run govulncheck on every push, PR and release (#141)

**Recommendation: Request changes** (posted as a comment). The scan is set up correctly and CI is green. Three things inside the diff should change before merge: the empty-Latest release problem this PR creates, the scan only covering linux/amd64, and the scan running next to a `contents: write` token.

### What I checked
- **CI on this PR:** `test`, `hooks` and `govulncheck` all pass (run 35180850010). In the govulncheck job log, setup-go resolved `1.26.8` from the `toolchain` line and exported `GOTOOLCHAIN: local`. `go install` downloaded x/vuln v1.8.0, x/telemetry, x/mod, x/tools and x/sync (`GOSUMDB=sum.golang.org`, `GONOSUMDB` empty) in about 17s. `govulncheck` was on PATH and scanned 11 packages, 16 modules and the go1.26.8 stdlib: "No vulnerabilities found."
- **PATH, GOTOOLCHAIN, caching:** all correct. setup-go adds `$(go env GOPATH)/bin` to PATH, and there's only one setup-go per job, which is what `workflows.md` asks for. `GOVULNCHECK_VERSION` is safe in job env; only `GOTOOLCHAIN` must stay out of it. The job gets a cache hit on the key the `test` job shares, so govulncheck's modules are never saved and get downloaded again on every run (about 17s). That's fine.
- **Dirty tree and double staging in release.yml:** no problem. `stage-assets.sh` is idempotent (`rm -rf` then rsync), and every staged path is gitignored. In a fresh clone of `fix/141`, I ran stage → `govulncheck ./...` → stage again, and `git status --porcelain` was empty, so goreleaser's dirty check passes. `go.sum` isn't touched.
- **Supply chain:** pinning `@v1.8.0` is enough. The proxy serves immutable content (`.info` Origin hash `709015412431`), and every module is checked against the sum.golang.org transparency log. A re-pointed upstream tag can't change what gets built. Don't switch to a `tool` directive in `cli/go.mod`. It would pull x/tools and x/telemetry into the CLI's module graph, and MVS would raise `golang.org/x/sys` in the **shipped binary** from v0.46.0 to at least v0.48.0 (x/vuln v1.8.0's requirement). Also avoid `golang/govulncheck-action`, whose default version is `latest`.
- **Docs:** every changed line reference resolves on this branch (`ci.yml:47-67`, `release.yml:8-9, 24-37, 24-43, 35-37, 45`, `.goreleaser.yaml:5-7`, `remote-install.sh:51-55`). x/vuln v1.8.0's go.mod does say `go 1.26.0`. I ran the local command from `testing.md` in a copy of the branch with a local **go1.26.3** (`GOTOOLCHAIN=auto`). It scanned the go1.26.8 stdlib, exited 0 and matched CI's output. `go run` of a program that exits 3 does return 1, as documented.

### Findings (most severe first)
1. **A blocked scan leaves an asset-less release as Latest** (`docs/guides/release.md:23, 41-45`). In scope, because this PR adds the failure path. On a *successful* release it also adds about 20s to the window where Latest has no assets. Recommended fix below and inline.
2. **Only linux/amd64 with cgo on is scanned, but the release ships darwin and linux with `CGO_ENABLED=0`** (`ci.yml:65-67`, `release.yml:35-37`). Inline.
3. **The scan runs in the `contents: write` job** (`release.yml:8-9, 24-37`). Move it to its own read-only job and add `needs:`. Inline.
4. Minor: **a vuln.go.dev or network failure (exit 1) also blocks the release**, and the docs only describe findings (exit 3). Inline on `testing.md`.

### Recommended fix for (1): let goreleaser publish a draft that `/release` creates
I checked goreleaser v2.18.2 (`internal/client/github.go`, `internal/pipe/release/release.go`, `internal/client/release_notes.go`):
- `CreateRelease` always starts as a **draft**, uploads the artifacts, then `PublishRelease` undrafts it (unless `release.draft: true`). So a release goreleaser creates is never visible without assets.
- `findRelease` uses `GetReleaseByTag` by default. That endpoint returns **published** releases only. So today it finds the published release `/release` made, and keeps `Draft=false` (`data.Draft = new(release.Draft)`): assets are uploaded to a release that is already Latest.
- With `release.use_existing_draft: true`, it looks for a **draft** by name instead. `name_template` defaults to `{{.Tag}}`, which matches `--title v<version>`. It then keeps the existing notes (default `mode` is keep-existing: `if existing != "" { return existing }`), uploads, and publishes.

The change:
- `.goreleaser.yaml`: under `release:` add `use_existing_draft: true`.
- `release.md:23`: `gh release create v<version> --draft --title v<version> --notes-file <section> --verify-tag`. goreleaser publishes the draft with those notes once the assets are uploaded.
- `release.md:41-45`: on a blocked scan nothing is public and Latest doesn't move. Delete the leftover draft (`gh release delete v<version> --yes`, the tag stays), fix, and cut `v<next patch>`. No pre-release step.
- `release.md:53`: stop saying ".goreleaser.yaml sets no draft".

This also covers goreleaser build or upload failures, which today leave the same empty Latest. The runner-up is to not pre-create the release at all: goreleaser creates it, and after `gh run watch --exit-status` `/release` runs `gh release edit v<version> --notes-file <section>`. That needs no config change, but Latest briefly shows goreleaser's commit-list notes, and the notes are never set if the session dies (the v0.6.0 outcome). `--release-notes` fed from CHANGELOG inside `release.yml` would work too, but it means parsing CHANGELOG in CI, and it doesn't stop `/release` pre-creating. I don't recommend an `if: failure()` step that marks the release as a pre-release: it races `/release`'s create.

**Consistency with `skills/release/SKILL.md`:** Phase 6 hard-codes `gh release create "v<version>" --title … --notes "<changelog entry>"` right after the tag push. It has no `--draft`, no `--verify-tag`, and doesn't point at the guide. This repo's guide already differs (`--notes-file --verify-tag`). Follow-up: Phase 6 should say "if the release guide's Cut section defines how the release object is created, follow it". Don't add `--draft` to the generic skill: without a goreleaser-style publisher, a draft never gets published.

### Out-of-scope list
- Weekly `schedule:`: agree, follow-up. #141's acceptance criteria don't ask for it.
- Gating `release.yml` on tests via `workflow_call`: agree, follow-up.
- govulncheck version pinned in two workflows: fine as a follow-up *if the scan stays one line*. If you adopt (2) and (3), the scan grows to a GOOS loop in two places, and a small composite action (`.github/actions/govulncheck`) in this PR avoids the drift. Note that Dependabot won't bump a `go install …@vX` string.
- Stale `CHANGELOG.md:147`: agree, follow-up. `release.md:22`'s `CHANGELOG.md:60` is stale too, on `main` and on this branch. Line anchors into CHANGELOG go stale with every `[Unreleased]` entry, so point at version headings instead.
- Missing from the list: finding (1), which I think belongs here, and the SKILL.md Phase 6 change, which is a follow-up.

### Done well
- The policy just uses govulncheck's own exit codes, so there's no custom filtering. The PR body shows the failure demos (uncalled → 0, called → 3, the go1.26.3 regression → 3).
- It follows `workflows.md`'s rules: one setup-go per job, no `GOTOOLCHAIN` in env. The docs explain why the toolchain line matters.
- It fixes two stale `release.yml` line references, and every new one resolves.

docs/guides/release.md
alexandrocuma: Fixed in 8d06f23 as recommended. `.goreleaser.yaml` sets `release.use_existing_draft: true`; the Cut in `release.md` now creates the release with `gh release create v<version> --draft --title v<version> --notes-file <section> --verify-tag`. Checked against goreleaser v2.18.2 source (what `~> v2` resolves to today): `findDraftRelease` matches `GetDraft() && GetName() == name` (name_template default `{{.Tag}}`), `getReleaseNotes` keeps the existing body, `doPublish` uploads and then `PublishRelease` sets `Draft: false`. `goreleaser check` (v2.18.2, checksum-verified) passes. The Build section now says: infra failure → `gh run rerun <id> --failed`; exit 3 or unrecoverable goreleaser failure → `gh release delete v<version> --yes` (tag stays), fix, cut the next patch. Distribute/Promote no longer say "no draft", and post-release verification checks `isDraft` false and Latest (not Draft).

.github/workflows/ci.yml
alexandrocuma: Fixed in 8d06f23. The scan now lives in `scripts/govulncheck.sh`, which reads `goos`, `goarch` and `CGO_ENABLED` from `.goreleaser.yaml` and scans each GOOS x GOARCH (4 runs, ~2s each). Locally all four pass. It matters in practice: with the toolchain set to go1.26.3, linux/* reports GO-2026-5972 as called (`mcp.AddClaude` → `exec.Command` → `asn1.Unmarshal`) but darwin/* does not; a call in a `_darwin.go` file is caught only on darwin/*. Exit 3 = called on any platform; any other failure exits 1 as infrastructure.

.github/workflows/release.yml
alexandrocuma: Fixed in 8d06f23. `release.yml` now has workflow-level `permissions: contents: read`; a separate `govulncheck` job (checkout with `persist-credentials: false`) runs `./scripts/govulncheck.sh`; `goreleaser` has `needs: govulncheck` and job-level `contents: write`, with its steps identical to main. The ci.yml job also got `contents: read` and `persist-credentials: false`. The version pin now lives only in the script, so there is no drift between workflows. actionlint 1.7.12 passes; line references in release.md and setup.md updated.

docs/development/testing.md
alexandrocuma: Fixed in 8d06f23. testing.md documents the exit codes: 0 nothing called, 3 called vulnerability, anything else (the script exits 1 and prints `infrastructure failure, not a finding`) is infrastructure: re-run with `gh run rerun <run-id> --failed`. release.md mirrors it in Build, and "do not run it again" now allows re-running failed jobs. Demonstrated locally by routing vuln.go.dev to a dead proxy: every platform reports `fetching vulnerabilities: ... connection refused`, script exits 1.

PR #156, opened 2026-09-17
fix: never remove stale agents or skills through a symlinked directory (#128)

### LINKED ISSUE #128: Stale agent/skill removal follows a symlinked target directory

alexandrocuma:
Closes #128

## What changed
Install and uninstall could delete files through a symlinked `agents/`, `skills/` or `commands/` directory, for example one that points into a dotfiles repo. They could also delete through such a directory when a parent was linked (`~/.claude`, `~/.config`, `~/.config/opencode`). Both now follow the rule opencode's `plugins/` already used: **write through a symlinked target directory, never remove through it.** A directory "behind" a symlink now counts too, for `plugins/` as well.

**Install (`removeStale`, `cli/cmd/backup.go`; `cli/internal/removeguard`)**
- **Symlink or behind one.** When the target directory is a symlink, or `removeguard.BehindSymlink` finds it behind one, stale entries are not removed and not shown as removals in a dry run.
  - The check compares `EvalSymlinks(dir)` with `EvalSymlinks(HOME)` joined with the directory's path under HOME. Because HOME is resolved on both sides, a link above it (macOS `/var` → `/private/var`, a linked `/home`) doesn't count.
  - Every target is built from HOME. There is no `CLAUDE_CONFIG_DIR` or XDG config dir to account for.
  - Install prints one warning per directory, `is a symlink` or `is behind a symlink (it resolves to …)`, listing the leftovers to remove by hand.
- **Manifest.** `removeStale` returns what it kept, each name once, and both installers record installed plus kept names. Leftovers are cleaned up once the link is replaced by a real directory. An entry that couldn't be checked or removed (e.g. permission denied) also stays recorded and is retried.
- **Exact name only.** The directory is read once, and an entry is acted on only when that exact name is listed. A spelling that differs in case or Unicode normalization is dropped from the manifest with a warning. On default APFS, a kept `retired-agent.md` could otherwise later remove a user's `Retired-Agent.md`. A user file with the *exact* recorded name can't be told apart; this is documented.
- **No mid-run redirect.** The directory is pinned with `os.OpenRoot` and checked against the earlier `os.Lstat` with `os.SameFile`. Removals go through `(*os.Root).Remove`/`RemoveAll`, which Go 1.25 has; `go.mod` declares go 1.25.11. If the directory is swapped before it's opened, nothing is removed that run. A swap after that can't redirect a remove.
- **Unchanged:** entries that aren't devexp's (bad name, case twin, same file, symlink, wrong type) are still skipped and dropped.

**opencode `plugins/` (`cli/internal/hooks/opencode.go`)**
- `checkPluginRoots` and `pruneOpencodeDir` apply the same behind-a-symlink rule, resolved from the home that `<home>/.config/opencode/plugins` sits in. A `plugins/` that can't be resolved counts as linked, so nothing is removed from it but writes still go ahead.

**Uninstall (`uninstall.sh`)**
- **What it skips.** It skips entries of an agents or skills directory that is a symlink or behind one (`behind_link`, the same rule via `cd -P`/`pwd -P`), and entries that are themselves symlinks. A symlinked entry is reported as a symlink even inside a symlinked directory, as install does. Both are listed before the confirmation.
- **Re-check after the prompt.** Each entry is checked again right before `rm`, because the confirmation can wait indefinitely. An entry, or its directory, that became a symlink meanwhile is left untouched with a warning.
- **Counting.** "Removed N item(s)" counts only successful removals, and a failed `rm` is reported. Before this PR a failed `rm` was counted.
- **Skills message.** "Skills will be kept (still in use by other installed CLI)" now appears only when that is the reason.

## Verification
- **On `main` / the first push (scratch HOME, real binaries):**
  - Install deleted a stale agent and a whole stale skill from a symlinked dotfiles tree.
  - Install deleted a stale agent under a linked `~/.claude`.
  - Install deleted a user's `Retired-Agent.md` for a recorded `retired-agent.md`.
  - `uninstall.sh` emptied a symlinked agents directory.
- **On this branch** each of those is left alone, with a warning. Against the first push's `uninstall.sh`, the new shell cases fail 21 times; on this branch they pass.
- **Go tests:**
  - Integration: `TestDoInstall_SymlinkedTargetDirs`, `TestDoInstall_ParentSymlinks` (linked `~/.claude` / `~/.config/opencode`, and a HOME behind a symlink), `TestDoInstall_StaleRemovalRealDirs`.
  - `removeStale`: `TestRemoveStale_SymlinkedTargetDir`, `_BehindSymlink`, `_ExactName` (case and normalization, simulated through the `rootLstat` seam so it means the same on any filesystem), `_DirReplaced` (swap before and after the directory is opened), `_DirCantBeChecked`, `_Kept` (hardlink twin, duplicates, a remove that finds the entry already gone), `_SymlinkInsideSkill`.
  - Hooks: `TestInstallOpencode_PluginsBehindSymlink`, `TestPruneOpencodeDir_BehindSymlink`, `TestCheckPluginRoots_UnresolvablePlugins`, `TestPluginsHome`.
  - `removeguard`: `TestBehindSymlink`, `TestBehindSymlink_Errors`, `TestIsSymlink`.
- **Suites:** `go test -race -cover ./...` passes, 1111 Go test passes. `uninstall.test.sh` 188 (128 on `main`). Every hook suite, `install.test.sh` (21) and `remote-install.test.sh` (24) pass.
- **Mutations.** The first push claimed 26/26; the review's own run found 3 survivors in 37 (G16 same-file twin kept, G18 already-gone remove kept, S6 check order). All three are now pinned by tests.
  - After this round: 45/45 killed on the final commit, all compiling (Go 30: 21 `removeStale`, 5 hooks, 4 `removeguard`; shell 15).
  - One (the hooks resolve-error branch) survived at first, and a seam test was added for it.

## Behaviour changes
- An `agents/`, `skills/`, `commands/` or `plugins/` directory that is a symlink **or behind one** keeps its stale devexp entries until they're removed by hand or the link becomes a real directory.
- A stale entry on disk only under another case or normalization is no longer removed; it is dropped from the manifest with a warning.
- `uninstall.sh` no longer removes through such a directory, never removes a symlinked entry, re-checks after the confirmation, and counts only real removals.

## Out of scope
- Install still writes through a symlinked *entry*, overwriting its link target. That is write-path safety, handed to #124.
  - For the merge: the removal-only helpers are `removeguard.BehindSymlink` / `removeguard.IsSymlink`, `removalDir` / `openRemovalDir` in `cli/cmd/backup.go`, and `pluginsHome` / `behindSymlink` in `cli/internal/hooks/opencode.go`.
  - This PR also touches `checkPluginRoots`.
- `uninstall.sh` removes by repo names rather than the manifest, and doesn't clean opencode `commands/`. Tracked in #123.
- `removeStale` pins directories with `os.Root`, but opencode `plugins/` removal still re-checks by path rather than through a pinned handle.


cli/cmd/backup.go
```diff
@@ -157,16 +169,34 @@ func removeStale(dir string, old, installed []string, shape staleShape, removeFn
 			ui.Warn(fmt.Sprintf("%q left untouched: the same file as %q, installed by this run", path, twin))
 			continue
 		}
+		if linked {
+			leftBehind = append(leftBehind, fmt.Sprintf("%q", path))
+			kept = append(kept, name)
```
alexandrocuma, line 223: **Important: an entry kept here can later remove a user's case-variant file.** Before this PR a stale name got one removal attempt, then was dropped. Now it stays recorded for as long as the link exists. `lstat(filepath.Join(dir, name))` resolves case and normalization variants on default APFS.

Repro with the real binary:
1. linked `agents/` + manifest `retired-agent.md` → kept.
2. The user removes it by hand as the warning says, writes their own `Retired-Agent.md`, and replaces the link with a real directory.
3. The next install prints `- "retired-agent.md"` and deletes `Retired-Agent.md`.

The run between steps 1 and 2 doesn't need to happen: while the link is still in place, the user's file passes `lstat` and is kept again.

Suggested fix (cheap, and it covers the error-kept path too): read the directory once and act only on an exact byte match.
```go
exact := map[string]bool{}
if ents, err := os.ReadDir(dir); err == nil {
	for _, e := range ents { exact[e.Name()] = true }
}
...
if !exact[onDisk(name, shape)] {
	ui.Warn(fmt.Sprintf("%q left untouched: on disk only under a different spelling", path))
	continue // dropped: what is there isn't the name devexp recorded
}
```
(If `ReadDir` fails, fall back to keeping the entry, as for an `lstat` error.) A test on a case-insensitive filesystem can skip itself when `os.Lstat(strings.ToUpper(p))` fails.

cli/cmd/backup.go
```diff
@@ -127,11 +129,20 @@ var lstat = os.Lstat
...
 //
-// An entry that is already gone needs nothing. Names and paths are printed
-// quoted, so nothing from the manifest reaches the terminal raw.
-func removeStale(dir string, old, installed []string, shape staleShape, removeFn func(path string) error, dryRun bool) {
+// Of the entries kept, those left behind a symlinked dir and those that
+// couldn't be checked or removed are returned; the others aren't devexp's on
+// disk. An entry that is already gone needs nothing. Names and paths are
+// printed quoted, so nothing from the manifest reaches the terminal raw.
+func removeStale(dir string, old, installed []string, shape staleShape, removeFn func(path string) error, dryRun bool) (kept []string) {
+	linked := isSymlink(dir)
```
alexandrocuma: **Minor, TOCTOU:** `linked` is computed once, before the loop. If `dir` is replaced with a symlink after this and before `removeFn(path)` (for example a dotfiles bootstrap running in another terminal), the remove goes through the link. I confirmed this with a scratch test that swaps the directory from the `lstat` seam. It isn't an attacker boundary, but it's cheap to close.

- **Minimal:** re-check `isSymlink(dir)` immediately before `removeFn`. This narrows the window but doesn't close it.
- **Closed (Go 1.25, which `go.mod` already requires):** pin the directory by fd.
```go
lfi, err := os.Lstat(dir)
linked := err == nil && lfi.Mode()&os.ModeSymlink != 0
r, rerr := os.OpenRoot(dir) // follows a link, hence the SameFile check
if rerr == nil {
	defer r.Close()
	if st, serr := r.Stat("."); serr != nil || !os.SameFile(st, lfi) {
		linked = true // swapped between Lstat and OpenRoot
	}
}
...
removeFn(r, onDisk(name, shape)) // r.Remove / r.RemoveAll
```
I prototyped this: after the swap, `r.Remove("old.md")` removed the original and left the dotfiles copy intact. `r.Lstat` can replace the path-based `lstat` too. The seam becomes `func(*os.Root, string) error`.

alexandrocuma (commented):
## PR Review: never remove stale agents or skills through a symlinked directory (#128)

**Recommendation: Request changes.** Everything below is small. The core fix is right, and I couldn't get install or uninstall to delete anything through a symlinked `agents/`, `skills/` or `commands/`. The findings are in or caused by this diff, so under the scope rule they should be fixed here.

**Summary.** `removeStale` now skips removal through a symlinked target directory, returns what it kept, and both installers record those names. `uninstall.sh` follows the same rule through `removable()`. There are three gaps. Keeping entries in the manifest indefinitely makes an old macOS case-folding hazard reachable. The directory check can go stale before the remove, in both the Go code and the shell script. And one uninstall message now contradicts the new warning.

### What I verified (scratch copy, scratch HOME)
- `go vet`, `gofmt -l`, and `go test -race ./...` all pass. `uninstall.test.sh`: 154 passed.
- Real binary, symlinked `~/.claude/agents`: nothing is removed through the link, one warning per run, and the stale name is recorded exactly once, including after two runs (no growth).
- Dangling symlinked `agents/`: the install fails at `mkdir … file exists` before `removeStale` runs, and the manifest is untouched.
- `os.RemoveAll`/`rm -rf` on a real skill directory unlinks symlinks inside it without following them (`TestRemoveStale_SymlinkInsideSkill`). Names with `..`, `/`, `\`, or control characters are still rejected before any path is built.
- `uninstall.sh` with `HOME` ending in `/`: the `dirname` grouping still matches and the entries are listed.
- **Mutations:** 37 run, 34 killed. Go: 21 run, 19 killed. Shell: 16 run, 15 killed. The survivors are below.

### Important
1. **A kept entry can later remove a user's file that differs only in case** (`backup.go:172-176`, inline). Repro on default APFS with the real binary:
   1. `agents/` links to dotfiles, and the manifest lists `retired-agent.md`. The run warns and keeps it.
   2. The user removes it by hand as told, then writes their own `Retired-Agent.md`. They replace the link with a real directory.
   3. The next install prints `- "retired-agent.md"` and deletes the user's `Retired-Agent.md`.

   Name-based ownership has always had this hazard, but only for a single run. This PR keeps the entry for as long as the link exists, or the remove keeps failing. A Unicode-normalization twin behaves the same way. **Fix:** remove (or keep) an entry only when that exact name is in `os.ReadDir(dir)`, byte for byte. Otherwise drop it with a warning. This is one `ReadDir` per call. The case of the user writing a file with the exact same name is inherent to name-based tracking, so I'd document it rather than fix it.

### Minor
2. **TOCTOU in the Go code: the directory is checked once, before the loop** (`backup.go:144`, inline). If `dir` is replaced with a link after `isSymlink(dir)` and before `removeFn(path)`, the remove goes through the link. I confirmed this with the `lstat` seam in a scratch test. For a local CLI this needs a concurrent dotfiles apply, not an attacker, so it's low. It's also cheap to close: Go 1.25 has `os.Root.RemoveAll` (see inline).
3. **TOCTOU in `uninstall.sh` across the confirmation prompt** (`uninstall.sh:109`, inline). `removable()` runs at collection time, and `rm` runs after `read`, which can wait indefinitely. Re-check in the remove loops.
4. **Contradictory skills message** (`uninstall.sh`, inline at the new listing block). When every skill is left behind a symlinked `skills/`, the output says `Skills will be kept (still in use by other installed CLI).` right next to `… is a symlink — … remove these by hand`, even with no other CLI installed. Reproduced. **Fix:** `elif ! $REMOVE_SKILLS; then` at line 173.
5. **Mutation survivors (tests don't pin these claims):**
   - **G16:** keeping (returning) a *same file on disk* twin survives. The docs say only link-blocked and error entries stay recorded, but no test covers a hardlink twin (inline on `backup_test.go:348`).
   - **S6:** swapping the order of the two `-L` checks in `removable()` survives. No test has a symlinked entry *inside* a symlinked directory, so which message and list it lands in is unpinned.
   - **G18:** dropping `&& !os.IsNotExist(err)` after `removeFn` survives. This is effectively equivalent (a remove race; the next run drops it), so no test is needed.
6. Nit, `backup.go:133`: `without following it.);` has stray punctuation.

### PR body
- "Mutations: 26/26 killed": see the survivors above.
- "'Removed N item(s)' counts only real removals": this is true for skipped links, but `(( removed++ ))` still counts an `rm` that failed (lines 229-249, pre-existing).

### Docs / CHANGELOG
They match the behavior. I checked the warning strings in the guide against the code in both the Go and shell paths, the dry-run line, the manifest-retention wording, and the parent-directory caveat. The `### Fixed` entry is in the right place under `[Unreleased]`.

### Follow-up candidates (pre-existing, not filed)
- **Symlinked `~/.claude`, `~/.config` or `~/.config/opencode`** still gets removals through into the dotfiles tree, which is #128's scenario one level up. The PR body says resolving parents misfires on `/var` → `/private/var` and on a linked `/home`. That only happens when comparing against an unresolved path. A cheap middle ground is `filepath.EvalSymlinks(dir) != filepath.Join(filepath.EvalSymlinks(HOME), ".claude", "agents")`: both sides resolve HOME, so system links cancel out. It should be applied to `plugins/` at the same time, for consistency.
- `uninstall.sh` counts a failed `rm` as removed.
- `manifest.Stale` doesn't dedupe. A duplicated manifest entry is listed and kept twice. It's stable across runs, not growing.
- Already tracked: #123 (uninstall by repo names, opencode `commands/`), #124 (writes through a symlinked entry).

### Done well
- The kept/dropped split is principled: only link-blocked and error entries are recorded, and anything proven not to be devexp's is dropped. Mutations G13, G14, G15 and G17 all confirm it.
- The link check sits after the shape and twin checks and before the dry-run branch, so a dry run never previews a removal through a link (G9 is killed).
- The integration test goes through the full replace-the-link-then-rerun cycle for both targets, and `treeBytes` diffs the whole dotfiles tree rather than spot-checking.

cli/cmd/backup.go
```diff
@@ -127,11 +129,20 @@ var lstat = os.Lstat
 //   - what is on disk can be checked and has the entry's shape: a regular
 //     file for staleFile and staleCommand, a real directory for staleDir. A
 //     symlink is never removed: it is the user's own setup, as for opencode
-//     plugin files.
+//     plugin files. (Inside a real skill directory, os.RemoveAll unlinks a
+//     symlink without following it.);
```
alexandrocuma: Nit: `without following it.);` has stray punctuation. Drop the `.)`, or close the parenthesis before the semicolon.

uninstall.sh
```diff
@@ -97,13 +97,22 @@ echo ""
...
+LEFT_THROUGH_LINK=()  # entries of a symlinked target directory
+LEFT_LINKS=()         # entries that are symlinks themselves
+
+# removable <dir> <name> <-f|-d>: whether <dir>/<name> is a file (-f) or
+# directory (-d) this script may remove. `devexp install` writes through a
+# symlinked target directory (a dotfiles setup) but never removes through one
+# (#128), and never removes an entry that is itself a symlink (#117): the
+# user's own setup. Such an entry is recorded to list instead. One line, like
+# every function here: a top-level `local` once aborted the script (#109).
+removable() { test "$3" "$1/$2" || return 1; if [[ -L "$1" ]]; then LEFT_THROUGH_LINK+=("$1/$2"); return 1; fi; if [[ -L "$1/$2" ]]; then LEFT_LINKS+=("$1/$2"); return 1; fi; }
```
alexandrocuma: **Minor, TOCTOU across the prompt:** this check runs at collection time, and `rm -f`/`rm -rf` run after `read -p "Proceed with removal?"`, which can wait indefinitely. Re-checking at removal is cheap. In each of the three remove loops:
```bash
if [[ -L "$(dirname "$f")" || -L "$f" ]]; then warn "$f changed since the preview (now behind a symlink) — left untouched"; continue; fi
```
(One line each, following the no-`local` rule.)

uninstall.sh
```diff
@@ -169,6 +175,26 @@ elif $REMOVE_CLAUDE || $REMOVE_OPENCODE; then
     echo ""
 fi
 
+# What is left in place is listed, once per symlinked directory.
```
alexandrocuma: **Minor, contradictory output.** When every skill is left behind a symlinked `skills/`, `SKILL_DIRS` is empty. The `elif` just above (line 173) then prints `Skills will be kept (still in use by other installed CLI).`, followed by this `… is a symlink — … remove these by hand` warning, even when no other CLI is installed. Reproduced with a Claude-only install and a linked `~/.claude/skills`.

Fix: make line 173 `elif ! $REMOVE_SKILLS; then`, so the "other CLI" message only appears when that is the reason.

cli/cmd/backup_test.go
```diff
@@ -0,0 +1,424 @@
...
+		root := t.TempDir()
+		dir := filepath.Join(root, "agents")
+		writeTestFile(t, filepath.Join(dir, "old.md"), "old\n")
+		writeTestFile(t, filepath.Join(root, "user.md"), "user data\n")
+		mustSymlink(t, filepath.Join(root, "user.md"), filepath.Join(dir, "linked.md"))
+		writeTestFile(t, filepath.Join(dir, "new.md"), "new\n")
+
+		var kept []string
+		out := captureStdout(t, func() {
+			kept = removeStale(dir, []string{"old.md", "gone.md", "linked.md", "../user.md", "NEW.md", "new.md"}, []string{"new.md"}, staleFile, os.Remove, false)
```
alexandrocuma: **Mutation survivor G16:** adding `kept = append(kept, name)` to the *same file as an installed one* branch (`sameFileAs`) passes every test. The docs say only link-blocked and error entries stay recorded. Suggest adding a hardlink twin here: `os.Link(filepath.Join(dir, "new.md"), filepath.Join(dir, "hard.md"))`, with `"hard.md"` in `old`. That pins that it's dropped and that `new.md` survives.

uninstall.test.sh
```diff
@@ -810,6 +810,75 @@ for home_case in unset empty relative; do
     check "HOME $home_case: the binary is not called" calls_are ""
 done
 
+# ── Symlinked target directories and entries (#128) ──────────────────────────
+# `devexp install` writes through a symlinked agents/ or skills/ (a dotfiles
+# setup) but never removes through one, and never removes a symlinked entry
+# (#117). uninstall.sh follows the same rule: what it leaves is listed, and a
+# real entry next to it is still removed.
+for linked in claude-agents claude-skills opencode-agents; do
```
alexandrocuma, line 818: **Mutation survivor S6:** swapping the two `-L` checks in `removable()` (entry before directory) passes. No case has a symlinked entry *inside* a symlinked directory, so the list and message it lands in are unpinned. Suggest adding one, e.g. `ln -s "$U/x.md" "$D/agents/some-agent.md"` in the claude-agents case, and asserting it's listed under `… is a symlink — devexp never removes files through it` and not as `left untouched`.

cli/cmd/backup.go
alexandrocuma: Fixed in 8146ac2. removeStale reads the directory once (`openRemovalDir`) and acts only on an exact listed name. Another spelling is dropped from the manifest with a warning: case via the listing, normalization via `rootLstat`. `TestRemoveStale_ExactName` simulates an insensitive filesystem through the `rootLstat` seam, so it means the same on any host. I reran your repro with real binaries: the previous build deleted `Retired-Agent.md`, this one leaves it and warns. The exact-same-name limit is documented in install.md.

cli/cmd/backup.go
alexandrocuma: Fixed in 8146ac2, as you suggested. `os.Lstat`, then `os.OpenRoot`, then `os.SameFile(root.Stat("."), lfi)`. The seam is now `func(*os.Root, string) error` and callers pass `(*os.Root).Remove` / `RemoveAll`; go.mod declares go 1.25.11, which has `Root.RemoveAll`. `TestRemoveStale_DirReplaced` covers a swap before OpenRoot (blocked, kept) and a swap after it (the original entry is removed, the dotfiles copy is untouched), for both files and skill dirs.

cli/cmd/backup.go
alexandrocuma: Fixed in 8146ac2.

uninstall.sh
alexandrocuma: Fixed in 64b3a7d. The three loops call a one-line `remove_entry`, which re-checks `[[ -L "$f" ]] || behind_link "$(dirname "$f")"` and skips with a warning. The new uninstall.test.sh cases swap the dir, the entry or a skill after the preview (through a FIFO answer) and assert nothing in the link target is removed.

uninstall.sh
alexandrocuma: Fixed in 64b3a7d (`elif ! $REMOVE_SKILLS`). The claude-skills linked case now asserts the message is absent.

cli/cmd/backup_test.go
alexandrocuma: Pinned in 8146ac2: `TestRemoveStale_Kept` has a hardlink `hard.md` → `new.md`, which is dropped with the same-file warning while `new.md` survives. G18 is pinned too, by a remove that finds the entry already gone: it is dropped with no warning, and the comment in removeStale explains why.

uninstall.test.sh
alexandrocuma: Pinned in 64b3a7d, but I chose the other order so it matches install. In Go, the shape check runs before the directory check, so a symlinked entry inside a symlinked dir is warned about as a symlink and dropped. `removable()` now checks the entry's `-L` first, and the claude-agents case asserts `… is a symlink — left untouched` and that the entry isn't in the remove-by-hand list. The swapped order is killed (T3).

alexandrocuma:
Review round addressed in 8146ac2 and 64b3a7d. The inline threads each have a reply. For the items outside the threads:
- **Parent symlinks:** new `removeguard.BehindSymlink` (`EvalSymlinks(dir)` vs `EvalSymlinks(HOME)` joined with the path under HOME). It's applied to `removeStale`, to `plugins/` (`checkPluginRoots`, `pruneOpencodeDir`) and, as `behind_link`, to `uninstall.sh`. Tests cover a linked `~/.claude`, a linked `~/.config/opencode`, and a HOME that is itself behind a link.
- **Failed `rm` counted:** only successful removals are counted now, and a failed `rm` warns.
- **Duplicate manifest entries:** kept names are recorded once.
- **PR body:** corrected. Mutations are now 45/45 on the final commit, and the three survivors from your run (G16, G18, S6) are pinned.

alexandrocuma (commented):
## Re-review of 8146ac2 and 64b3a7d: my 9 findings

**Recommendation: ready to merge.** All nine findings are fixed, and I found no regressions. What's left is non-blocking.

### Each finding, re-checked (scratch HOME, real binary built from 64b3a7d)
| # | Finding | Result |
|---|---|---|
| 1 | Kept entry removes a user's case-variant file | **Fixed.** I re-ran the repro with the real binary: `agents/` linked to dotfiles, a duplicated `retired-agent.md` in the manifest, then the user writes `Retired-Agent.md` and replaces the link. The output is `left untouched: on disk only as "Retired-Agent.md"`, the entry is dropped from the manifest, and the user's file survives two more runs. |
| 2 | TOCTOU in the Go code | **Fixed.** I ran scratch tests of both swap orders. Swapped before `OpenRoot`: blocked with "changed while devexp was checking it" and the entry is kept. Swapped after: the remove goes through the pinned root, and the dotfiles copy is intact. |
| 3 | TOCTOU across the uninstall prompt | **Fixed.** I swapped `~/.claude/agents` for a link while `read` was waiting. Both agents were reported as "changed since the preview" and left untouched, the dotfiles copy is intact, and the count is `Removed 1` (the real skill). |
| 4 | "Skills will be kept (other CLI)" message | **Fixed.** It is gone when the skills are kept because of a link. |
| 5 | Survivors G16, S6, G18 | **Fixed.** The analogous mutants N20, T11 and N21 are all killed now. |
| 6 | Stray punctuation | Gone. |
| 7 | "26/26" claim in the PR body | Updated. |
| 8 | Failed `rm` counted | **Fixed.** With a `chmod 555` agents dir the output is `could not remove …` and `Removed 0 item(s)`. |
| 9 | Linked `~/.claude` / `~/.config` not detected (follow-up) | **Done.** The real binary reports `is behind a symlink (it resolves to …)` and nothing is removed. `uninstall.sh` agrees. |

### Regression probes
- **`BehindSymlink` false positives.** None found for: a HOME under `/var` or via `/private/var`, a trailing `/`, a trailing `/.`, a lower-case spelling of HOME on APFS, or an on-disk `.Config` reached as `.config`. A relative HOME, a dir outside HOME and a missing dir all return errors. All of these fail closed, and callers reject them earlier anyway. `behind_link` gives the same answers for the same cases, including `HOME=/`.
- **`checkPluginRoots` as the write gate.** A fresh opencode plugin install with `~/.config` linked and no `plugins/` writes `devexp.js` through the link (scratch test). An unresolvable `plugins/` only sets `linked`, so it never blocks the install.
- **`os.Root`.** `Root.RemoveAll` unlinks symlinks inside a skill without following them (`TestRemoveStale_SymlinkInsideSkill`). Every supported OS has `openat`.
- **`ReadDir`.** It runs once per call, and only when there are candidates. An unreadable dir keeps every candidate with a warning (I checked `chmod 300` with the real binary). Names with newlines can't be candidates because `isInstalledName` rejects control characters.
- **Go/shell parity.** Link-or-behind decisions match. Error paths differ only in wording: Go says "can't be checked" and keeps; the shell lists the entry as behind a link. Neither removes anything.

### Mutations on the new code
37 valid mutants, 35 killed. Three initial mutants didn't compile, so I fixed and re-ran them; they're counted once.
- **Survived, T2:** `behind_link` without its `[[ -L ]]` short-circuit. This is equivalent, because a symlinked dir always resolves elsewhere. No action needed.
- **Survived, N23:** switching the Claude skills removal in `install_claude.go` to path-based `os.RemoveAll(filepath.Join(skillsTarget, n))` passes. The integration tests don't swap the directory mid-run, so the installer's use of `os.Root` isn't pinned. It's low-risk wiring, and a one-line test with a swapping `removeFn` would pin it. Optional.

### Nits (non-blocking)
- The `"<dir>" can't be checked … left untouched: …` warning lists manifest names (`"old-cmd"`), while the other warnings list paths (`old-cmd.md`).
- `rm -f` on an entry that vanished during the prompt returns 0, so it is counted as removed.

### Follow-ups (not in the new code; not filed)
- `uninstall.sh` still matches by repo name with `test -f`. On APFS, a user's `Dev-Agent.md` would be removed for the repo's `dev-agent.md`. This belongs with #123.
- opencode `plugins/` removal still re-checks by path rather than through a pinned handle. The PR body already acknowledges this.

PR #157, opened 2026-09-17
fix: atomic symlink-safe config writes; orphaned hook registrations and per-target enabled (#124, #150)

### LINKED ISSUE #124: Config and manifest writes: atomic and symlink-safe everywhere

### LINKED ISSUE #150: Hook registration: stale entries from other install roots, per-target enabled check, uninstall recursion
## Problem
Three gaps in Claude Code hook registration, found reviewing PR #147 (#137/#138). None was introduced by that PR.

1. **Registrations from another install root are never pruned.** Stale-hook pruning only recognises devexp commands under the current repo root. Suppose devexp was installed from another checkout or asset root, and a hook script was later removed from the registry. That registration is never pruned by `devexp install`, and `uninstall.sh` never removes it.
2. **`InstallClaude` checks `hook.Enabled` rather than `EnabledFor(TargetClaudeCode)`.** A hook enabled for opencode only could be treated as enabled for Claude Code.
3. **The `uninstall.sh` python step doesn't catch `RecursionError`.** Deeply nested JSON in `settings.json` raises it, and `set -e` then aborts the whole uninstall instead of skipping that step.

## Acceptance Criteria
- [ ] A devexp-form registration from another install root, whose script no longer exists, is pruned by install and removed by uninstall. User commands are never touched.
- [ ] Claude Code hook registration uses the per-target enabled check. A test covers a hook enabled for opencode only.
- [ ] Deeply nested `settings.json` makes uninstall skip the settings step with a message, leave the file untouched, and continue.
- [ ] Tests for each.

## Non-Goals
- Atomic writes (#124), dry-run previews (#148).

alexandrocuma:
Closes #124
Closes #150

## What changed

**#124: every config and manifest save is atomic and symlink-safe.** New helper `cli/internal/fsutil.WriteFileAtomic`:
- It follows a symlink, or a chain of links, to the final target and replaces that target. The links stay.
- It writes a temp file in the target's directory, fsyncs it, renames it over the target, then fsyncs the directory.
  - A **new** file's temp is created `O_EXCL` with the requested mode and never chmodded, so the umask applies (0600 under `umask 077`, as with `os.WriteFile`).
  - A **replacement** starts 0600 and takes the old file's permission bits.
- It refuses, leaving the file untouched: a dangling link or loop, a directory or other non-regular file, a read-only file, a directory with no room for a temp file, and an EXDEV/EBUSY rename.
- A replace makes a new inode, so it breaks hard links and drops xattrs, ACLs, setuid/setgid/sticky bits and ownership. The docs say so.

Where it's used:
- `settings.json` (`hooks.InstallClaude`), the manifests, the opencode plugin files and the legacy opencode config cleanup.
- The opencode `config.json` MCP merge:
  - It no longer overwrites a `config.json` it can't parse (JSONC, a top level or `mcp` that isn't an object, `null`).
  - Such a file is left untouched. **The MCP step is skipped** with a warning that names the file and the servers to add by hand; agents, skills and hooks still install, dry run included.
  - With `--mcps-only` it is an error.
  - opencode itself reads `config.json` as JSONC.
- Agents, opencode commands and skills are written atomically.
  - **An installed entry that is a symlink is never written.** That covers an agent, a command, a skill directory, and a file or directory inside a skill. It is kept with a warning, in dry runs too, and its name stays in the manifest.
  - A real run no longer reports a skipped `SKILL.md` as added.
- `uninstall.sh`:
  - Both python steps save under the same contract, umask included.
  - The MCP step cuts only the removed servers out of the file text, with a check against the parsed result.
  - A failing python step warns and the uninstall continues.

**#150: hook registration.**
- **Orphaned registrations:** install prunes, and uninstall removes, devexp-form registrations of scripts gone from **another** devexp root (a directory with `hooks/registry.json`).
  - User commands, exec-form (`args`) handlers, unclean paths and a leading `//` are never touched. Go and python agree on all of these.
  - A registration whose whole root was deleted can't be recognised and stays; this is documented.
- **Per-target enabled:** Claude Code registration uses `EnabledFor(claude_code)`.
- **Deep nesting:** a `settings.json` or `config.json` nested more than 500 levels makes the uninstall step skip with a message and leave the file untouched, instead of crashing the uninstall.
  - The limit is an explicit linear pre-scan (`too_deep`), so the result is the same on every python version. Before, it depended on when json raised `RecursionError`: near 1,000 levels on 3.9-3.11, not on 3.13+. CI's 3.12 caught this.
  - `RecursionError` is still caught as a backstop.
  - Go install's `encoding/json` limit is 10,000, which no python version reaches reliably. So a file nested 501-10,000 deep is edited by install but skipped, untouched, by uninstall. Skipping is safe.

**Merged with #156 (#128).**
- There is one `IsSymlink`, in `fsutil`. `removeguard.IsSymlink` was identical and is removed; `checkPluginRoots` and its warning use `fsutil.IsSymlink`, and the test moved with it.
- Both policies stand: nothing is removed through a symlink or a linked parent (#156), and no symlinked entry is written (#124).
- All CHANGELOG entries are kept.

## Review follow-up (#157 review)
- **Umask:** new files honour it. See `umask_unix_test.go` in `fsutil`, `mcp` and `manifest`, plus a python case. A real binary under `umask 077` produced 0600 settings.json, config.json, manifests and agents.
- **JSONC `config.json`:** only the MCP step is skipped (`TestDoInstallOpencode_UnmergeableConfig`).
- **Skill symlink preview, and no false "added" line:** `TestInstallClaude_SymlinkInsideRealSkillDir`.
- **Survivors pinned:**
  - atomic agent, skill and command writes (read-only directory tests);
  - the roots cache;
  - `IsAbs` (`t.Chdir`);
  - python `normpath` and `//`;
  - `O_EXCL` plus retry (`tempSuffix` seam);
  - a replacement's temp staying private.
- **Docs:** what a rename-based save doesn't keep, and the window between the Lstat checks and the write.

## Verification
- **Go:** `go test -race -cover ./...` passes, gofmt and vet are clean. Coverage: fsutil 91.5%, hooks 92.8%, skills 89.1%, agents 89.0%, cmd 82.1%.
- **Shell and JS suites, all 0 failed:** `uninstall.test.sh` 250 (after merging #156), and it passes with python 3.9.6, 3.10, 3.11, 3.12.3 (ubuntu:24.04, as CI), 3.13 and 3.14. The containers skip only the two Go round trips; `install.test.sh` 21, `remote-install.test.sh` 24, every Claude Code hook suite, every opencode JS suite.
- **`fsutil` failure tests:**
  - failure injection: short write, fsync EIO, rename error, EXDEV, EBUSY;
  - a child process SIGKILLed right before the rename;
  - refusal cases;
  - umask cases 077, 022, 027 and 0.
- **Real binary, scratch HOME:**
  - A symlinked settings.json keeps its link and 0600 mode.
  - Under `umask 077` every new file is 0600.
  - A JSONC `config.json` gives a warning, the full install still runs, and the file keeps its bytes.
- **Mutations:** 94/96 killed.
  - The depth limit scored 8/8: off-by-one, strings, escapes, closing brackets, each pre-check, and the limit raised to 10,000.
  - The re-review survivors are all killed: python `O_EXCL`, retry and bound, and the private replacement temp; Go's retry bound and constant suffix.
  - The unused `ConfigRefusedError.Unwrap` was removed.
  - The original set scored 53/55. The review-fix set scored 27/27: umask, `O_EXCL` and retry, MCP refusal paths, skill preview, atomic entry writes, roots cache, `IsAbs`, python `//`, normpath and umask, and the plugins/ warning after the merge.
  - The 2 survivors can't fail by construction: the MCP splice verification, and `RecursionError` in the settings compare (parsing fails first).

## Behaviour changes
- A replaced file keeps its own permission bits; a new file gets 0644 minus the umask.
- A dangling symlinked `settings.json`, `config.json` or manifest is refused. Before, install created the file the link pointed at.
- A read-only plugin file is refused rather than silently replaced.
- An unmergeable opencode `config.json` skips the MCP step with a warning. It is an error under `--mcps-only`.
- A symlinked agent, command, skill or file inside a skill is never written through.
- `claude_code.enabled` is honoured.
- The uninstall MCP step no longer reformats `config.json`.

## Notes
- #124's AC2 ("never written through") was superseded by the later issue comment. The installer replaces a symlinked config's final target and keeps the link. The existing *removal* paths still refuse symlinks.

- Install's MCP merge still re-encodes all of `config.json`. A comment-preserving edit (hujson patch, or the uninstall member splice) is a follow-up.
- The wizard's `listHookNames` and `resolveHookDisabled` still use top-level `Enabled`.
- `remote-install.sh` moves the binary non-atomically across devices.
- The reviewer's pre-existing items are being filed separately: the MCP entry reported as "updated" on every run, `opencode.json` taking precedence over `config.json`, backup and `ExtractFile` writes, and the uninstall exit status after a warning.


cli/internal/fsutil/atomic.go
```diff
@@ -0,0 +1,160 @@
...
+	syncDir(dir)
+	return nil
+}
+
+// resolveTarget returns the file WriteFileAtomic replaces for path, and the
+// permission bits the new file gets.
+func resolveTarget(path string, perm os.FileMode) (string, os.FileMode, error) {
+	li, err := os.Lstat(path)
+	if os.IsNotExist(err) {
+		return path, perm, nil
```
alexandrocuma: **Important: a new file ignores the umask.**

A missing path returns `perm`, and `f.Chmod(mode)` (line 86) applies it exactly, because chmod ignores the umask. `os.WriteFile`, which this replaces, masked it.

Under `umask 077`, real binaries with a scratch HOME:
- `main` creates `settings.json`, `config.json`, `.devexp-manifest.json` and the agent files 0600.
- This branch creates them 0644.

opencode `config.json` can carry MCP `env` and `headers` resolved from `.env`, so those become world-readable.

Suggested fix: for a new file, create the temp with `os.OpenFile(tmpName, os.O_WRONLY|os.O_CREATE|os.O_EXCL, perm)` (random suffix, retry on `EEXIST`), so the kernel applies the umask. `Chmod` only when replacing an existing file. Pin it with a unix-only test under `syscall.Umask(0o077)`.

alexandrocuma (commented):
## PR Review: atomic, symlink-safe config writes; orphaned hook registrations; per-target enabled (#124, #150)

**Recommendation: Request changes**

**Summary.** `fsutil.WriteFileAtomic` is solid where it counts. I found no path where a partial file can replace the original. Symlink chains, relative links, links across directories and symlinked parent directories all behave as documented. The uninstall MCP splice also held up under my fuzzing. Two in-diff problems block the merge:

1. **New files now ignore the umask.** Under `umask 077` they are created 0644 instead of 0600. `config.json` can hold MCP secrets, so this is a security regression.
2. **An opencode `config.json` with comments now stops the whole opencode install**, dry run included. opencode itself reads `config.json` as JSONC.

The other in-diff items are minor, plus some test gaps found by mutation.

Branch checked: `297980f`. HOME always pointed at scratch dirs; no real `~/.claude` or `~/.config` was touched.

---

### 🟡 Important

**1. New files ignore the umask and are world-readable (`cli/internal/fsutil/atomic.go:86`, `:108`).**
- `resolveTarget` returns `perm` for a missing file. `os.CreateTemp` creates the temp file 0600, then `f.Chmod(mode)` sets the mode, and chmod ignores the umask. `os.WriteFile`, which this replaces, applied it.
- Repro: real binaries, scratch HOME, `umask 077`, `devexp install --reinstall-mcps` for each target.
  - `main`: `settings.json`, `config.json`, `.devexp-manifest.json` and agent files are all `-rw-------`.
  - This branch: the same new files are `-rw-r--r--`.
- `~/.config/opencode/config.json` gets `env` and `headers` resolved from `devexp.config.json` and `.env`, so tokens become readable by other users on a shared machine.
- `TestWriteFileAtomic_NewFile` uses 0640 under the default umask 022, so it can't see this.
- **Fix:** for a new file, create the temp with `os.OpenFile(name, O_WRONLY|O_CREATE|O_EXCL, perm)` in a retry loop with a random suffix, so the kernel applies the umask. Call `Chmod` only when replacing an existing file. Add a test under `syscall.Umask(0o077)`, for example in a `//go:build unix` file like `repo/umask_unix_test.go`. The docs then need "`perm` minus the umask for a new file" instead of "0644 for a new file".

**2. An opencode `config.json` with comments stops the whole opencode install (`cli/internal/mcp/opencode.go:26-29`).**
- opencode reads the global `config.json` through its JSONC parser, as it does `opencode.json` and `opencode.jsonc`: `loadGlobal` → `loadFile` → `ConfigParse.jsonc`, in `packages/opencode/src/config/config.ts`. The docs say "OpenCode supports both JSON and JSONC", so comments and trailing commas are valid there.
- `doInstallOpencode` returns on this error before agents, skills and hooks. The refusal also comes before any check for whether anything would change, so it fires when every MCP server is already configured, and in `--dry-run`.
- Repro: scratch HOME, `config.json` = `{ // my model\n "model": "x", }`, then `devexp install --dry-run` with only `opencode` on PATH. Result: `opencode install: mcp: … is not valid JSON … fix it and re-run`, and nothing else is previewed or installed.
- Refusing is far better than `main`, which replaced such a file with only the MCP servers. But turning "lost my comments" into "devexp won't install at all" for a valid opencode config is too blunt.
- **Fix, now:** treat an unmergeable `config.json` like the other refusals in this PR. Warn, name the file, list the MCP servers to add by hand, skip only the MCP step and continue. With `--mcps-only`, returning the error is fine. **Follow-up:** a comment-preserving edit, for example `tailscale/hujson` `Patch`, or the same member-splice approach as the uninstall step.

### 🔵 Minor (in the diff, so fix now)

**3. A symlink inside a real skill directory isn't shown in a dry run, and the real run reports it as added (`cli/internal/skills/installer.go:37-50`).**
- `CopyDir` doesn't run in a dry run. So with `~/.claude/skills/beta/SKILL.md` → dotfiles, the dry run prints `write …/beta/` and no warning.
- The real run warns, then prints `+ beta/SKILL.md` for the file it just skipped. I reproduced both with a probe test.
- The CHANGELOG says a symlinked "file or directory inside a skill" is left untouched "with a warning naming it, in a dry run too".
- **Fix:** in a dry run, walk the destination for symlinked entries (the same `IsSymlink` walk, no writes). Have `CopyDir` report whether it skipped `SKILL.md`, so `ui.Added` isn't printed for it. Or narrow the CHANGELOG claim.

**4. Test gaps: my mutants that survived** (see Mutation results below).
- Swapping `fsutil.WriteFileAtomic` back to `os.WriteFile` in `agents/installer.go:93` or `skills/installer.go:84` passes every test. The CHANGELOG says agent, command and skill files are now written atomically, but no test pins it. A `TestWrites_NeverInPlace`-style check for those files would.
- `roots[root] = true` instead of `isRoot` (`hooks/installer.go:412`) survives. No test has two missing scripts under the same non-devexp root in one pass, where the second is wrongly pruned when the cache is wrong.
- Removing the `normpath` check from `uninstall.sh`'s `is_orphaned` survives. Add an unclean path such as `<root>/hooks/../hooks/claude-code/gone.sh` to the orphan look-alike cases.
- Dropping `filepath.IsAbs(p)` from `isOrphanedDevexpHook` survives. The `"relative"` case runs with cwd = the package directory, where `plain/` doesn't exist, so it passes for the wrong reason. `t.Chdir(base)` makes it real.

**5. Docs: what a rename-based save doesn't keep (`docs/guides/install.md`, "How devexp saves files").** Only ownership is listed. The rename also:
- breaks a hard link, leaving the other name with the old contents (checked);
- drops extended attributes and ACLs;
- drops setuid, setgid and sticky bits (`& 0o777` / `.Perm()`).

These are acceptable for these files, but worth one line each.

**6. Nit: Go/python parity on a leading `//` (`uninstall.sh:588` vs `hooks/installer.go:394`).** `os.path.normpath('//r/hooks/claude-code/x.sh')` keeps the `//` (POSIX), but `filepath.Clean` folds it to `/`. So uninstall removes a registration that install keeps. It's harmless because the registry marker is still required. Checking `p.startswith('//')` would line them up.

**7. Nit: the symlinked-entry check can race.** `keepSymlinkedEntry` / `IsSymlink` runs an `Lstat`, then `WriteFileAtomic` follows links. A link created in between is written through. The window is tiny and needs a same-user writer, so it's fine to leave. A `noFollow` flag on the helper would close most of it.

### ✅ What's done well
- `resolveTarget` refuses dangling links, loops, non-regular files and read-only files before creating anything. The temp file is `O_EXCL` via `CreateTemp`, in the **target's** directory, so the rename never crosses directories behind a symlink. Every error path removes the temp file.
- Failure injection (short write, `EIO` fsync, `EXDEV`/`EBUSY` rename) and the SIGKILL-before-rename subprocess test are good evidence.
- The MCP splice is careful work. 3,000 random configs (CRLF and tabs, escaped keys such as `"m\u0063p"`, duplicate keys at both levels, `NaN`, `1e400`, big integers, lone-surrogate escapes, nested `"mcp"` keys) gave 0 invalid or wrong outputs. A comment, a BOM or invalid UTF-8 is skipped with the file untouched.
- Recognising another root by its `hooks/registry.json` marker, instead of a list of names, is the right call. Relative paths, `~`/`$VAR`, wrappers and unmounted volumes (marker unreadable) are all kept.
- `claude_code.enabled` now matches opencode's `EnabledFor`.

### Mutation results (my own run, scratch copy)
- **Go:** 40 mutants that compile, 34 killed, 6 survived.
  - 2 survivors can't be tested or are equivalent: removing the `syncDir` call, and removing `closed = true` (the double `Close` on success is harmless).
  - The other 4 are real gaps (item 4): the agents write swap, the `CopyDir` write swap, the roots cache, and `IsAbs`.
  - Pinned: symlink resolve, dangling link, W_OK, mode kept, temp cleanup, file fsync, regular-file check, each installer's symlinked-entry check and its manifest name, `EnabledFor`, every `isOrphanedDevexpHook` and `isDevexpRoot` guard, the trailing-data, `mcp`-object, `null` and `UseNumber` checks, the blank file, and the atomic writes for `settings.json`, `config.json`, the manifest, plugin files and the legacy config.
- **Shell (`uninstall.test.sh`):** 21 mutants, 18 killed, 3 survived.
  - **The `os.path.normpath(p) != p` check in `is_orphaned` survives.** No shell case feeds an unclean orphan-shaped path, though the Go table has one.
  - The splice guard `same = True` survives, as your PR body says.
  - `found[-1]` → `found[0]` survives, but it is equivalent: the loop removes every duplicate either way.

### Overlap with #156 (#128)
- **Duplicate helper.** `removeguard.IsSymlink` and `fsutil.IsSymlink` are byte-identical in behaviour. Whichever merges second should keep one: `fsutil` is the natural home, and `removeguard` can import it.
- **Symlink policies agree.** Both write through a symlinked target *directory*. #156 never removes a symlinked entry, and this PR never writes one. Both keep such names in the manifest.
- **Expect textual conflicts** in `uninstall.sh` (#156's hunks end at ~258, this PR's start at 259), `uninstall.test.sh`, the imports in `hooks/opencode.go`, `CHANGELOG.md`, `docs/guides/install.md` and `docs/architecture/overview.md`.

### Follow-up candidates (pre-existing, outside this diff; listed, not filed)
- `mcp.InstallOpencode` reports `context7 — updated` and rewrites `config.json` on **every** run, on `main` too. It compares `json.Marshal(map)` (sorted keys) with `json.Marshal(ocEntry)` (struct field order), so they never match.
- devexp writes the legacy `~/.config/opencode/config.json`. opencode now seeds and prefers `opencode.jsonc` / `opencode.json`, and merges them **after** `config.json`, so an `mcp` entry there overrides devexp's.
- `backup.go:30` still uses `os.WriteFile` with its errors ignored, and `assets.ExtractFile` is an unused plain write.
- `uninstall.sh` prints "Uninstall complete." and exits 0 after any `|| warn` fallback. That matches the existing plugin step, but a caller can't tell anything went wrong.
- (Already in the PR body) the wizard's `listHookNames` / `resolveHookDisabled` use top-level `Enabled`, and `remote-install.sh` moves the binary non-atomically.

cli/internal/mcp/opencode.go
```diff
@@ -19,9 +23,9 @@ type ocEntry struct {
 }
 
 func InstallOpencode(mcps []MCP, env map[string]string, configPath string, dryRun, reinstall bool) error {
-	config := map[string]interface{}{}
-	if data, err := os.ReadFile(configPath); err == nil {
-		json.Unmarshal(data, &config) //nolint:errcheck
+	config, err := loadOpencodeConfig(configPath)
+	if err != nil {
+		return fmt.Errorf("mcp: %w — it was left untouched and no MCP servers were added; fix it and re-run", err)
```
alexandrocuma: **Important: this stops the whole opencode install for a valid opencode config.**

opencode reads `~/.config/opencode/config.json` with its JSONC parser (`loadGlobal` → `loadFile` → `ConfigParse.jsonc` in `packages/opencode/src/config/config.ts`), so comments and trailing commas are legal there. `doInstallOpencode` returns on this error before agents, skills and hooks.

The check also runs before anything is compared, so it fires:
- when every MCP server is already configured;
- in `--dry-run`. Repro: `{ // c\n "model": "x", }` → the dry run prints only this error.

It's much better than `main`'s silent overwrite. But please warn, list the servers to add by hand, and skip just the MCP step, still returning the error for `--mcps-only`. A comment-preserving edit can be a follow-up.

cli/internal/skills/installer.go
```diff
@@ -33,6 +34,11 @@ func InstallClaude(srcDir, targetDir string, disabled []string, dryRun bool) ([]
 			continue
 		}
 		destDir := filepath.Join(targetDir, name)
+		if fsutil.IsSymlink(destDir) {
+			warnSymlinked(destDir)
+			installed = append(installed, name)
+			continue
+		}
 		if dryRun {
```
alexandrocuma, line 43: **Minor: a symlink inside a real skill directory isn't previewed.**

In a dry run `CopyDir` doesn't run, so `beta/SKILL.md` → dotfiles prints only `write …/beta/`, with no warning. The real run warns, then line 50 prints `+ beta/SKILL.md` for the file it skipped.

The CHANGELOG says entries inside a skill are warned "in a dry run too". Walk the destination for symlinks in the dry-run branch, and don't report `SKILL.md` as added when `CopyDir` skipped it.

cli/internal/agents/installer.go
```diff
@@ -85,7 +90,7 @@ func InstallClaude(srcDir, targetDir, model string, disabled []string, dryRun bo
 		if err := os.MkdirAll(targetDir, 0755); err != nil {
 			return installed, err
 		}
-		if err := os.WriteFile(dest, content, 0644); err != nil {
+		if err := fsutil.WriteFileAtomic(dest, content, 0644); err != nil {
```
alexandrocuma, line 93: **Test gap (mutation survivor).** Replacing this with `os.WriteFile(dest, content, 0644)` passes every test, and so does the same swap in `skills.CopyDir` (`skills/installer.go:84`). The CHANGELOG says agent, command and skill files are now written atomically. Something like `TestWrites_NeverInPlace` (check the inode changes, or hook the `rename` seam) would pin it.

cli/internal/hooks/installer.go
```diff
@@ -363,6 +378,73 @@ func isStaleDevexpHook(cmd, repoDir string) bool {
...
+		return false
+	}
+	if _, err := os.Stat(p); !os.IsNotExist(err) {
+		return false
+	}
+	root = filepath.FromSlash(root)
+	isRoot, seen := roots[root]
+	if !seen {
+		isRoot = isDevexpRoot(root)
+		roots[root] = isRoot
```
alexandrocuma, line 412: **Test gap (mutation survivor).** `roots[root] = true` passes every test. If the cache were wrong, a second missing script under the same non-devexp root would be pruned in the same pass. Add a user root with two missing `hooks/claude-code/*.sh` registrations to `TestInstallClaude_PrunesOrphanedHooksFromOtherRoots`.

cli/internal/hooks/installer_test.go
```diff
@@ -1655,3 +1655,255 @@ func TestRequoteDevexpHooks_DropsEmptiedEntry(t *testing.T) {
...
+		"missing script in a root with no registry":                        {cmd: filepath.Join(userDir, "hooks", "claude-code", "removed.sh")},
+		"missing script in a root that is gone":                            {cmd: filepath.Join(base, "deleted", "hooks", "claude-code", "removed.sh")},
+		"missing script under repoDir (not orphaned: isStaleDevexpHook's)": {cmd: filepath.Join(repoDir, "hooks", "claude-code", "removed.sh")},
+		"with an argument":                                                 {cmd: plainGone + " --flag"},
+		"double-quoted":                                                    {cmd: `"` + plainGone + `"`},
+		"through a wrapper":                                                {cmd: "bash " + plainGone},
+		"chained":                                                          {cmd: plainGone + ";true"},
+		"through a variable":                                               {cmd: "$HOME/plain/hooks/claude-code/removed.sh"},
+		"relative":                                                         {cmd: "plain/hooks/claude-code/removed.sh"},
```
alexandrocuma, line 1746: **Passes for the wrong reason.** The test's cwd is the package directory, where `plain/` doesn't exist, so `isDevexpRoot("plain")` is false whatever the `filepath.IsAbs(p)` guard does. Dropping that guard survives. `t.Chdir(base)` before this table makes the case real.

uninstall.sh
```diff
@@ -383,6 +560,51 @@ def command_path(cmd):
...
+
+def is_orphaned(p):
+    # Whether p, the path a devexp-form command runs, is a script gone from
+    # <root>/hooks/claude-code/ of a devexp install root: this repo, or another
+    # checkout or asset cache devexp was installed from, whose registry has
+    # since dropped the hook (#150). The name can't be checked against a
+    # registry, so the root is. A user's hook directory has no registry; a root
+    # deleted outright can't be told from a user's and is left alone. Same rules
+    # as isStaleDevexpHook / isOrphanedDevexpHook in cli/internal/hooks.
+    if not p.startswith('/') or os.path.normpath(p) != p:
```
alexandrocuma: **Nit and test gap.**

1. Removing `os.path.normpath(p) != p` survives `uninstall.test.sh`: no shell case uses an unclean orphan-shaped path, though the Go table does.
2. Parity: `normpath('//r/hooks/claude-code/x.sh')` keeps the leading `//` (POSIX), but Go's `filepath.Clean` folds it. So uninstall removes a registration that install keeps. It's harmless, since the registry marker is still needed, but `or p.startswith('//')` would match Go.

docs/guides/install.md
```diff
@@ -129,8 +129,28 @@ devexp owns only the hook handlers it registers: `{"type": "command", "command":
...
+  - a path, or link target, that is a directory or another non-regular file;
+  - a file you can't write (a rename would still replace a read-only file in a writable directory);
+  - a directory where no temp file can be created;
+  - a rename the file system refuses, such as a bind-mounted file (`EXDEV`/`EBUSY`).
+- **Where the symlink rule differs.** Some paths keep their earlier, stricter rules:
+  - The steps that *remove* devexp's entries from opencode's `config.json` leave a symlinked `config.json` untouched, with a warning to edit it by hand. That covers the legacy plugin entry and `uninstall.sh`'s MCP step.
+  - `devexp uninstall` does the same for a symlinked manifest.
+  - An installed agent, command or skill entry that is itself a symlink is never written (see [What gets overwritten vs. preserved](#what-gets-overwritten-vs-preserved)).
+  - Plugin files must not be symlinks at all (see [Stale-file cleanup](#stale-file-cleanup)).
+- Ownership isn't copied: the saved file belongs to the user running devexp.
```
alexandrocuma: **Minor (docs).** Besides ownership, the rename also:
- breaks a hard link (the other name keeps the old contents; checked);
- drops xattrs and ACLs;
- drops setuid, setgid and sticky bits.

One line each would do. Once the umask fix lands, line 139's "0644 for a new file" also needs to say "minus the umask".

cli/internal/fsutil/atomic.go
alexandrocuma: Fixed in c288feb. A new file's temp is now opened `O_EXCL` with `perm` and a random suffix and never chmodded, so the umask applies. A replacement starts 0600 and takes the old file's bits. `uninstall.sh`'s `write_atomic` does the same. Tests run under `syscall.Umask` in `fsutil`, `mcp` and `manifest` (unix only, serial), plus a python case. A real binary with `umask 077` and a scratch HOME gave 0600 for settings.json, config.json, both manifests and the agents. The docs now say perm minus the umask.

cli/internal/mcp/opencode.go
alexandrocuma: Fixed in c06767a. `InstallOpencode` returns `mcp.ConfigRefusedError`. `doInstallOpencode` warns with the file, the reason and the servers to add by hand, skips only the MCP step, and still installs agents, skills and hooks, dry run included. `--mcps-only` still returns the error, and any other MCP error still stops the install. `TestDoInstallOpencode_UnmergeableConfig` covers all three, and your JSONC repro passes on the real binary. A comment-preserving edit is left as a follow-up.

cli/internal/skills/installer.go
alexandrocuma: Fixed in 1ec2280. `copyDir(dryRun)` walks the destination for symlinks without writing, and the preview warns. A real run prints `+ <skill>/SKILL.md` only when it wrote SKILL.md. Covered by `TestInstallClaude_SymlinkInsideRealSkillDir` (dry and real). The CHANGELOG's 'in a dry run too' is now true.

cli/internal/agents/installer.go
alexandrocuma: Pinned in 1ec2280. `TestInstall_AgentFilesNeverWrittenInPlace` covers all three agent installers, and `TestInstall_SkillFilesNeverWrittenInPlace` covers `CopyDir` and commands. An existing file in a read-only directory must be refused with its bytes kept, where `os.WriteFile` would write in place. All four swaps are now killed.

cli/internal/hooks/installer.go
alexandrocuma: Pinned in b95f329. `TestPruneStaleHooks_RootsCachedPerAnswer` has two missing scripts under a user root and two under a devexp root, in both orders. `roots[root] = true` is now killed.

cli/internal/hooks/installer_test.go
alexandrocuma: Fixed in b95f329: the table runs under `t.Chdir(base)`, and I added a `./plain/...` case. Dropping `IsAbs` is now killed.

uninstall.sh
alexandrocuma: Both done in b95f329. `is_orphaned` now refuses a leading `//`, as `filepath.Clean` folding it does in Go. The shell look-alikes include `<root>/hooks/../hooks/claude-code/gone.sh` and `//<root>/...`, and I added a `//` case to the Go table. Removing the normpath check or the `//` check is killed.

docs/guides/install.md
alexandrocuma: Done in c288feb. 'How devexp saves files' now lists what a replace doesn't keep: ownership, hard links (the other name keeps the old contents), xattrs/ACLs and setuid/setgid/sticky. The `WriteFileAtomic` doc comment says the same and notes the window between the Lstat checks and the write (also commented at both entry checks).

alexandrocuma (commented):
## Re-review at `668819e`

**Recommendation: one small test-only change, then merge.** The production code is right. The only in-diff finding left is a test gap on the python side of the umask fix.

### My earlier findings 1–7: all fixed
- **1. Umask.** Checked with the real binary under `umask 077`: all 111 files a Claude Code and opencode install creates are `-rw-------` (`settings.json`, `config.json`, both manifests, agents, skills, commands, plugin files). Under `umask 022` they are 0644. A replaced 0600 `config.json` stays 0600, and no temp file is left behind.
  - A new file's temp is `O_EXCL` with `perm` and is never chmodded. A replacement's temp is created 0600 and chmodded to the old file's bits.
  - A symlink already sitting at the temp name, pointing at a real file or at nothing, is never followed: the next name is tried and the other file is untouched.
  - Retries stop after 101 tries with an error. On a failed write or rename the temp file is removed.
  - Python `write_atomic` behaves the same way; I checked it with a seeded `os.urandom`.
- **2. Commented `config.json`.** A full install and a dry run warn, name the file and list the servers to add by hand, then install everything else; `config.json` is byte-identical afterwards. `--mcps-only` still exits 1.
- **3. Symlink inside a skill.** A dry run now warns about it. A real run warns and no longer prints `+ deliver/SKILL.md`, and the dotfiles copy is untouched.
- **4. Test gaps.** Each of my earlier survivors is now killed: the agents and skills write swaps, the roots cache, `IsAbs` (via `t.Chdir`) and the python `normpath` check.
- **5–7.** The docs, the `//` parity and the race comments are all in place and accurate.

### The merge with #156
- **`IsSymlink`.** `removeguard.IsSymlink` had one caller, `warnLeftBehind`, which now uses `fsutil.IsSymlink`. The two are identical, and `TestIsSymlink` moved over with the function.
- **Merge diff.** Against `main`, `backup.go`, `install_claude.go` and `removeguard` show no merge-introduced changes (only the `IsSymlink` move).
- **`checkPluginRoots`.** The body is unchanged from `main`:
  - a behind-symlink or unresolvable `plugins/` only sets `linked`;
  - writes still go through `checkPluginDest` + `fsutil.WriteFileAtomic`;
  - `pruneOpencodeDir` refuses on `linked`.
- **`uninstall.sh`.** Against `main`, every #124 hunk starts at line 298 or later, and #156's `behind_link`/`remove_entry` section is unchanged. A combined run with a linked `~/.claude` behaved correctly:
  - agents behind the link are listed and kept;
  - opencode agents are removed;
  - the MCP member is spliced out with CRLF kept;
  - `settings.json` hooks are removed through the link, which stays a link, with 0600 kept.
- **#156 repros on this branch, all correct:**
  - With a linked `~/.claude`, the stale agent is kept with the "behind a symlink" warning and stays in the manifest.
  - A case twin `Retired-Agent.md` is kept.
  - When `agents/` is swapped for a link at the uninstall prompt, both entries are "changed since the preview — left untouched" and the dotfiles copies survive.

### Mutations on the new code
- **Go:** 23 mutants, all compiling. 20 were killed. Survivors:
  - `try < 100` → `try < 1`: the test only needs one retry. Nit.
  - A constant temp suffix: no test has a stale temp at the next name. Nit.
  - `Unwrap` returning nil: nothing uses it. Nit.
- **Shell (`uninstall.test.sh`):** 7 mutants, 4 killed. The 3 survivors are python `write_atomic` gaps, commented inline:
  - **Q2:** dropping `O_EXCL` on the new-file temp (it would then follow a symlink at that name);
  - **Q6:** a replacement temp created 0644 instead of through `mkstemp` (0600);
  - **Q7:** not retrying on `FileExistsError`.

  None of these is a live bug: the code is right, and both uninstall steps only ever replace existing files. But the "keep the copies identical" contract means the python copy should pin what the Go tests pin (`TestWriteFileAtomic_TempNameTaken`, `TestWriteFileAtomic_ReplacementTempStartsPrivate`).

### Suites run at `668819e` (a clean `git archive` copy, scratch HOME)
- Go: `gofmt` and `go vet` are clean, and `go test -race ./...` passes.
- Shell and JS: `uninstall.test.sh` 239/239, `install.test.sh` 21/21, `remote-install.test.sh` 24/24; every Claude Code and opencode hook suite passes.

uninstall.sh
```diff
@@ -298,64 +298,193 @@ if $REMOVE_OPENCODE && [[ -f "$REPO_DIR/mcps/registry.json" ]]; then
...
+    directory = os.path.dirname(target) or '.'
+    tmp = None
+    try:
+        if mode is None:
+            # A new file: created with 0644 and never chmodded, so the umask
+            # applies (0600 under umask 077), as with a plain open().
+            while True:
+                tmp = os.path.join(directory, '.' + os.path.basename(target) + '.tmp-' + os.urandom(8).hex())
+                try:
+                    fd = os.open(tmp, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o644)
```
alexandrocuma, line 352: **Test gap (mutation survivors Q2 and Q7).**
- Replacing `os.O_EXCL` with `os.O_TRUNC` here passes all 239 tests. So does re-raising `FileExistsError` instead of retrying.
- A python twin of `TestWriteFileAtomic_TempNameTaken` would pin both. Seed `os.urandom` so the first name is taken by a symlink to a file outside, then check that the outside file is untouched and the target is written. The awk-extracted `write_atomic` harness this suite already uses would do.

The code itself is right: I ran exactly that probe and it passes.

uninstall.sh
```diff
@@ -298,64 +298,193 @@ if $REMOVE_OPENCODE && [[ -f "$REPO_DIR/mcps/registry.json" ]]; then
...
+                    break
+                except FileExistsError:
+                    tmp = None
+        else:
+            fd, tmp = tempfile.mkstemp(prefix='.' + os.path.basename(target) + '.tmp-', dir=directory)
```
alexandrocuma, line 359: **Test gap (mutation survivor Q6).** Swapping `mkstemp` for `os.open(..., O_CREAT|O_EXCL, 0o644)` passes every test, so nothing pins that a replacement's temp is private until it is chmodded. That is the counterpart of `TestWriteFileAtomic_ReplacementTempStartsPrivate`.

To pin it, wrap `os.fsync` in the harness to record `os.fstat(fd).st_mode & 0o777` under `umask 022`, and expect `0o600`.

PR #158, opened 2026-09-17
fix(hooks): secret-in-write-guard matches real token shapes, allows placeholders; detects sk-service- keys (#143, #152)

### LINKED ISSUE #143: secret-in-write-guard: key-shaped placeholders and quoted PEM headers are blocked

### LINKED ISSUE #152: secret-in-write-guard: detect legacy OpenAI sk-service- keys
## Problem
`secret-in-write-guard` doesn't detect OpenAI's older service-account key form (the `sk-service-…` prefix). Current service-account keys (`sk-svcacct-…`) are detected since #101. Public secret scanners (Trivy, TruffleHog) still detect the older form, because keys issued in it may still be live.

## Acceptance Criteria
- [ ] Confirm the format's shape and minimum body length from public sources, and cite them in the pattern comment.
- [ ] Both twins block it, with the length pinned at the edge (one short is allowed, exact length blocks).
- [ ] Look-alike identifiers starting with `sk-service-` are allowed.
- [ ] sh/js parity kept; tests mirrored.

## Non-Goals
- Other vendors' formats.

alexandrocuma:
Closes #143
Closes #152

## What changed
`secret-in-write-guard`, Claude Code hook and opencode module together (same patterns, same order):

- **GitHub tokens matched by their real shape (#143).**
  - The classic pattern accepted `_` in the body, so long snake_case names containing a prefix like `ght_` were blocked. Classic tokens now match only their documented alphanumeric body.
  - GitHub App installation tokens (`ghs_`, including Actions' `GITHUB_TOKEN`) have been moving to a stateless format since 2026-04-27, and that format does contain `_`. They get their own pattern, so they still block. This corrects the #143 follow-up comment, which proposed "no `_`" everywhere and would have missed them.
  - That pattern expects the compact JWT header GitHub documents. One rare kind of name, where a prefix is followed by a segment that starts the same way, still blocks; a test pins it.
- **Key-shaped placeholders allowed (#143):** documented example key IDs, `your-…` phrases, and a body of one repeated character.
  - The exemption applies only when the whole matched value is the placeholder, and every later position is still searched.
  - A real key next to a placeholder, or joined to one, still blocks.
  - Values wrapped in `<…>` never matched on `main` either; they are now tested, not newly allowed.
- **A private-key header blocks only when key material follows it closely (#143).**
  - Only text near the header, before an `END` or `BEGIN` line, counts. A header quoted in docs or code, or a template with an elided body, is allowed whatever comes later in the file.
  - Key-like text on the header's line or just after it (a long identifier, a fingerprint) still blocks. This is documented.
  - Real PEM blocks still block, including escaped, concatenated, encrypted and PGP-armored ones, ones with a dash rule under the header, and escaped bodies (escaped slashes, escaped newlines joining lines).
  - Other backslashes near a quoted header (Windows paths, escaped code strings, `\u` escapes, escaped prose, LaTeX) don't count as key material.
  - A body wrapped far narrower than PEM's usual line width, or written in pieces across several edits, isn't recognized. This is documented in `docs/reference/hooks.md` and both hook headers.
- **Older OpenAI `sk-service-` keys detected (#152).** The shape and length follow Trivy's `openai-service-api-key` rule and TruffleHog's OpenAI detector, which the pattern comment cites. Look-alike `sk-service-account-…`/`sk-service-worker-…` names are allowed.
- **Linear time (review).** Earlier commits in this PR added repetitions that could rescan the rest of a write from every retry. A timed-out Claude Code hook lets the write through, and the opencode handler runs synchronously. Every repetition that can reach past the next start is now bounded. Decisions are unchanged except for rare edge shapes beyond the new bounds, such as key material that starts far from its header (documented).

## Verification
- **Suites:** `secret-in-write-guard.test.sh` 159 → 321, `.test.js` 141 → 301. Each now has timing tiers at 100 KB and 2 MB.
- **CI suites:** after merging `main`, every hook suite, every installer script test, gofmt, vet and `go test -race ./...` pass under bash.
- **Timing, previous head of this PR:** on inputs built to force backtracking, it took 2–13 s (sh) and 0.5–6.5 s (js) at 100 KB on the quadratic cases. The cubic private-key cases took 20 s (sh) and 6 s (js) at 4 KB, and at 100 KB they ran past a 60 s cap (one js case finished in 19 s). The timing tiers fail on that head.
- **Timing, now:** the same inputs, plus escaped-line shapes, take ≤0.12 s at 100 KB and ≤1.2 s at 2 MB for the shell hook, including start-up. The module handler alone takes ≤6 ms at 100 KB and ≤160 ms at 2 MB. Time grows linearly with size. The earlier "2 MB in 0.2–0.5 s" claim was wrong: it wasn't measured on these inputs.
- **Against `main`:** run on `main`'s hooks, the new test files fail only on new cases. 42 new allow cases are blocked there, and 16 new block cases (17 in the shell suite) are allowed there. No pre-existing case differs.
- **Parity:** 11,454 generated inputs, built in memory. Both implementations made the same decision on every input.
- **Independent corpus:** the review's corpus has 1,203 inputs, with 0 parity differences. What `main` blocks and this branch allows is limited to three known classes (narrow-wrapped PEM bodies, bodies split across edits, stateless GitHub tokens with a non-compact JWT header) and rare shapes beyond the new bounds.
- **Mutation testing:** 117 of 118 mutants killed in each implementation. The set covers every bound, the search window and stop, the escaped-body rules, and the survivors both reviews found. The one survivor makes a length bound on an escaped line unbounded, which doesn't change any decision. The earlier "82/82" held only for my own mutant set.
- **Fixtures:** every secret-shaped fixture is assembled at runtime from split prefixes and FAKE bodies.

- `gh[postaur]_` keeps two letters GitHub doesn't document; kept as the conservative choice.
- GitHub has announced a format change for user-to-server tokens; a non-JWT form would need a new pattern.
- No placeholder rule for `github_pat_` or the legacy `sk-` pattern.
- The guard has no overall scan budget, and some key formats aren't matched (tracked separately).


alexandrocuma (commented):
## PR Review: secret-in-write-guard matches real token shapes, allows placeholders, detects sk-service- keys (#143, #152)

**Recommendation**: Request changes

**Summary**: The direction is right and the two twins stay in lockstep. I ran an independent corpus of about 1,170 inputs through the real hook (Claude Code `Write`/`MultiEdit` envelope on stdin) and the real opencode module, on this branch and on `main`. The placeholder exemptions hold up: I found no realistic real Anthropic, OpenAI, AWS, GitHub, Slack or PEM key that they let through. One problem is caused by this diff and needs fixing before merge. Several of the new patterns backtrack super-linearly, and a slow guard fails open. There are also two smaller PEM issues and some test gaps. Every secret-shaped input was built in memory from split prefixes and FAKE bodies.

---

### 🟡 Important

**1. The new patterns backtrack super-linearly, and a stalled guard fails open**
Files: `hooks/claude-code/secret-in-write-guard.sh` lines 45, 53, 60, 79, 83, 89 · `hooks/opencode/secret-in-write-guard.js` lines 31, 39, 46, 65, 69, 75

**Issue**: Several added constructs use unbounded, overlapping quantifiers:
- the private-key pattern: the header's word runs, then an unbounded lazy search for material;
- the `your-…` phrase lookaheads (Anthropic, OpenAI project keys, Slack);
- the `sk-service-` name run before the 48-character run;
- the stateless GitHub id segments.

On `main` the same inputs take milliseconds. On this branch:
- **Private-key pattern:** growth is roughly cubic. An input of a few kilobytes built around the header stalls the sh twin for minutes and the js twin for over a minute.
- **The other patterns:** growth is quadratic. About 100 KB takes several seconds, and about 1 MB takes minutes.

This doesn't match the PR's "2 MB adversarial inputs take 0.2–0.5 s".

Why it matters for a guard:
- **Claude Code:** a `PreToolUse` command hook that times out doesn't block the tool call; the call goes ahead. The default timeout is 600 s, and the registry sets none. A write the guard can't finish scanning is let through, secret included.
- **opencode:** the match runs synchronously, so it freezes the event loop for as long as it takes.

The threat model is accidental leaks, so this is unlikely to trigger by chance. It's still a fail-open path, this diff added it, and it's cheap to close.

**Suggested fix**: Bound the repetition in each new construct:
- the header word runs;
- a material search window of a few hundred characters, which is far more than encrypted-PEM or PGP armor headers need;
- the number of words in a `your-…` phrase;
- the length of the `sk-service-` name;
- the length and number of stateless id segments.

I tried exactly this on a scratch copy of both twins:
- both suites pass (252 and 232);
- the sh and js twins still agree on every input in my corpus;
- no real key that blocks on this branch is allowed;
- every adversarial shape I tried runs in ≤0.5 s (sh) and ≤0.25 s (js) at 2 MB.

Please also add a perf regression test per construct, so the bound can't quietly come back.

---

### 🔵 Minor (inside this diff, so please fix here)

**2. Whether a quoted private-key header is allowed depends on the rest of the write** (sh:89, js:75)
The material search runs until the next `-----` or the end of the content. A header quoted in docs or code is re-blocked by any later run of base64-class characters anywhere in the write, such as a long camelCase identifier, a filesystem path, a hex commit SHA or an SSH fingerprint. In a real source file or runbook that's common, so #143's "header quoted in code is allowed" holds only for short snippets. The bounded window from finding 1 fixes the "anywhere later" part.

**3. The "next PEM boundary line" gap** (sh:89, js:75)
Material after any `-----` sequence isn't looked at. No standard private-key format puts a dash run between the header and the body, so an accidental leak is unlikely to hit this. It's still a free fix: stop the search only at a following `-----END` or `-----BEGIN`, not at any `-----`. With that change, the header-followed-by-certificate allow case still passes, and a dash rule between header and body blocks again.

**4. Test gaps that mutation found** (in each twin, on top of the PR's 82):
- **`sk-service-` lookahead boundary:** narrowing the placeholder boundary so it no longer counts `_`/`-` survives both suites. That mutant would let a real service key through when a placeholder-shaped run comes before it. Add a block case for that shape, as the Anthropic, OpenAI project-key and Slack patterns already have.
- **Anthropic `admin` key-type segment:** dropping `admin` from the optional key-type segment survives. No placeholder test uses that segment.
- **Stateless id segments:** narrowing the id segments to digits survives. Either pin the letters on purpose with a test, or narrow the class to match GitHub's documented `ghs_APPID_JWT`.

**5. Docs/PR text**
- **`<…>` values:** `docs/reference/hooks.md` and the PR body list `<…>` values as a new allowance. `<` is not a key character, so those values never matched on `main` either. Consider describing them as "never matched" instead.
- **Public wording:** it lists the intentional allowance classes without giving a way to get a real key through. That's fine.
- **PR body claims:** the performance and "82/82" claims should be updated after finding 1.

---

### Notes (not blocking; judgement calls)

- **Short-wrapped PEM bodies:** a PEM body re-wrapped far below any standard line width (64/70/76) is no longer matched; `main` matched it. No standard encoder produces that, so it's acceptable under the accidental-leak model. One sentence in "What `secret-in-write-guard` doesn't see" would make the limitation explicit.
- **Stateless GitHub pattern depends on a compact JWT header:** the pattern relies on the JWT starting with `eyJ`. That matches GitHub's `ghs_APPID_JWT` announcement. `main`'s broader class also caught variants whose header isn't compact JSON; this branch doesn't. That's low risk, but worth keeping in mind when GitHub publishes the user-to-server format.
- **New narrow false positive:** the stateless pattern also matches identifiers where a `gh?_` fragment is followed by a segment that starts with `eyJ`. It's rare.

### Pre-existing (follow-up candidates, not caused by this PR)

- The guard has no scan budget. Any slow pattern turns into fail-open under Claude Code's timeout semantics. Consider capping or chunking the scan, with a fail-closed timeout inside the hook.
- OpenAI service keys in the looser shape some scanners accept, with separators near the watermark, aren't matched. The new pattern follows Trivy's stricter rule.
- A PEM block that is URL-encoded as a whole isn't matched.
- PuTTY private key files (no `BEGIN` header) aren't matched.
- A temporary AWS key ID directly followed by more capitals isn't matched. This is a documented boundary choice.
- `github_pat_` and legacy `sk-` still have no placeholder rule (already listed in the PR).

### ✅ What's done well

- **Placeholder exemptions:** they are anchored to the whole matched value and to where the key's characters end, and every later start is still searched. I couldn't make a realistic real key pass through them, including keys joined to placeholders, wrapped in `<…>`, or preceded or followed by example text.
- **Stateless installation tokens:** the #143 follow-up would have missed them, and giving them their own pattern is the right correction.
- **Parity:** across my 1,174 inputs, sh and js made the same decision every time, on the branch and on the candidate fix.
- **Fixtures:** they are assembled at runtime from split prefixes and FAKE bodies, and threshold edges are pinned for most constructs.

alexandrocuma:
Thanks for the review. Every finding is addressed in 0a4ff8a, 75864ad and 57e5cf4 (merged with `main` in 0d691d5):

1. **Super-linear backtracking.** Every repetition that can reach past the next start is now bounded in both twins: phrase words, service-key name segments, GitHub id segments, the private-key header words and the material search. Both suites gain timing tiers at 100 KB and 2 MB, built at runtime, that fail on the previous head. I've corrected the timings in the PR body; the earlier claim wasn't measured on inputs like yours.
2. **PEM allowance depended on the rest of the write.** Only text near the header counts now, so the realistic later-in-file cases are allowed and tested. Key-like text on the header's own line or just after it still blocks, as documented.
3. **Boundary gap.** The search stops only at an `END` or `BEGIN` line. Block cases cover a dash rule under the header for every key type. JSON with escaped slashes also blocks again.
4. **Test gaps.** Added the `sk-service-` boundary, `admin` key-type and alphabetic id-segment cases. The rare name the stateless pattern still blocks is pinned as a known case; I kept requiring the compact JWT header that GitHub documents. The full mutant set is killed in both twins.
5. **Docs.** `<…>` values are described as never matched and now tested. Narrow-wrapped or split bodies are listed with the guard's other limits in `docs/reference/hooks.md` and both hook headers.

alexandrocuma (commented):
## Re-review at 0d691d5

**Recommendation**: Approve with minor comments

**Summary**: The blocking finding from the first review is fixed. I re-ran every earlier adversarial shape, plus about twenty new ones aimed at the new bounds, through the real sh hook and the real js module. Everything now runs in linear time. Every remaining difference from `main` is either one of the documented classes or a much rarer case of the same kind. The comments below are small and don't block the merge.

### Performance: fixed

- **Earlier shapes:** every shape from the first review, including the cubic private-key ones, takes ≤0.44 s (sh, hook start-up included) and ≤0.27 s (js) at 2 MB.
- **New shapes:** they target the bounds directly:
  - several key-type words inside the header bound, followed by a full window of near-length runs;
  - backslash runs mixed with near-miss `END` lines;
  - service-key names at the segment count and length caps;
  - stateless GitHub ids with long segments that contain the next prefix;
  - `your-…` phrases with very long words;
  - several shapes mixed into one write.
- **Result:** all of them are linear. Time grows about 20× from 100 KB to 2 MB, and the worst case stays under the numbers above.
- **Python backtracking:** none of the bounded quantifiers inside alternations backtracks super-linearly in Python's `re`. Each bounded group can only be split one way, and the private-key window stops at the next header.

### False negatives from the bounds

- **Placeholder phrase bound:** it only narrows an exemption. A phrase longer than the bound blocks, so it can't let a real key through.
- **Service-key name segments:** a segment longer than the per-segment bound still blocks, because the 48-character run matches inside it. Only a name with more segments than the cap is missed. That's far beyond any real service-account name, and `main` never matched these keys at all.
- **Stateless GitHub ids:** a token with more id segments than the cap is allowed, and `main` blocks it. GitHub documents one app-id segment, so this is theoretical.
- **Private-key window (please document):** a PGP or PEM body that starts further from its header than the window is allowed, and `main` blocks it. Standard armor and encryption headers are far shorter; I checked OpenPGP.js-style, GnuPG with several comments, and legacy encrypted PEM with AES and DES IVs, and all of them block. The case needs an unusually long run of armor comments. `docs/reference/hooks.md` doesn't mention it yet; one clause in "doesn't see" would cover it.
- **The three remaining classes:** on my corpus (1,203 inputs, with the new shapes), everything `main` blocks and this branch allows is in the three documented classes: narrow-wrapped LF bodies, bodies split across edits, and non-compact JWT headers. The only exceptions are the two rare boundary cases above, the private-key window and the stateless id-segment count.

### 🔵 Minor

**1. Backslash as key material brings back some header false positives** (sh:102, js:88)
- I tried the private-key pattern without the `\x5c` in the material class. Every false positive in this list disappears, and they are all quoted-header cases #143 set out to allow:
  - a Windows path near a quoted header;
  - a code string with `\n` escapes between short words;
  - a JSON-escaped doc with short lines;
  - `\uXXXX` escapes;
  - LaTeX.
- The backslash only rescues narrow-wrapped escaped bodies. Standard-width escaped bodies (including `\/` from PHP-style encoders) already block without it. Narrow-wrapped LF bodies are already an accepted miss, so the rescue buys little.
- **Options:**
  - count only an escaped slash or escaped newline (`\/`, `\n`, `\r`) as material, not any backslash; or
  - keep it, and add "escaped strings and Windows paths near the header" to the documented "still blocks" examples.

**2. Wording**
- The CHANGELOG ("with the same decisions for real keys") and the PR body ("the decisions for real keys are unchanged") are true except for the two boundary cases above. The "private-key window" item covers the documentation.

### Mutation testing

I ran 42 of my own mutants per twin against the new bounds, alongside the earlier ones:
- the window, stop lines and backslash class;
- the header, phrase, service-name and GitHub id bounds, each at ±1 and unbounded.

Two no-op controls ran under the same load and survived, so the timing tiers don't cause false kills.

34 of the 38 mutants per twin are killed. Of the 4 survivors, 3 are equivalent and don't need a test:
- **Stateless GitHub id segment length bounded:** an over-long numeric id is still blocked by the classic-token pattern.
- **Service-key name segment length made unbounded:** a longer segment already blocks, because the 48-character run matches inside it. Every decision stays the same, and the time stays linear because the segment count is capped.
- **Private-key window made greedy instead of lazy:** whether a match exists doesn't depend on greediness.

One survivor is a real test gap:
- **Service-key name segments required to be non-empty.** The branch blocks a key whose name part has an empty segment (a doubled or leading separator, or no name before the separator), but no test pins it. If a later change requires segments to be non-empty, such keys would pass unnoticed. **Please add one block case in each suite.**

### Parity

Across the 1,203 inputs, sh and js made the same decision every time, on this head.

alexandrocuma:
Thanks for the re-review. This round (76668d3, 1032739, 8483e45):

1. **Backslashes:** only escapes that occur inside escaped key bodies count as key material now. An escaped slash is part of a line, and an escaped newline joins two lines long enough to be key material. Other backslashes near a quoted header (Windows paths, escaped code strings, `\u` escapes, escaped prose, LaTeX) no longer block, and each class has an allow case. Escaped bodies at standard width and narrow-wrapped still block.
2. **Docs:** the "doesn't see" paragraph and both hook headers now say key material that starts far from its header isn't recognized.
3. **Wording:** the CHANGELOG and PR body now say decisions are unchanged except for rare edge shapes beyond the new bounds.
4. **Mutation gap:** added a block case for a service-key name with an empty segment in both suites. Re-running the full mutant set turned up one more gap, for an escaped slash in a single body line, which now has a test. The one remaining survivor doesn't change any decision.

Parity, the timing tiers, every suite and CI all pass.

PR #160, opened 2026-09-17
fix(hooks): dangerous-cmd-guard ends targets at shell expansions and runs in linear time (#151, #146)

### LINKED ISSUE #146: opencode dangerous-cmd-guard: regex matching can be slow on crafted long lines

### LINKED ISSUE #151: dangerous-cmd-guard: target boundary next to shell expansions
## Problem
`dangerous-cmd-guard` decides where a command's target ends from literal characters: whitespace, end of line, quotes, parens, backticks and shell operators. A shell *expansion* right next to a target isn't treated as the end of the target. Nor are some spellings of the home directory beyond the ones the rules name. So a delete of a protected directory written that way isn't matched. Both the Claude Code hook and the opencode module behave the same, on `main` and in PR #149.

Found in the PR #149 security review. Concrete cases are kept out of this public issue.

## Acceptance Criteria
- [ ] A protected-directory delete whose target is directly followed by a shell expansion is blocked by both implementations.
- [ ] The equivalent home-directory spellings are recognised wherever `$HOME`/`~` are.
- [ ] Benign paths that merely start with a protected prefix stay allowed (allow cases).
- [ ] Mirrored block and allow cases in `dangerous-cmd-guard.test.sh` and `.test.js`; sh/js parity kept.

## Non-Goals
- Encoded or indirect payloads, and commands saved to a file and run later. These are documented limits of a command-line matcher.

alexandrocuma:
Closes #151
Closes #146

## What changed
`dangerous-cmd-guard`, in both the Claude Code hook and the opencode module:

**#151: a target ends at a shell expansion (security).**
- A protected-directory target used to end only at whitespace, a quote, a closing paren or a list operator. So an expansion or glob placed right after the target wasn't matched, and neither was a home directory spelled another way, even though the shell still resolves both to the directory.
- In both implementations, a target now ends where an expansion, glob (including the single-character wildcard) or redirection starts.
- Every rule that recognises the home directory accepts its other bash and zsh spellings, including braced subscripts.
- `:` is a plain character, except as a zsh modifier after an unbraced home parameter.
- Paths with a literal component right after the protected prefix stay allowed.

**Which `rm` counts (review fixes).**
- `rm` must be a word of its own. It can't be the end of a longer word or option, such as a container `--rm` flag.
  - It may still follow a positional parameter, or be an argument of another command.
  - It may be followed directly by an expansion, backtick, brace or glob character, or a redirect.
- The wildcard-delete rule needs the target in the same simple command. Its scan stops at a real `;`, `|` or non-redirect `&`. Once a quote, escape, backtick, `$(`, `${`, `$[` or a process substitution opens before that separator, it runs on to the next `|`, as on `main`, so a separator that is data never ends it.
- This removes false positives the wider target end had made more common, and a few `main` already had. Real invocations still block.

**#146: linear time on long commands.**
- **opencode rules:** the JS rules took seconds to minutes on crafted long lines because of regex backtracking. Each rule now decides in linear time, with the same decisions as the grep patterns.
- **Masking (both hooks):** the inert-text masking re-checked every later pipeline stage for each stage, which was quadratic. This also hit the Claude Code hook: a long pipeline took over a minute on `main`. It's now computed once per pipeline.
- **Nesting:** nesting deeper than a shared limit (100) is scanned whole in both implementations, so Python's recursion limit no longer makes the two decide differently.

## Verification
- **Suites:** `dangerous-cmd-guard.test.sh` 169 → 380, `.test.js` 170 → 434.
  - The JS suite includes crafted-input timing tiers: 100 KB within 500 ms and 1 MB within 2 s.
  - The sh suite includes a 400 KB pipeline within 15 s, which times out on `main`.
- **After merging `main`:** every hook suite, the installer script tests and `go test -race ./...` pass.
- **Against `main`'s hooks:** the new block cases fail, and so do the timing checks. The only allow cases that fail are false positives `main` had.
- **Against earlier heads:** 30 block cases fail on `ad40f41` per suite, covering every re-review regression class. On `64e9680`, the 11 process-substitution and `$[` block cases fail, plus 1 allow case.
- **Timings on crafted lines (BSD grep), 73 inputs:**
  - 1 MB: JS at most 247 ms, Claude Code hook at most 2.6 s.
  - 4 MB: JS at most 1.02 s, Claude Code hook at most 9.9 s.
  - Both scale linearly.
- **Parity, 0 differences:**
  - Real sh hook vs JS module: the reviewer's two 4,000-command focused corpora and 186 probes (BSD and GNU grep), plus 12,000 + 6,000 ASCII fuzz commands (BSD + GNU grep).
  - Masking, Python vs JS byte for byte: 100,000 inputs.
  - Branch JS vs a backtracking reference built from the grep patterns: 450,000 inputs.
  - The only sh/JS differences are on non-ASCII whitespace, which is pre-existing and also present on `main`.
- **Differential vs `main` (JS), 3 × 150,000 generated commands:** every command `main` blocks and the branch allows is found by reverting one intended change at a time. The classes are `rm` inside a longer word, a real unquoted separator, or a literal `:`. 0 are unexplained.
- **Mutations:** JS 117/123 killed, sh 93/94 killed. This covers both reviews' lists and mine, ported to the current code. After the last test additions, the three non-equivalent survivors were re-run and killed. The rest are equivalent:
  - JS stop sets on rules whose suffix ignores the segment end;
  - an equivalent `git clean` class;
  - a CRLF split;
  - the find-cache start check;
  - the opener loop bound, where the char at the end is always a separator;
  - sh's redundant quote after `.claude`.

## Behaviour change
- A protected-directory delete whose target is directly followed by any expansion now blocks, even when a literal follows it or the variable isn't empty at runtime. Put a literal path component first.
- Quoting before a real separator makes the wildcard rule scan on to the next `|`, so it can over-block, e.g. a quoted argument, then `&&`, then a copy into a variable path under `/tmp`.
- Both are documented in `docs/reference/hooks.md`.

## Out of scope (tracked in #161)
Other spellings of the destructive commands and their targets, separators inside quotes, and non-ASCII whitespace handling.


alexandrocuma (commented):
## PR Review: dangerous-cmd-guard ends targets at shell expansions and runs in linear time (#151, #146)

**Recommendation**: Request changes

**Summary**: The #146 half is solid. I built my own checks for it: the rewritten opencode rules make the same decisions as `main` and as a backtracking copy of the new grep patterns, masking output is byte-identical, both implementations fall back at the same nesting depth, and both now run in linear time. The #151 half is the right idea, but three things in its own diff need fixing first: two home/glob forms in the class it sets out to cover still get through, and one of the new end characters blocks a very common allowed command. The tests also leave several of the new claims unpinned.

Concrete inputs are kept out of this public review, which describes each problem only by its kind.

---

### 🔴 Critical

None.

---

### 🟡 Important (in this diff, fix before merge)

**1. A glob that starts with the single-character wildcard isn't treated as an expansion.**
`hooks/claude-code/dangerous-cmd-guard.sh` line 541 (`END`), `hooks/opencode/dangerous-cmd-guard.js` line 27 (`END`)
**Issue**: END lists `*` and `[`, but the single-character wildcard only counts when an extglob paren follows it. So a delete of the root, the home directory, `/tmp` or `~/.claude` isn't matched when a glob that starts with that wildcard comes right after the target, even though the glob matches the directory's entire contents. This is the class the PR sets out to close ("a glob that matches the directory"), and `main` misses it too. It affects all three target rules in both implementations.
**Suggested fix**: Add the bare wildcard to END's character class. The paren-form alternative then only needs `@( +( !(`. Add block cases for each target in both suites. I checked this change in a scratch copy: both suites pass, and sh and JS still agree.

**2. `${HOME…}` with a subscript inside the braces isn't recognised as the home directory.**
`dangerous-cmd-guard.sh` line 544 (`HOME_RE`), `dangerous-cmd-guard.js` line 30 (`HOME`)
**Issue**: The docs say `${HOME}` is recognised "with any operator, flags or modifier". But the operator set after `HOME` has no subscript opener, so the bash array-style subscripts and zsh range subscripts don't match. They are not in the tests or in #161. The unbraced form with a subscript is matched.
**Suggested fix**: Add `[` to the character set that can follow `HOME` inside the braces, in both implementations. Add block cases for the bash and zsh forms, with and without `/.claude` after them.

**3. `:` in END causes false positives that `main` didn't have.**
`dangerous-cmd-guard.sh` lines 541, 558, 577; `dangerous-cmd-guard.js` lines 27, 137
**Issue**: A `:` only changes an expansion directly after an *unbraced* parameter name (zsh `$HOME:h`). After a literal `/`, `~`, `/tmp` or `.claude`, zsh and bash both keep it as a plain character (checked with `print`). The wildcard rule still uses the loose `rm\b[^|]*` prefix, which also matches an `rm` inside a longer word or option and continues across `;` and `&&`. Together they block a very common command form: a container run that removes itself on exit while mounting `/tmp` or `~/.claude` at a container path. Both implementations block it now; `main` allowed it.
**Suggested fix**: Take `:` out of END. Accept it only directly after the unbraced `$HOME` spellings, as an extra ending alternative in the root/home rule. Don't put `(:…)*` inside `HOME`, because that would backtrack in the JS rules. I checked this in a scratch copy: both suites pass (the `$HOME:h` case still blocks), the false positives go away, and a sh/JS parity fuzz found no new differences. Also consider anchoring `rm\b` to a command position (not after `-` or a letter). Otherwise the new `$`/`{` ends block things like a `/tmp/$x:` mount or a copy into `/tmp/$(…)` later in a line that ran `rm` earlier. That prefix predates this PR, but this PR widens its effect.

**4. Tests don't pin several of the PR's claims (mutation survivors).**
I mutated a scratch copy of each implementation and ran its own suite against it. Mutations that change decisions but still pass:
- **JS `sequence` checks only the first segment.** (Security-relevant.) If a rule's prefix appears in an earlier pipeline stage and the real invocation comes in a later one, the rule stops matching, and every suite still passes. Add block cases where the command runs again after a pipe, for the wildcard rule and the force-push rule.
- **The nesting limit can change in either implementation alone.** Moving it to 99 or 101 in the Claude Code hook, or to 101 in the opencode module, still passes. Both suites test only 50 and 150 levels, so "the same limit in both" isn't pinned. Add cases at 99 and 100 levels to both suites. I checked by hand that both implementations agree at 99, 100 and 101 levels for `(…)`, `$(…)`, `"$(…)"`, `<(…)` and backticks.
- **JS `claudeGlob` stops skipping `.claude` occurrences before the segment.** A `.claude…/*` in an earlier stage, before the delete, then blocks; nothing tests that.
- **Rules that change the separator set still pass:** the unanchored-wildcard rule stopping at `&`, `git reset --hard` stopping at `|`, and the fork-bomb rule stopping at `;`. "Decisions are unchanged" relies on these.
- **Spellings in the new END and HOME that no test uses:** `$^HOME` and `${^HOME}`, `${HOME` with the `#` or `/` operators, a `~name` containing a dot, and the `@(` and `!(` extglobs. Dropping any of them passes both suites (sh and JS).
- **The Claude Code hook's #146 fix isn't pinned.** Putting back the per-stage rescan of the rest of the pipeline gives identical decisions and still passes `dangerous-cmd-guard.test.sh`, because only the opencode suite has timing tests. Add one coarse timing case to the sh suite: a few hundred KB of pipeline with a generous budget, which `main` needs minutes for. The opencode module's cached `find` isn't pinned either: making it re-search every time still passes.
- **A force flag followed by `=` has no test in the sh suite.** Dropping that alternative still passes.

---

### 🔵 Minor (in this diff)

- **`docs/reference/hooks.md` line 175** says "A letter, digit, `/`, `.`, `-` or `_` continues the target". Every character not in END continues it, so the list is incomplete. Reword it once #1 and #3 settle END.
- **`docs/reference/hooks.md` line 179** says "at most about 1.5 s in the Claude Code hook" for a crafted 1 MB line. On my machine (BSD grep), a 1 MB single-line pipeline took about 2.3 s for the whole hook, about 1.9 s of it in the Python step. It does grow linearly (0.5 MB 1.4 s, 1 MB 1.9 s for the Python step; `main` took over 2 minutes at 0.5 MB). Say "a few seconds" or re-measure the worst input.
- **The trade-off.** A protected prefix followed directly by *any* expansion now blocks, even when a literal follows the expansion (a log file named by a variable plus an extension, a brace list of named files). That matches how a quote already ended a target, and the fix is documented, so it's acceptable. The narrower rule would be: block only when what follows the prefix is made of nothing but expansions and quotes until the word ends. Consider it later if agents run into it often.

---

### ✅ What's done well

- **Rewritten JS rules:** 0 decision differences across 400,000 generated commands (rm flag runs, CR/VT/FF/NBSP/U+2028, multi-line and backslash-continued text, repeated prefixes, `.claude` runs, Unicode). I compared against both `main`'s module (no command `main` blocks is allowed) and a backtracking reference built from the branch's grep patterns.
- **Masking:** `main` and branch `maskInert` are byte-identical over the same corpus. Python and JS masking agree at the nesting limit for every nesting kind I tried.
- **Linear time holds** on my own crafted inputs, not just the suite's (brace-flag runs, `$~` runs, `~name` runs, many lines, 99-deep nesting repeated, heredoc runs): JS stays under about 320 ms per MB and the sh hook under about 2.3 s per MB.
- **Fail-closed paths are intact:** the Python-error and grep-error exits still block, and `fail-closed.test.sh` passes.
- The per-pipeline `downstream_ok` and the explicit depth limit are clean, minimal changes. The comments explain *why* the JS isn't a copy of the grep patterns.

---

### Follow-up candidates (existing before this PR; not filed; outside #161's scope as written)

1. **Separators end a rule's scope even when quoted.** The wildcard rule and the force-push rule stop at a separator character even when it sits inside quotes, or inside a command substitution, in an earlier argument of the same command. The destructive target after it is then missed.
2. **Loose `rm` prefix.** The unanchored-wildcard rule's `rm` prefix isn't anchored to a command word and continues across `;`/`&&`. This causes false positives, and #3 makes them more common.
3. **Other ways to write the home and `/tmp` targets.** Repeated slashes after the home directory or `/tmp`, and the macOS real path of `/tmp`, aren't matched (#161 covers root and delete-flag spellings only).
4. **Force-push spellings.** Combined short flags that include the force flag, and force refspecs, aren't matched.
5. **opencode over-blocks on some Unicode input.** It treats non-ASCII spaces (NBSP, U+3000, BOM) as whitespace and puts word boundaries around non-ASCII letters, which BSD grep doesn't. So the opencode module blocks a few commands the Claude Code hook allows. This is the only sh/JS decision difference I found, and it's the same on `main`.
6. **zsh-only forms (low priority).** Negation globs under `extendedglob`, and nested or conditional `${…}` home spellings.

alexandrocuma:
Thanks for the review. The fixes are in `ed5e7ca` (hooks and tests) and `0bcdcfb` (docs), and `origin/main` is merged. Each finding, by class:

**1. Single-character glob.** Fixed in both implementations. The bare wildcard now ends a target, and the paren alternative covers only the remaining extglobs. Block cases cover each target: root, home, `/tmp` and `~/.claude`.

**2. Braced home spelling with a subscript.** Fixed in both implementations. A subscript opener is accepted after the name inside the braces. Block cases cover the bash and zsh forms, with and without `/.claude` after them.

**3. False positives from `:`, and the loose `rm` prefix.**
- `:` no longer ends a target. It counts only directly after an unbraced home parameter, as a separate alternative in the root/home rule.
- `rm` must now be a word of its own, not the end of a longer word or option. It may still be an argument of another command, so wrappers can't hide it.
- The wildcard rule needs the target in the same simple command as that `rm`. Its scan stops at `;`, `|` and a `&` that isn't part of a redirect. The opencode module finds the same boundary with a linear scan that mirrors the grep alternation.
- Every false positive you listed now has an allow case. Block cases show that a real `rm` after `;`, `&&`, `||`, `&` or `|`, as an argument, or with a redirect still blocks.
- As with `|` before, a separator inside quotes or a substitution in an earlier argument also ends the scan. That limit is documented and tracked with the other pre-existing items.

**4. Unpinned claims.** New cases cover:
- later stages and commands for the wildcard and force-push rules;
- the nesting limit at 99, 100 and 101 in both suites;
- the `.claude…/*` skip and whitespace;
- every home spelling and extglob opener;
- the force flag followed by `=`;
- `rm` word boundaries;
- redirect-joined `&`;
- a 400 KB pipeline timing check in the Claude Code suite, which times out on `main`.

I ported your mutant lists to the current code and added mutants for the fixes:
- **JS:** 92/97 killed.
- **sh:** 69/70 killed.
- **Survivors, all equivalent:**
  - The fork-bomb and reset stop sets don't matter, because their suffix search ignores the segment end.
  - The `git clean` class is the same language, and neither form backtracks.
  - A CRLF line split only removes a trailing CR, which the patterns already treat as whitespace.
  - The find-cache start check can't fire, because the parser's positions per character never decrease (checked on 562k calls).
  - In sh, the optional quote after `.claude` is redundant, because the target-end set already includes quotes.
  - A differential over 400k generated commands found no decision difference for any JS survivor.

**Minor.** Both notes are fixed:
- The docs now list every character that continues a target.
- The Claude Code timing now says a few seconds for a crafted 1 MB line: about 2.6 s for a 1 MB pipeline, measured.

The docs also state the expansion-then-literal trade-off and the new `rm` scope.

**Verification:**
- Every suite, `go test -race` and the installer tests pass after the merge.
- The sh hook and the JS module agree on 12,000 ASCII fuzz commands with BSD grep and 6,000 with GNU grep. Masking matches byte for byte on 200k inputs.
- The branch's JS matches a backtracking reference built from the new grep patterns on 450k inputs.
- Compared with `main`, every command `main` blocks and the branch allows traces to one of the three intended changes: `rm` inside a longer word, a separator between `rm` and the target, or `:`. None is unexplained.

alexandrocuma (commented):
## Follow-up review: `ed5e7ca`, `0bcdcfb` (at `ad40f41`)

**Recommendation**: Request changes. The fixes are good, but two in-diff changes now allow destructive commands that `main` blocks.

Concrete inputs are kept out of this public review, which describes each problem only by its kind.

### Earlier findings 1–3: verified fixed
- **Single-character glob, braced home spelling with a subscript, and the `$HOME:` modifier:** all block in both implementations. The `:` ending covers the unbraced `$HOME` spellings, `$~HOME` / `$=HOME` / `$^HOME` and the double-quoted form. The braced and quoted-then-literal spellings stay allowed, and zsh keeps those literal (checked with `print`).
- **The false positives I listed** (container runs that remove themselves on exit while mounting a protected directory, a later copy into `/tmp/$…`, `rm` inside a longer word) are all allowed now.

### 🟡 Important: new regressions against `main` in this diff

**1. A `;` or `&` inside quotes, an escape, backticks or a substitution now ends the wildcard rule's scan.**
`hooks/claude-code/dangerous-cmd-guard.sh` line 566; `hooks/opencode/dangerous-cmd-guard.js` `commandEnd`
- **What changed:** on `main` only `|` ended the scan. Now a `;` or `&` does too, even when it is data, not a separator: inside single or double quotes, backslash-escaped, in `$(…)` or backticks, in `${…}`, or in `$'…'`.
- **Effect:** a real delete of `/tmp/*` or `~/.claude/*` that has such an argument before the target is allowed. `main` blocked every form I tried.
- **Docs:** `docs/reference/hooks.md` calls this the same limit "as with `|` before". For `;` and `&` it is new in this PR.
- **Suggested fix** (cheap and linear): if a quote, backslash, backtick or `$(` / `${` / `$'` appears before the first separator, keep scanning to the next `|` as `main` did. In grep, that is one more optional group after the scan. In JS, check the characters in `[start, commandEnd)` and jump to the next `|` when one of them opens.
  - **Trade-off:** it over-blocks only when quoting comes before a real separator (`rm -rf "$x" && cp … /tmp/$y`). That is still less than `main` blocked.
  - **Checked in a scratch copy:** both suites pass unchanged (JS 368, sh 322). The sh hook and the JS module agree on 4,000 focused fuzz commands. It stays linear on crafted 1 MB inputs: JS ≤ 125 ms, sh hook ≤ 1.4 s.

**2. The wildcard rule now misses an `rm` followed directly by an expansion or a redirect.**
`dangerous-cmd-guard.sh` line 566 (`["'[:space:]]`); `dangerous-cmd-guard.js` rule 3 prefix (`(?=[\s"'])`)
- **What changed:** `rm` must now be followed by whitespace or a quote. But `rm` directly followed by a parameter expansion, a command substitution, backticks or a redirection (`<`, `>`, `&>`) is still the `rm` command, with its arguments after. Both shells word-split or strip those. `main` blocked these.
- **Suggested fix:** also accept `$`, a backtick, `<`, `>` and `&>` after `rm`. In grep, use an alternation that keeps `>&` / `<&` / `&>` whole so the scan's redirect handling still applies. Checked in the same scratch copy: suites pass and sh/JS agree.

**3. (Low) `rm` directly after a positional parameter.**
- **What changed:** a digit before `rm` now means "part of a longer word". A positional parameter written before `rm` (`$1` to `$9`, unset or empty) still leaves `rm` as the command, and `main` blocked it.
- **Suggested fix:** also accept a `$` plus one digit before `rm`. Checked in the same scratch copy.

### Newly-allowed classification
I generated about 220,000 commands and replayed each one against `main`'s JS, the branch's JS, and single reverts of the prefix rule, the separator stop and `:`. Every command `main` blocks and the branch allows traces to one of:
- `rm` inside a longer word or option (intended);
- a real separator between `rm` and the target (intended);
- a literal `:` (intended);
- the three regressions above.

The branch's JS matched a backtracking reference built from the new grep patterns on every input (0 differences).

### Tests
Mutations on the new code:
- **JS:** 22 mutants, 19 killed. Survivors:
  - an uppercase letter before `rm` (changes over-blocking only);
  - a single quote right after `rm`, which is not pinned;
  - `commandEnd` not resetting its redirect-join state, which changes decisions only on syntactically invalid redirects. It still breaks sh/JS parity, so pin it.
- **sh:** 21 mutants, 18 killed. Survivors:
  - an uppercase letter before `rm` (changes over-blocking only);
  - dropping the wildcard rule's direct `rm`-then-`/tmp` alternative, which changes the decision for a flagless delete with a single space before the target;
  - dropping the `$HOME:` ending from the `-fr` letter-order rule.

Add block cases for #1–#3, and pin the survivors.

### ✅ Also verified
- **Redirect edge cases:** `>&N`, `2>&1`, `&>`, `&>>`, `N>&-`, `<&0`, `>&` with a following word, `{fd}>&-`, `<>&N` and zsh `>>&` all continue the scan. `|&` and `>|` stop it, as `|` did on `main`. Both implementations agree.
- **Parity:** sh and JS agree on 4,000 focused fuzz commands, 0 differences. Non-ASCII whitespace still shows up only as the known over-blocking in the opencode module.
- **Linear time at 1, 2 and 4 MB** for the new scan (`;`/`&`/redirect runs, repeated `rm >&`, `.rm`/`-rm` runs, `"$HOME:` runs): JS ≤ 590 ms at 4 MB, sh hook ~1.9 / 3.6 / 7.2 s.
- **CI is green.** Fail-closed paths are unchanged.

alexandrocuma:
Thanks for the re-review. The fixes are in `c4c60db` (hooks and tests) and `64e9680` (docs), with `origin/main` (#158) merged in `af38ddb`. By finding class:

**1. A `;` or `&` that is data ended the wildcard scan.** Fixed in both implementations.
- The scan still stops at a real separator.
- Once a quote, backslash, backtick, `$(` or `${` appears between `rm` and that separator, the scan runs on to the next `|`, as it did on `main`. `$'…'` is covered by its quote.
- The docs no longer call this the same limit as `|`. They describe the fallback and its cost: quoting before a real separator now over-blocks.

**2. `rm` directly followed by an expansion or redirect.** Fixed.
- `rm` may now be followed by whitespace, a quote, an expansion, a backtick, a brace or glob character, or a redirect. That covers brace-expansion forms that still run `rm`, which `main` also blocked.
- The root/home rule is unchanged: it needs whitespace after `rm`, as on `main`.
- In the opencode module, a target directly after `rm` is only recognised in the one whitespace-led form the grep pattern allows. That keeps sh/JS parity when an expansion glued to `rm` starts a home spelling.

**3. A positional parameter before `rm`.** Fixed in both implementations.

**4. Survivors from your mutation run.** Block and allow cases now pin the single quote after `rm`, the redirect-join reset, the flagless single-space delete, the `-fr` `$HOME:` form, and an uppercase letter before `rm`. Every regression class from this review has block cases in both suites, and all of them fail on the previous head.

**Checks:**
- **Suites:** every suite, `go test -race` and the installer tests pass.
- **Parity, sh vs JS, 0 differences:**
  - your focused corpus, BSD and GNU grep;
  - your probe set;
  - ASCII fuzz with BSD and GNU grep;
  - masking, byte for byte.
- **Differential vs `main`:** every command `main` blocks and the branch allows falls in an intended class: `rm` inside a longer word, a real unquoted separator, or a literal `:`. 0 are unexplained. The branch's JS also matches a backtracking reference built from the grep patterns, with 0 differences.
- **Timing:** still linear at 1 MB and 4 MB.
- **Mutations:** your list and ours were run, and the remaining survivors are equivalent. Counts are in the PR description.

alexandrocuma (commented):
## Final follow-up: `c4c60db`, `64e9680` (at `64e9680`)

**Recommendation**: Request changes. One small gap is left in the same class as the separator regression; the rest is ready.

Concrete inputs are kept out of this public review, which describes each problem only by its kind.

### Earlier regression classes: verified fixed
Everything below blocks again in both implementations:
- a `;` or `&` inside double or single quotes, including `$'…'` (with and without an escaped quote inside), a backslash escape, backticks, `$(…)`, `${…}` (nested too, and with a command substitution inside);
- `rm` glued to an expansion, a backtick, a redirect, `&>`, `>&`, a brace, `*`, `?`, `[` or `,`;
- a positional parameter before `rm`.

Intended allows stay allowed: `rm` inside a longer word or option, a real unquoted separator, a literal `:`. A target glued straight onto `rm` with no separator is still allowed, which is correct: the shell reads that as one word, not `rm`. The `start + 1` change in the opencode module opens no gap there.

### 🟡 Important (in this diff)

**A process substitution that holds a `;` or `&` still ends the wildcard rule's scan.**
`hooks/claude-code/dangerous-cmd-guard.sh` (`OPENER`, `START_OPENER`); `hooks/opencode/dangerous-cmd-guard.js` (`commandEndQuoted`)
- **Issue:** the opener set covers quotes, escapes, backticks, `$(` and `${`, but not process substitution (`<(`, `>(`, and zsh `=(`). Its body can hold a separator that is data. So a delete of `/tmp/*` or `~/.claude/*` with such an argument before the target is allowed, both with a space after `rm` and glued to it. `main` blocks every form I tried. The shell runs these as a real delete: the substitution becomes a path argument (checked with `echo`/`print`).
- **Suggested fix:** add `[<>=]\(` to both opener sets and the matching check in `commandEndQuoted`. I tried it on a scratch copy of this head:
  - both suites pass (JS 415, sh 361);
  - sh/JS agree on the probes and on 4,000 focused fuzz commands;
  - the forms above block;
  - a real separator before a process substitution in a *later* command stays allowed.

### Newly-allowed classification
Over 300,000 generated commands (3×100k seeds, token set extended with process substitution, nested `${…}`, `$'…'` with an escaped quote, `rm` glued to `{ } , * ? [`, `$[…]`, `x[;]`, `(…)`):
- Every command `main` blocks that the branch allows is in an intended class, except the process-substitution class above.
- Two lookalikes are not real invocations:
  - `rm` followed by `=`, `:`, `.` or `/` is not the `rm` command;
  - a separator inside `$[…]` makes bash abort with an arithmetic syntax error before running anything (checked with `echo`).
- The branch's JS matched a backtracking reference built from the new grep patterns on all 300,000 (0 differences).

### Non-blocking notes
- **False positives from the opener fallback are the documented trade-off.** One common agent shape now blocks although `main` allowed it: a quoted or substituted first argument, then `&&` and a later command that writes to `/tmp/$…`. It fails safe; consider narrowing later.
- **The wider set after `rm`** adds only rare false positives: a glob or brace word starting with `rm` before a `/tmp/$…` or `~/.claude/*` argument of another command.
- **Speed:** linear to 4 MB. JS ≤ 352 ms; the sh hook ≤ 5.4 s at 4 MB, ≤ 2.0 s at 1 MB on the new opener and glued-`rm` inputs.

### Tests (mutation survivors on the new code)
- **JS:** 25 mutants, 20 killed. Survivors:
  - the target search starting one character later than `start + 1`. This changes a flagless single-space delete of the home `.claude` directory (no glob), so pin one;
  - `rm` glued to `,`, `?` or `[` (no block case in the JS suite);
  - the fallback's `|` search starting one character past a pipe end (over-blocking only).
- **sh:** 24 mutants, 20 killed. Survivors:
  - `rm` glued to `,`, `?` or `[` (the same gap as in JS);
  - the positional-parameter class narrowed to `$1`–`$9`. This is equivalent in effect, because `$0` is never empty.

Pin the flagless home-`.claude` delete and the three glued characters in both suites.

alexandrocuma:
The final-review fix is in `ae6e4a6` (hooks and tests) and `a2d145d` (docs).

- **Openers:** process substitutions (`<(`, `>(`, zsh `=(`) now count, in both implementations. So a `;` or `&` inside one no longer ends the wildcard-delete scan. Old-style arithmetic `$[` counts too, because it can hold `&` in the same way.
  - Right after `rm`, `=(` is left out: that form is an array assignment, not `rm`, and the opencode lookahead already rejects it.
  - The look-alikes you listed stay allowed.
- **Pins:** both suites now cover every process-substitution form, `$[`, flagless `rm` of the home target, and `rm` glued to `,`, `?` or `[`, plus the allowed look-alikes.
- **Timing:** the opencode scan now compares character codes. After this round's new test cases ran, two crafted 1 MB inputs in the same process were ten times slower. Now they aren't.
- **Checks:**
  - Every suite passes.
  - sh/JS parity shows 0 differences on your corpora and probes, with BSD and GNU grep.
  - The differential vs `main` has 0 unexplained newly allowed commands.
  - Timing is linear on opener runs.
  - Mutations on the changed lines are all killed except one equivalent bounds check. On the sh side, one survivor narrows `$N` to `$1`–`$9` before `rm`. It changes only an over-block, since `$0` never expands to `rm`.

Issue #155, opened 2026-09-17
CI: scheduled vulnerability scan, test-gated release workflow, stale CHANGELOG line refs


alexandrocuma:
## Problem
Follow-ups from the PR #153 review (#141):
1. **Advisories published while `main` is quiet** aren't caught until the next push, PR or tag. Add a weekly `schedule:` run of the vulnerability scan.
2. **`release.yml` runs no tests before goreleaser.** Make `ci.yml` reusable (`workflow_call`) and have the release workflow `needs:` it, so a tag can't publish from a commit whose tests fail.
3. **Stale line references in docs:**
   - `docs/development/setup.md` and `docs/development/testing.md` cite `CHANGELOG.md:147` for the #97 gofmt/vet note.
   - `docs/guides/release.md` cites `CHANGELOG.md:60`.
   Both have drifted. Point at version headings instead of line numbers.

## Acceptance Criteria
- [ ] A scheduled weekly run of the scan; a failure is visible (a failed run on `main`).
- [ ] A tag push runs the full CI suites before goreleaser, and a failure publishes nothing.
- [ ] No doc cites a `CHANGELOG.md` line number.

## Non-Goals
- Branch protection settings.

alexandrocuma:
# Verified Execution Plan — #155

**Ticket**: CI: scheduled vulnerability scan, test-gated release workflow, stale CHANGELOG line refs
**Type**: Config / infrastructure (CI) + docs
**Groomed**: 2026-09-17
**Validated against**: commit `8aa2d7a23da41c2f1d387c35d9a4f72a7506428a` (`fix/155`, from `main` 8aa2d7a)
**URL**: https://github.com/alexandrocuma/devexp-toolkit/issues/155

---

## Ticket Health Report

### ✅ Confirmed claims

- `.github/workflows/ci.yml` has only `pull_request` + `push: branches: [main]`; no `schedule`, no `workflow_call` (`ci.yml:3-6`).
- `release.yml` runs no tests: its only jobs are `govulncheck` and `goreleaser` (`release.yml:11-50`).
- The `govulncheck` job is **duplicated** today — `ci.yml:47-62` and `release.yml:14-27` are the same 3 steps against the same `scripts/govulncheck.sh`.
- `goreleaser` is the only job with `contents: write` (`release.yml:32-33`); workflow default is `contents: read` (`release.yml:8-9`); `needs: govulncheck` at `release.yml:30`. #141's structure is exactly as the ticket describes.
[...]

PR #164, opened 2026-09-17
fix(release): cut the platform release the way the repo's release guide says (#154)

### LINKED ISSUE #154: /release: create the platform release the way the repo's release guide says
## Problem
`skills/release/SKILL.md` Phase 6 always runs `gh release create … --notes` right after the tag push. It doesn't use `--verify-tag`, and it ignores the release guide's own Cut section.

A repo's guide can require something different. In devexp-toolkit (PR #153, #141), the release is created as a draft that goreleaser publishes only after the build succeeds. With the generic flow, a failed or blocked build leaves an empty release marked Latest.

## Acceptance Criteria
- [ ] Phase 6 uses the guide's Cut commands for tag and release creation whenever a release guide exists, and falls back to the current generic commands only when it doesn't.
- [ ] The generic fallback uses `--verify-tag` and a notes file.
- [ ] The generic skill never adds `--draft` on its own: a repo without a publisher step would never publish the draft.
- [ ] `docs/reference/skills.md` and the `gen-docs`/`update-docs` Release Guide template say that the Cut section may define release creation.

## Non-Goals
- Changing any repo's CI.

alexandrocuma:
Closes #154

## What changed
`/release` Phase 6 (now "Tag and Create the Release") reads the repo's release guide **Cut** section and creates the platform release the way that section says, instead of always publishing one right after the tag push.

- **Where the command comes from:** the guide gives it → run it as written, substituting only version, tag and notes file; the guide says nothing → use the generic commands; the guide mentions it but the command is missing, ambiguous or still `[CONFIRM]` → **stop the cut** and send the user to `/devxp`. The third case must never quietly become the second.
- **Every refusal is pre-push.** Phase 1 classifies the Cut section read-only — with Phase 6's own `sed` range and empty-output guard, so the classification can't drift from what Phase 6 will do — alongside the per-target prerequisites check, and a Cut that can't be executed blocks the cut there. The notes extraction and its `[ -s ]` check sit before the commit too, since they read the CHANGELOG Phase 4 finalised. So the order is classify → notes → commit/tag/push → tag-on-remote check → create, and nothing can refuse after the tag is public. The preflight inventory shows the resolved `Release object :` (including the publisher target, and `generic — no release guide` for a repo with none), and the Phase 2 gate prints the command it is about to run.
- **Empty is not silent.** If the `sed` range finds no `## Cut…` heading, that is not proof the guide says nothing (`## The Cut`, `### Cut`); a grep for release-object wording decides between row 2 and row 3.
- **Generic commands** use a notes **file** (`awk`-extracted from the version's own CHANGELOG section into `/tmp/.release-${ticket}-notes.md`, the scratch path Phase 8 already sweeps) and `--verify-tag`, and they use the tag the cut actually pushed rather than a literal `v<version>`, since a guide may name another tag format. The extraction is checked with `[ -s ]` — a `v`-prefixed version yields 0 bytes and would ship a release with no notes.
- **The tag is confirmed on the remote before any release is created**, guide-provided commands included, because the check is read-only: `git ls-remote --tags origin "refs/tags/$tag" | grep -q .`. `ls-remote` exits 0 whether or not the tag exists, so it is the output that must be non-empty; `glab` has no `--verify-tag` and tags the default branch itself when the tag is missing.
- **Never `--draft` on its own:** a draft belongs only to a guide that names what publishes it. Otherwise the draft is never published and users get no release at all.
- **An unpublished release means the cut isn't finished:** Phase 6 names the target whose Build publishes it and binds it **watch-only** even when the change didn't affect it, so Phase 7 watches that pipeline and reads the guide's Post-release verification signals (Phase 9's table marks that target watch-only). Phase 8's completion gate now also requires the release object to be published before anything is retired — with no escape hatch: an unpublished object leaves the watch-only target `awaiting-external`, which already holds the phase, and `/release <ticket>` retires later. If the publisher fails, the generic shape is: confirm it is *still* unpublished (`gh release view --json isDraft,isPrerelease`, or `glab release view`), then delete it (never `--cleanup-tag` / `--with-tag`) — a published object is rolled forward, not deleted — fix on a branch, cut the next patch. The tag stays.
- Phase 2's gate line and Phase 9's `Platform:` line reflect all four outcomes; two Guidelines bullets cover the rule.
- **Docs:** `docs/reference/skills.md` (with `--verify-tag` qualified as GitHub-only), one identical line in the Release Guide template of `agents/gen-docs.md` and `agents/update-docs.md` (kept byte-identical, as the repo requires), and a clause in `docs/guides/release-targets.md`.

This is what bit us on v0.9.2: the guide's draft flow (added in #141) meant the skill's generic command would have published an empty release, so I ran the guide's command by hand.

## Verification
- **Walked six paths** after the reordering: this repo's guide (row 1, draft flow); no guide (row 2, plain published release); a half-written Cut with `[CONFIRM]` (row 3, stops at preflight); a monorepo guide naming `api@<version>` with a silent Cut (row 2, tag propagated); GitLab; no platform CLI.
- **Ran the reviewer's failure shapes for real:** `ls-remote` on an absent tag (exits 0, empty — the new guard stops; present tag proceeds), the `awk` extractor with and without the `v` prefix (0 bytes vs 15 070), the `sed` range against `## Cut …`, `## The Cut` and `### Cut`, and the grep that distinguishes row 3 from row 2.
- **`glab` and `gh` flags checked against the installed CLIs:** `gh` has `--verify-tag`, `--notes-file`, `--cleanup-tag` on `release delete`, and `isDraft`/`isPrerelease` on `release view`; `glab` has `--notes-file` but no `--verify-tag`.
- **Every cited path and section name exists.**
- `gen-docs.md` and `update-docs.md` Release Guide blocks `diff` clean (65 lines each).
- `./install.sh --dry-run` deploys the changed assets and ends `All done.`
- `gofmt`, `go vet`, `go test -race ./...`, all 7 Claude Code hook suites, all 5 opencode hook suites, and `install.test.sh` / `remote-install.test.sh` / `uninstall.test.sh` pass.

## Review
All 12 inline comments from the first review are addressed in 88816e0; the seven from the re-review are addressed in 074ee1e. Every thread has a reply.

## Out of scope (follow-up filed)
`agents/changelog.md` has the same hazard in its own publish step: inline multiline `--notes`, no `--verify-tag`, no awareness of the guide — #165. This repo's guide labelling the bullet `GitHub Release object:` rather than `Release object:` is being taken separately.


skills/release/SKILL.md
```diff
@@ -185,16 +193,42 @@ git tag -a "v<version>" -m "Release v<version>"
...
+
+Never let the third row slide into the second. A repo whose guide asks for anything other than a plain published release is exactly the repo where a guess ships an empty release to users.
+
+Generic commands — only when the guide defines none:
 
 ```bash
-gh release create  "v<version>" --title "v<version>" --notes "<changelog entry>"   # GitHub
-glab release create "v<version>" --name  "v<version>" --notes "<changelog entry>"  # GitLab
+gh   release create "v<version>" --title "v<version>" --notes-file "$notes" --verify-tag   # GitHub
+git ls-remote --tags origin "refs/tags/v<version>"                                         # GitLab: stands in for --verify-tag
```
alexandrocuma: 🔴 **Critical — the GitLab `--verify-tag` stand-in fails open, so this path can still do the thing the PR exists to prevent.**

`git ls-remote --tags origin "refs/tags/v<version>"` **exits 0 with empty output** when the tag is absent. I ran it:

```
$ git ls-remote --tags origin refs/tags/v99.99.99; echo "exit=$?"
exit=0
$ git ls-remote --tags origin refs/tags/v99.99.99 && echo "AGENT WOULD PROCEED"
AGENT WOULD PROCEED
```

Line 223 says only *"the `ls-remote` check has to pass first"* — pass **what**? Exit status is the natural reading, and it's the one an agent chaining `cmd-a && cmd-b` will take. It then runs `glab release create`, which by its own help *"creates a release from the latest state of the default branch, and tags it with the tag name you specify"* — the default branch gets tagged, exactly the hazard this paragraph claims to close.

**Suggested wording** (make the test the output, and say what failure means):

```bash
git ls-remote --tags origin "refs/tags/v<version>" | grep -q . \
  || { echo "tag never reached the remote — stop, do not create the release"; exit 1; }
glab release create "v<version>" --name "v<version>" --notes-file "$notes"
```

and on line 223: *"`ls-remote` exits 0 whether or not the tag exists — it is the **output** that has to be non-empty. Empty output means the tag never reached the remote: stop, and do not run `glab release create`."*

skills/release/SKILL.md
```diff
@@ -176,7 +176,15 @@ Then apply each affected target's **Versioning** rule from the guide — typical
...
-### Phase 6 — Tag and Publish
+### Phase 6 — Tag and Create the Release
+
+The guide's **Cut** section owns this phase the way each target's section owns Phase 7. Read it before running anything:
+
+```bash
+sed -n '/^## Cut/,/^## /p' docs/guides/release.md 2>/dev/null
+```
+
+Commit and tag — with the guide's Cut format for the release commit, the tag name and the tag message wherever it states them:
```
alexandrocuma: 🔴 **Critical — the "stop the cut" branch is written *after* the point of no return, so it can only fire once the tag is public.**

Executed top-to-bottom, Phase 6 now reads: read the Cut section (184) → **commit, tag, push** (189-194) → *"point of no return"* (196) → notes file (200-203) → *then* the decision table whose third row says **"stop the cut"** (211).

For scenario (c) — a guide that mentions a release object but leaves the command half-written or `[CONFIRM]` — the agent discovers it must stop only after it has pushed the tag. In this repo that tag has already started `release.yml` (`docs/guides/release.md:21`, *"Any pushed `v*` tag starts the `release` workflow"*), and with `release.use_existing_draft: true` goreleaser finds no draft and creates its own release with its own commit-list notes (`docs/guides/release.md:23`). "Stop" at that point is not a stop.

There's a matching inconsistency upstream: Phase 2's gate line (SKILL.md:124) now promises to print `(<command from the guide's Cut | generic>)`, which requires the three rows to have been **resolved before the gate** — but nothing in Phase 0 or Phase 1 says to read or classify the Cut section, and Phase 1's `[CONFIRM]` check (line 90) is explicitly *per target*, so a `[CONFIRM]` in **Cut** is never caught in preflight.

**Suggested fix:** hoist the `sed` read **and the decision table** above this "Commit and tag" block, and add one line to Phase 1: *"Read the guide's **Cut** section read-only and classify it against Phase 6's table. A Cut that mentions a release object without a runnable command, or still carries `[CONFIRM]`, blocks the cut the way a `[CONFIRM]` step blocks a target — report it at preflight and do not offer the gate."* Then row 3 costs nothing, and Phase 2 can honestly print the command it's about to run.

skills/release/SKILL.md
```diff
@@ -176,7 +176,15 @@ Then apply each affected target's **Versioning** rule from the guide — typical
 
 ---
 
-### Phase 6 — Tag and Publish
+### Phase 6 — Tag and Create the Release
+
+The guide's **Cut** section owns this phase the way each target's section owns Phase 7. Read it before running anything:
+
+```bash
+sed -n '/^## Cut/,/^## /p' docs/guides/release.md 2>/dev/null
```
alexandrocuma, line 187: 🟡 **Important — this `sed` returns empty for a heading that isn't literally `## Cut…`, and empty is indistinguishable from row 2.** That is precisely the row-3-slides-into-row-2 failure line 213 forbids, caused by the extractor rather than by the reader.

Verified:

```
## Cut (shared by all targets)   →  prints the section          ✅  (this repo, and the gen-docs template)
## Cut  … as the LAST section    →  prints to EOF               ✅  (acceptable — too much, never too little)
## The Cut                       →  prints NOTHING              ❌
### Cut                          →  prints NOTHING              ❌
```

Any guide written before this template, or by hand, is a candidate. The agent then reads "the guide says nothing about creating a release" and takes the generic path — in a repo whose guide *does* describe a draft.

**Suggested wording**, right under the snippet: *"Empty output is not proof the guide is silent — the heading may simply be worded differently. Confirm with `grep -niE 'release object|release create|--draft' docs/guides/release.md` before using the generic commands. A guide that clearly discusses release creation under a heading this range can't find is an unfinished guide: row 3."*

skills/release/SKILL.md
```diff
@@ -185,16 +193,42 @@ git tag -a "v<version>" -m "Release v<version>"
...
+awk -v v="<version>" '$0 ~ "^## \\[" v "\\]" {f=1; next} f && /^## \[/ {exit} f' CHANGELOG.md > "$notes"
+```
+
+Then create the platform release object. Where that command comes from:
+
+| The guide's Cut section… | Do |
+|--------------------------|----|
+| gives the command | run exactly that, substituting version, tag and notes file — never add, drop or reorder a flag |
+| says nothing about creating a release | use the generic commands below |
+| mentions it, but the command is missing, ambiguous or still `[CONFIRM]` | **stop the cut.** Quote what the guide says and send the user to `/devxp` to finish the Cut section |
```
alexandrocuma: 🟡 **Important — row 3's "stop the cut" needs somewhere to land.** Two loose ends:

1. **When** it fires — see the comment on line 187; as written it can only fire after the tag push.
2. **How it is reported** — Phase 9's `Platform:` line (SKILL.md:339, also changed in this PR) offers exactly three outcomes: *published*, *created unpublished — published by `<target>`*, and *tag only — no platform detected*. A cut stopped at row 3 on GitHub fits none of them: a platform **was** detected, and "tag only" reads as "this repo has no platform", which is false and misleading on a resumed run.

**Suggested:** add a fourth outcome to line 339, e.g. `/ not created — the guide's Cut is unfinished (run /devxp)`, and have row 3 say which report line it produces.

skills/release/SKILL.md
```diff
@@ -185,16 +193,42 @@ git tag -a "v<version>" -m "Release v<version>"
...
 
-**Tag push is the point of no return for the cut.** Everything before it is local and revertible; everything after is public. If a step here fails, stop — do not proceed to Phase 7, and do not retry a push that may have partially succeeded without checking `git ls-remote --tags origin` first.
+**A release the guide leaves unpublished is not shipped yet.** When the Cut's command creates a draft, a pre-release or anything else users cannot see, something else publishes it — typically a tag-triggered pipeline in a target's **Build**. Until then the cut is incomplete: name that target, never report the release as published, and let Phase 7 watch that pipeline and read the guide's **Post-release verification** signals for it (no longer a draft, assets attached, the version resolving as latest).
+
+**If that publisher fails or never runs**, follow the guide's steps for the target. Where it gives none, the generic shape is: delete the unpublished release object — it was never public, so this is not the release deletion a rollback forbids — fix on a branch, then cut the next patch version. The tag stays; tags are never moved, deleted or re-pushed.
```
alexandrocuma: 🟡 **Important — "delete the unpublished release object" isn't bounded in time, only in intent.**

The carve-out is correct *if* the object is still unpublished at the moment of deletion. But this paragraph is reached after the publisher failed, and the guide's own documented first move for the commonest failure is `gh run rerun <run-id> --failed` (`docs/guides/release.md:45`), which *"still finds the draft"* and publishes it. A user who re-runs in the GitHub UI while the agent is here, or an agent that retried and then re-read a stale status, deletes a **published** release with users already on it — the one deletion Rollback forbids (`docs/guides/release.md:65`).

The instruction also gives no command, while the guide gives an exact one — so the generic shape is less safe than the specific one it is generalising.

**Suggested wording:**

> Where it gives none, the generic shape is: **confirm the object is still unpublished, then delete it** — `gh release view "v<version>" --json isDraft,isPrerelease` must still show it unpublished; only then `gh release delete "v<version>" --yes` (never `--cleanup-tag`). If it has since been published, it is a normal release: do not delete it, roll forward instead. Then fix on a branch and cut the next patch version. The tag stays; tags are never moved, deleted or re-pushed.

skills/release/SKILL.md
```diff
@@ -185,16 +193,42 @@ git tag -a "v<version>" -m "Release v<version>"
...
+| mentions it, but the command is missing, ambiguous or still `[CONFIRM]` | **stop the cut.** Quote what the guide says and send the user to `/devxp` to finish the Cut section |
+
+Never let the third row slide into the second. A repo whose guide asks for anything other than a plain published release is exactly the repo where a guess ships an empty release to users.
+
+Generic commands — only when the guide defines none:
 
 ```bash
-gh release create  "v<version>" --title "v<version>" --notes "<changelog entry>"   # GitHub
-glab release create "v<version>" --name  "v<version>" --notes "<changelog entry>"  # GitLab
+gh   release create "v<version>" --title "v<version>" --notes-file "$notes" --verify-tag   # GitHub
```
alexandrocuma: 🟡 **Important — the generic commands hardcode `v<version>`, but row 2 can be reached by a guide that names a different tag format.**

Scenario (d) splits in two. A guide can state a tag format — the template you added at `agents/gen-docs.md:556` explicitly offers `<target>@<version>` for monorepos — while saying **nothing** about a release object. That is row 2, so these commands run, with the literal `v<version>`:

- GitHub: `--verify-tag` aborts (safe-fail, but the cut ends with a pushed tag and no release, and the instructions don't say that's what happened).
- GitLab: the tag doesn't exist under that name → straight into the failure in my comment on line 219.

Line 187 correctly tells the agent to take the tag name from the guide for `git tag`, but that substitution never propagates into these commands.

**Suggested:** *"Generic commands — only when the guide defines none. Use the tag the cut actually pushed, not literally `v<version>`: if the guide names a tag format, these carry it too."* Also worth one line: the `ls-remote`/`--verify-tag` check is read-only and cheap, so run it before **any** release-creation command, including a guide-provided one that happens to omit `--verify-tag` (row 1 forbids *adding* flags, but not running a separate read-only check first).

alexandrocuma (commented):
## PR Review: fix(release): cut the platform release the way the repo's release guide says (#154)

**Recommendation: Request changes** — the direction is right and the acceptance criteria are met, but three of the instructions are executable in a way that reproduces the very failure the PR is closing. Posting as a comment review; all findings below are *inside this diff*, so per the deliver scope rule they're fix-now, not follow-ups.

**Summary.** Phase 6 becomes guide-driven, with a three-row decision table, notes-as-a-file, `--verify-tag`, and a hard prohibition on a self-invented `--draft`. The unpublished-release state is threaded through Phase 2, Phase 7 and Phase 9. I walked all six scenarios you listed and ran both extractors and both CLIs for real.

### What I verified

| Check | Result |
|---|---|
| `gh` flags (`gh 2.100.0`) | `--verify-tag`, `--notes-file`, `--draft` all present — claim accurate |
| `glab` flags (`glab 1.116.0`) | `-F/--notes-file` and `-n/--name` present, **no** `--verify-tag`; its own help confirms *"If the Git tag you specify doesn't exist, the command creates a release from the latest state of the default branch, and tags it with the tag name you specify"* — claim accurate |
| `awk` notes extractor | pulls the 228-line `[0.9.2]` section and stops at the next `## [` — accurate. Also: a body line containing `## [` inside a list item does **not** truncate it; a last-in-file section runs to EOF |
| `sed` Cut extractor on this repo's guide | prints the 8-line Cut section (+ the trailing `## Target: cli` header) — works |
| Release Guide template in `agents/gen-docs.md` vs `agents/update-docs.md` | extracted both blocks and `diff`ed: **byte-identical** ✅ |
| Table row 1 vs this repo's `docs/guides/release.md:23` | parseable — the bullet does carry a complete, runnable `gh release create … --draft … --verify-tag` |
| Go suite (`-race -cover`), 7 Claude Code hook suites, 5 opencode hook suites, 3 installer script suites, `./install.sh --dry-run` | all pass; dry-run deploys 8 skills |

### Scenario walk (what the instructions actually produce)

- **(a) this repo** — `sed` prints the Cut section, row 1 matches, and the command run is the guide's `gh release create v<version> --draft --title v<version> --notes-file "$notes" --verify-tag`. Correct. Phase 6 then names `cli` as the publisher and Phase 7 watches `release.yml`. ✅
- **(b) no guide** — Phase 0 already offers `/devxp`/cut-only; `sed` prints nothing, row 2, generic commands. ✅
- **(c) half-written or `[CONFIRM]` Cut** — row 3 fires… but only *after* the tag has already been pushed (finding 2). ⚠️
- **(d) guide names a different tag format / commit message** — covered for `git tag` by the new line 187, **not** covered for the generic release commands, which still hardcode `v<version>` (finding 5). ⚠️
- **(e) GitLab** — the `ls-remote` stand-in fails open (finding 1). 🔴
- **(f) no platform CLI** — line 227 handles it; the notes file is written and unused, harmless. ✅

### Safety answers to your specific questions

- *Creates a release nothing publishes?* Only via finding 6 (the publishing target may not be among the bound targets).
- *Publishes something the guide wanted unpublished?* No — row 1 forbids dropping `--draft`.
- *Deletes a published release?* Not by intent, but the delete step is unbounded in time (finding 3).
- *Moves/deletes a tag, pushes twice?* No — explicitly forbidden, and the point-of-no-return paragraph survived the move.
- *Scratch notes file vs Phase 8 globs / empty `$ticket`?* `/tmp/.release-154-notes.md` does match `/tmp/.release-"$ticket"-*`. With an empty ticket id the path is a fixed name, so **no** blanket-glob hazard — but the file is then never retired, and the same is true whenever the release doesn't complete (finding 8).

### ✅ Done well

- The three-row table plus *"Never let the third row slide into the second"* is exactly the right shape for prose an agent executes under pressure — the failure mode is named, not just the happy path.
- The carve-out on deleting a draft (*"it was never public, so this is not the release deletion a rollback forbids"*) prevents the obvious misreading of the Rollback rule.
- Phase 2, Phase 7, Phase 9 and both Guidelines bullets were all carried along — no half-updated surface.
- The two docs agents' templates really are byte-identical; the PR body's claims all check out against the installed CLIs.

### Follow-up candidates (pre-existing, outside this diff — not blocking)

1. `agents/changelog.md:146-147` — same hazard, inline `--notes`, no `--verify-tag`, no guide awareness. Already #165. ✅
2. `docs/guides/release.md:23` — the bullet is labelled `GitHub Release object:` while the new template prescribes `Release object:`; the next `/devxp` refresh will want to reconcile them. The runnable command is also buried mid-paragraph next to historical prose (*"Before this, `/release` created a published release…"*) — the hardest possible shape for row 1's "run exactly that".
3. `docs/guides/release.md:21` — the Cut states a tag message format (`Release v<version> — <summary>`) the skill's generic tag block doesn't produce; finding 2/5 would make this moot, but the guide could be explicit.

---

*Reviewed read-only in the worktree at `fix/154` @ `edb22b7`.*

skills/release/SKILL.md
```diff
@@ -185,16 +193,42 @@ git tag -a "v<version>" -m "Release v<version>"
...
+
+**Never add `--draft` to the generic commands.** A draft is right only where the guide names what publishes it; with no publisher the draft stays unpublished forever and users get no release at all.
+
 With no platform detected, the tag **is** the release. Report that rather than failing.
 
-**Tag push is the point of no return for the cut.** Everything before it is local and revertible; everything after is public. If a step here fails, stop — do not proceed to Phase 7, and do not retry a push that may have partially succeeded without checking `git ls-remote --tags origin` first.
+**A release the guide leaves unpublished is not shipped yet.** When the Cut's command creates a draft, a pre-release or anything else users cannot see, something else publishes it — typically a tag-triggered pipeline in a target's **Build**. Until then the cut is incomplete: name that target, never report the release as published, and let Phase 7 watch that pipeline and read the guide's **Post-release verification** signals for it (no longer a draft, assets attached, the version resolving as latest).
```
alexandrocuma: 🟡 **Important — "let Phase 7 watch that pipeline" only works if that target was bound in Phase 0, and it may not have been.**

Phase 0 binds *affected* targets: the groom plan's list, or the branch's changed paths mapped onto each target's path (SKILL.md:72). The target whose **Build** publishes the release object is not necessarily one of them. A docs-only or hook-only release in this very repo can map to `toolkit-clone` alone (or to nothing), while the draft is published by `cli`'s `release.yml`. Then:

- Phase 7 ships the bound targets and never watches the publisher;
- Phase 8's completion gate — *"the cut succeeded **and** every bound target is `shipped` or explicitly `skipped`"* (SKILL.md:292) — is satisfied, so the worktree, plan, release state and scratch are all retired;
- the release object is still a draft, and this PR's whole premise is that that means **nothing shipped**.

**Suggested:** two sentences. In this paragraph: *"Bind that target for this run even if the change didn't affect it — watch-only: Phase 7 watches its pipeline and reads its **Post-release verification** signals, but runs none of its build or distribute commands and needs no target gate."* And in Phase 8's gate: *"…and the release object is published (or the user has explicitly accepted it staying unpublished)."*

skills/release/SKILL.md
```diff
@@ -185,16 +193,42 @@ git tag -a "v<version>" -m "Release v<version>"
 git push && git push --tags
 ```
 
-Publish on the platform detected in Phase 1:
+**Tag push is the point of no return for the cut.** Everything before it is local and revertible; everything after is public. If a step here fails, stop — do not proceed to Phase 7, and do not retry a push that may have partially succeeded without checking `git ls-remote --tags origin` first.
+
+The release notes are always the version's own `CHANGELOG.md` section, passed as a **file**. A multiline body inlined into `--notes` is at the mercy of the shell:
+
+```bash
+notes="/tmp/.release-<ticket>-notes.md"   # retired with the rest of the scratch in Phase 8
```
alexandrocuma: 🔵 **Minor — two things about the notes file.**

**(a) The path is a placeholder where Phase 8's cleanup is a variable.** Phase 8 deletes `/tmp/.release-"$ticket"-*` (SKILL.md:324). An agent that copies `"/tmp/.release-<ticket>-notes.md"` literally — plausible, since every other token in this block (`<version>`, `<platform>`) *is* meant to stay symbolic until substituted — creates `/tmp/.release-<ticket>-notes.md`, which never matches that glob. Write it as `notes="/tmp/.release-${ticket}-notes.md"` so it lines up with Phase 0's `ticket=` binding and Phase 8's glob by construction.

Also, the comment *"retired with the rest of the scratch in Phase 8"* is only true on the happy path: Phase 8 is skipped entirely whenever the release isn't complete (by design — "failure preserves"), and its safety gate aborts **all** cleanup when the ticket id is empty, which Phase 0 explicitly supports (*"Without a ticket id — release the current branch"*). No hazard — the path is a fixed name, not a glob, so there's no blanket-wipe risk — but the comment overpromises. Suggest *"retired with the rest of the scratch in Phase 8, when the release completes with a ticket id"*.

**(b) Nothing checks that the extraction found anything.** The `awk` writes a 0-byte file when the heading doesn't match, and `gh release create --notes-file` on an empty file happily ships a release with **no notes**. The most likely mismatch is the `v` prefix — this repo's headings are `## [0.9.2]`, the tags are `v0.9.2`:

```
$ awk -v v="v0.9.2" '...' CHANGELOG.md | wc -c
0
$ awk -v v="0.9.2"  '...' CHANGELOG.md | wc -l
228
```

Suggest adding, right after the `awk`: `[ -s "$notes" ] || { echo "no [<version>] section in CHANGELOG.md — stop"; exit 1; }` and a note that `<version>` here is the bare version, **without** the `v` the tag carries.

skills/release/SKILL.md
```diff
@@ -302,7 +336,7 @@ With `$ticket` verified, retire each artifact:
   Changelog:    <N entries — N feat, N fix, N breaking / skipped>
   Version:      <vOLD → vNEW (patch|minor|major) / tag-only, no version file>
   Tag:          <v<version> pushed / not pushed / already existed (resumed)>
-  Platform:     <published on github|gitlab / tag only — no platform detected>
+  Platform:     <published on github|gitlab / created unpublished on github|gitlab — published by <target> / tag only — no platform detected>
```
alexandrocuma: 🔵 **Minor — needs a fourth outcome.** A cut stopped by table row 3 (SKILL.md:211) with a platform detected matches none of these three strings, and `tag only — no platform detected` would be actively wrong. Suggest appending ` / not created — the guide's Cut is unfinished (/devxp)`.

docs/reference/skills.md
```diff
@@ -59,7 +59,8 @@ The first four orchestrators *build* software and `/release` *ships* it; `/monit
...
-- **The release guide is the only source of ship commands** — never improvised; a `[CONFIRM]` step or a missing prerequisite (unauthenticated CLI, absent secret) marks the target `blocked`. No guide → offers `/devxp`, otherwise cut-only
+- **The release guide is the only source of release commands** — its **Cut** section may define how the tag and the platform release are created (a draft that a build pipeline publishes once its assets are uploaded, say), and each target's section defines the ship steps. Never improvised; a `[CONFIRM]` step or a missing prerequisite (unauthenticated CLI, absent secret) marks the target `blocked`. No guide → offers `/devxp`, otherwise cut-only
+- **The cut's generic fallback is deliberately plain:** with no Cut definition, `/release` creates a published release with the tag verified (`--verify-tag`) and the changelog section passed as a notes **file**, never an inline `--notes` string. It never adds `--draft` on its own — a repo with no publisher step would leave the draft unpublished forever
```
alexandrocuma: 🔵 **Minor — `--verify-tag` is GitHub-only, and the reference reads as if it applies to both platforms.**

Verified against `glab 1.116.0`: `glab release create` has `-F/--notes-file` and `-n/--name`, and **no** `--verify-tag` — which is exactly why SKILL.md:219 adds the `git ls-remote` stand-in. As written this bullet promises a guarantee GitLab doesn't get.

**Suggested:** *"…creates a published release with the tag verified (`--verify-tag` on GitHub; an explicit `git ls-remote` check before `glab`, which has no such flag) and the changelog section passed as a notes **file**…"*

agents/gen-docs.md
```diff
@@ -554,6 +554,7 @@ Kinds: `library` · `cli` · `web` · `service` · `ios` · `android` · `deskto
 
 - Version source of truth: `<file>` — scheme: semver
 - Tag format: `v<version>` (or `<target>@<version>` in a monorepo with independent versions)
+- Release object: `<command that creates the platform release>` — write this line only when the repo needs something other than a plain published release (e.g. a draft that a target's build pipeline publishes once it has uploaded its assets). `/release` runs it verbatim; omit it and `/release` creates a published release from the version's changelog section. Name whatever publishes it in that target's **Build**
```
alexandrocuma: 🔵 **Minor — "runs it verbatim" contradicts the skill's row 1, and the template's own placeholder is the proof.**

The example value is `<command that creates the platform release>`, and this repo's real guide writes `gh release create v<version> --draft --title v<version> --notes-file <section> --verify-tag` — three placeholders. `skills/release/SKILL.md:209` says *"run exactly that, **substituting** version, tag and notes file"*. A guide author reading "verbatim" may reasonably conclude they must hardcode a version, and an agent reading it may hesitate to substitute.

**Suggested:** *"`/release` runs it as written, substituting only the version, tag and notes file; it never adds, drops or reorders a flag."*

(Also worth noting for a later pass, not this PR: this repo's own guide labels the bullet `GitHub Release object:` rather than `Release object:` — `docs/guides/release.md:23`. Harmless for matching, but the next `/devxp` refresh will want them to agree.)

Same line in `agents/update-docs.md:562` — I confirmed the two Release Guide template blocks are byte-identical today, so **both** need the same edit to stay that way.

agents/update-docs.md
```diff
@@ -559,6 +559,7 @@ Kinds: `library` · `cli` · `web` · `service` · `ios` · `android` · `deskto
 
 - Version source of truth: `<file>` — scheme: semver
 - Tag format: `v<version>` (or `<target>@<version>` in a monorepo with independent versions)
+- Release object: `<command that creates the platform release>` — write this line only when the repo needs something other than a plain published release (e.g. a draft that a target's build pipeline publishes once it has uploaded its assets). `/release` runs it verbatim; omit it and `/release` creates a published release from the version's changelog section. Name whatever publishes it in that target's **Build**
```
alexandrocuma: 🔵 **Minor — mirror of the `agents/gen-docs.md:557` comment.** I extracted the `#### Release Guide` block from both agents and `diff`ed them: currently **byte-identical** ✅ (64 lines each). Any reword of "runs it verbatim" has to be applied here in the same commit to keep that invariant.

skills/release/SKILL.md
alexandrocuma: Fixed in 88816e0 — you're right, and I reproduced it: `git ls-remote --tags origin refs/tags/v99.99.99` exits 0 with no output, so the `&&` chain proceeded.

The check now tests the output and stops explicitly, and it was moved up so it guards **every** release-creation command, including a guide-provided one that omits `--verify-tag`:

```bash
git ls-remote --tags origin "refs/tags/$tag" | grep -q . \
  || { echo "tag never reached the remote — stop, do not create the release"; exit 1; }
```

The prose now says: *"`ls-remote` exits 0 whether or not the tag exists — it is the **output** that has to be non-empty. Empty output means the tag never reached the remote: stop, and create nothing."* Verified: absent tag → stops, present tag → proceeds.

skills/release/SKILL.md
alexandrocuma: Fixed in 88816e0. The `sed` read and the decision table are hoisted above the commit/tag block, so row 3 reads *"stop the cut here, before the commit — nothing is pushed"*, and the paragraph after the table makes the cost explicit: stopping costs nothing here; after the tag push it costs a public tag that may already have started a pipeline.

Phase 1 now classifies the cut once, read-only, next to the per-target prerequisites check:

> **The cut itself**, once: read the guide's **Cut** section read-only and classify it against Phase 6's table. A Cut that mentions a release object without a runnable command, or still carries `[CONFIRM]`, blocks the cut the way a missing prerequisite blocks a target — report it here and do not offer the gate; `/devxp` is the fix. Row 1 resolves the exact command the gate will show.

The preflight inventory gained a `Release object :` line, and the Phase 2 gate now prints *"Tag, then create the release on `<platform>` — `<the command Phase 1 resolved>`"* rather than promising a classification nothing had made.

skills/release/SKILL.md
alexandrocuma: Fixed in 88816e0, with your wording. Right under the snippet:

> Empty output is not proof the guide is silent — the heading may simply be worded differently (`## The Cut`, `### Cut`). Confirm with `grep -niE 'release object|release create|--draft' docs/guides/release.md` before treating it as row 2. A guide that clearly discusses release creation under a heading this range cannot find is an unfinished guide: row 3.

Walked it on a fixture with `## The Cut` + a `Release object:` bullet: `sed` returns nothing, the grep hits, verdict row 3. And on a monorepo guide that genuinely says nothing about a release object: grep is empty, verdict row 2.

skills/release/SKILL.md
alexandrocuma: Both fixed in 88816e0.

1. Row 3 now fires before the commit — see the reply on the Phase 6 ordering comment.
2. Phase 9's `Platform:` line gained the fourth outcome: `… / not created — the guide's Cut is unfinished (/devxp) / tag only — no platform detected`, and row 3 itself now names it: *"report it as `not created — the guide's Cut is unfinished`"*.

skills/release/SKILL.md
alexandrocuma: Fixed in 88816e0, essentially your wording, with the commands the guide gives:

> Where it gives none, the generic shape is **confirm it is still unpublished, then delete it** — a re-run of the publisher, by you or by the user in the platform's UI, can publish it while you are standing here:
> ```bash
> gh release view   "$tag" --json isDraft,isPrerelease   # must still show it unpublished
> gh release delete "$tag" --yes                         # never --cleanup-tag
> ```
> If it has since been published it is a normal release: do not delete it — that is the deletion a rollback forbids — roll forward instead.

Checked both fields exist (`gh release view v0.9.2 --json isDraft,isPrerelease` → `{"isDraft":false,"isPrerelease":false}`) and that `--cleanup-tag` is a real flag on `gh release delete`, which is why it's called out.

skills/release/SKILL.md
alexandrocuma: Fixed in 88816e0. The commit/tag block now binds the tag once and everything downstream uses it:

```bash
tag="v<version>"        # or the guide's format, e.g. <target>@<version>
…
git tag -a "$tag" -m "Release v<version>"
```

and the generic commands are `gh release create "$tag" --title "$tag" …` / `glab release create "$tag" --name "$tag" …`, with the prose saying they use the tag the cut actually pushed, never a literal `v<version>`. Your second point is in too: the `ls-remote` check is read-only and cheap, so it runs before any release-creation command, guide-provided included — *"row 1 forbids changing that command, not checking before it."* Walked scenario (d) on a fixture whose Cut names `api@<version>` and says nothing about a release object: row 2, tag `api@1.4.0`, both commands carry it.

skills/release/SKILL.md
alexandrocuma: Fixed in 88816e0, both sentences.

Phase 6: *"Name that target and **bind it for this run even if the change did not affect it — watch-only**: Phase 7 watches its pipeline and reads its **Post-release verification** signals for it, but runs none of its build or distribute commands and needs no target gate."* Phase 7's opening was widened to admit that target (watched and verified, never gated, built or distributed; its state recorded like any other target's).

Phase 8's gate now reads: *"The release is complete when the cut succeeded, the release object is published (or the user has explicitly accepted it staying unpublished), **and** every bound target is `shipped` or explicitly `skipped`."* So a docs-only release that leaves a draft behind no longer retires anything.

skills/release/SKILL.md
alexandrocuma: Both fixed in 88816e0.

(a) `notes="/tmp/.release-${ticket}-notes.md"` — the variable Phase 0 binds, so Phase 8's glob matches by construction. The comment is now *"retired with the scratch in Phase 8, when the release completes with a ticket id"*.

(b) The guard is there, with the `v`-prefix trap named:

```bash
[ -s "$notes" ] || { echo "no [<version>] section in CHANGELOG.md — stop"; exit 1; }
```

> `<version>` here is the bare version as the changelog heading writes it, **without** the `v` the tag carries — the commonest way to get an empty file, and an empty notes file ships a release with no notes.

Reproduced your numbers against this repo's CHANGELOG: `v0.9.2` → 0 bytes (guard stops), `0.9.2` → 15 070 bytes.

skills/release/SKILL.md
alexandrocuma: Added in 88816e0: `… / not created — the guide's Cut is unfinished (/devxp) / tag only — no platform detected`, and row 3 of the table names that outcome so the two can't drift.

docs/reference/skills.md
alexandrocuma: Fixed in 88816e0, your wording: *"creates a published release with the tag verified (`--verify-tag` on GitHub; an explicit `git ls-remote` check before `glab`, which has no such flag) and the changelog section passed as a notes **file**…"*

agents/gen-docs.md
alexandrocuma: Fixed in 88816e0: *"`/release` runs it as written, substituting only the version, tag and notes file, and never adding, dropping or reordering a flag; omit the line and `/release` creates a published release from the version's changelog section."* — which now matches row 1 of the skill's table word for word.

Applied to `agents/update-docs.md` in the same commit; re-diffed the two `#### Release Guide` blocks afterwards and they are still byte-identical (65 lines each).

Noted on this repo's guide saying `GitHub Release object:` — out of scope here, the coordinator is taking it separately.

agents/update-docs.md
alexandrocuma: Fixed in 88816e0: *"`/release` runs it as written, substituting only the version, tag and notes file, and never adding, dropping or reordering a flag; omit the line and `/release` creates a published release from the version's changelog section."* — which now matches row 1 of the skill's table word for word.

Applied to `agents/update-docs.md` in the same commit; re-diffed the two `#### Release Guide` blocks afterwards and they are still byte-identical (65 lines each).

Noted on this repo's guide saying `GitHub Release object:` — out of scope here, the coordinator is taking it separately.

skills/release/SKILL.md
```diff
@@ -255,7 +310,7 @@ mkdir -p ~/.claude/agent-memory/release
 
 ### Phase 8 — Retire Delivery Artifacts  *(only when the release is complete)*
 
-**Gate strictly on completion.** The release is complete when the cut succeeded **and** every bound target is `shipped` or explicitly `skipped`. If any step failed or was declined, or any target is `awaiting-external`, `blocked` or `failed`, skip this phase entirely and preserve everything — worktree, plan, scratch, release state — for the resumed run. Scope every action to this ticket; the repo-wide sweep belongs to `/improve` (C2) and `/cleanup`.
+**Gate strictly on completion.** The release is complete when the cut succeeded, the release object is published (or the user has explicitly accepted it staying unpublished), **and** every bound target is `shipped` or explicitly `skipped`. If any step failed or was declined, or any target is `awaiting-external`, `blocked` or `failed`, skip this phase entirely and preserve everything — worktree, plan, scratch, release state — for the resumed run. Scope every action to this ticket; the repo-wide sweep belongs to `/improve` (C2) and `/cleanup`.
```
alexandrocuma: 🟡 **Must fix — "or the user has explicitly accepted it staying unpublished" is an escape hatch with no prompt behind it.**

I grepped the whole skill: there is no prompt, no wording, and no phase that produces that acceptance. Every other gate in this file is a written prompt with a literal question (Phase 2's *"Confirm release?"*, 7a's *"Ship `<target>`?"*, 7c's *"Promote to `<next stage>`?"*, 7e's *"Run rollback?"*). This clause is the one thing standing between "the publisher never ran" and deleting the worktree, branch, plan, release state and scratch — and it is satisfiable by an agent deciding on its own that the user seemed fine with it.

Two clean ways out:

- **Drop the clause.** An unpublished release object means the watch-only target is `awaiting-external`, which already blocks this phase; `/release <ticket>` resumes and retires later. This is the smaller diff and matches *"Waiting is a state, not a failure."*
- **Or give it a prompt**, in the same shape as the others:
  ```
  The release object for v<version> on <platform> is still unpublished — <target> publishes it, and its pipeline <failed / has not run>.
  Retire this delivery's artifacts anyway? (yes / no — keep them for /release <ticket>)
  ```
  with "no" as the outcome on anything that isn't an explicit yes.

I'd take the first.

skills/release/SKILL.md
```diff
@@ -176,33 +179,85 @@ Then apply each affected target's **Versioning** rule from the guide — typical
...
 
+```bash
+gh release view   "$tag" --json isDraft,isPrerelease   # must still show it unpublished
+gh release delete "$tag" --yes                         # never --cleanup-tag
+```
+
+If it has since been published it is a normal release: do not delete it — that is the deletion a rollback forbids — roll forward instead. Either way, fix on a branch and cut the next patch version. The tag stays; tags are never moved, deleted or re-pushed.
+
+If the guide says a target's pipeline is **triggered by the tag** (CI builds and deploys on tag push), the push has already started that target's ship — Phase 7 then *watches* that pipeline instead of running its build/distribute commands.
 ---
```
alexandrocuma, line 259: 🟡 **Must fix — missing blank line: this `---` is now a setext heading underline, not the phase separator.**

`88816e0` removed the blank line that used to sit between the paragraph on line 255 and this `---`. In CommonMark, `---` directly under a paragraph line makes that line an `<h2>`. So:

- the Phase 6 / Phase 7 separator **disappears**, and
- *"If the guide says a target's pipeline is **triggered by the tag** (CI builds and deploys on tag push), the push has already started that target's ship…"* renders as a **heading**, immediately above `### Phase 7`.

I checked the rest of the file — this is the only place it happens (the other match is the frontmatter's closing `---`, as expected). A skill file's heading structure is how an agent finds a phase, so this isn't only cosmetic.

**Fix:** one blank line between line 255 and the `---`.

skills/release/SKILL.md
```diff
@@ -87,6 +87,8 @@ ls package.json go.mod pyproject.toml Cargo.toml version.go VERSION 2>/dev/null
 gh auth status >/dev/null 2>&1 && echo "platform: github" || { glab auth status >/dev/null 2>&1 && echo "platform: gitlab" || echo "platform: none — tag only"; }
 ```
```
alexandrocuma, line 89: 🔵 **Nit — the preflight classification has no command and no empty-output guard, so it can be ambiguous in exactly the way Phase 6 no longer is.**

This paragraph says *"read the guide's **Cut** section read-only and classify it against Phase 6's table"*, but the `sed` range and — more importantly — the *"empty output is not proof the guide is silent"* guard live only in Phase 6 (lines 187-190). An agent classifying here improvises an extraction, gets nothing out of a `## The Cut` heading, and prints `generic — Cut is silent` in the inventory and at the gate.

It isn't unsafe: Phase 6 re-runs the `sed` and re-applies the guard, so it still lands on row 3 and still stops before the commit. But the user will have been shown, and said yes to, a gate promising the generic commands, and then watched the cut stop anyway. Since the whole point of hoisting was to make the gate honest, it's worth closing.

**Suggested:** *"…read the guide's **Cut** section read-only with Phase 6's `sed` range **and its empty-output guard**, and classify it against Phase 6's table."* Or move the snippet and the guard paragraph up here and have Phase 6 reference them.

skills/release/SKILL.md
```diff
@@ -100,6 +102,7 @@ Release preflight — <ticket or branch>
   Current version: <vX.Y.Z> → proposed <vX.Y.Z> (<patch|minor|major>)
   Platform       : <github | gitlab | none — tag only>
   Release guide  : <docs/guides/release.md, last verified YYYY-MM-DD | missing — cut only>
+  Release object : <from the guide's Cut: `<command>` | generic — Cut is silent | blocked — Cut unfinished, run /devxp>
```
alexandrocuma: 🔵 **Nit — no value for "there is no guide at all".**

The three values are `from the guide's Cut: …`, `generic — Cut is silent`, and `blocked — Cut unfinished`. Phase 0's third row (no guide → cut-only) produces none of them: `generic — Cut is silent` is nearly right but names a Cut section that doesn't exist, and an agent that reads the new Phase 1 paragraph literally (*"read the guide's **Cut** section"*, unconditional) may instead reach for `blocked`, which would refuse a cut that Phase 0 just agreed to do cut-only.

**Suggested:** add a fourth value — `| generic — no release guide` — and one clause on line 89: *"With no guide this is `generic — no release guide`; Phase 0's cut-only offer already covered it."*

skills/release/SKILL.md
```diff
@@ -176,33 +179,85 @@ Then apply each affected target's **Versioning** rule from the guide — typical
...
 
-Publish on the platform detected in Phase 1:
+**Tag push is the point of no return for the cut.** Everything before it is local and revertible; everything after is public. If a step here fails, stop — do not proceed to Phase 7, and do not retry a push that may have partially succeeded without checking `git ls-remote --tags origin` first.
+
+The release notes are always the version's own `CHANGELOG.md` section, passed as a **file**. A multiline body inlined into `--notes` is at the mercy of the shell:
+
+```bash
+notes="/tmp/.release-${ticket}-notes.md"   # retired with the scratch in Phase 8, when the release completes with a ticket id
+awk -v v="<version>" '$0 ~ "^## \\[" v "\\]" {f=1; next} f && /^## \[/ {exit} f' CHANGELOG.md > "$notes"
+[ -s "$notes" ] || { echo "no [<version>] section in CHANGELOG.md — stop"; exit 1; }
```
alexandrocuma: 🔵 **Nit — this is now the only stop left that fires after the tag push, and it doesn't need to be.**

Everything else that can stop the cut was moved in front of the commit in this commit. But `[ -s "$notes" ]` reads `CHANGELOG.md`, which Phase 4 finalised — it depends on nothing the commit or the push produces. When it fires, the result is the shape the commit message says it fixed for row 3: a public tag that may already have started a pipeline, and no release object.

**Suggested:** move the whole notes block (lines 214-218) above the "Commit and tag" block, next to the decision table, and leave a one-line pointer here. Then *every* refusal in Phase 6 is pre-push, which is a much easier rule to hold under pressure than "these three are pre-push and that one isn't".

skills/release/SKILL.md
```diff
@@ -176,33 +179,85 @@ Then apply each affected target's **Versioning** rule from the guide — typical
...
+
+If it has since been published it is a normal release: do not delete it — that is the deletion a rollback forbids — roll forward instead. Either way, fix on a branch and cut the next patch version. The tag stays; tags are never moved, deleted or re-pushed.
+
+If the guide says a target's pipeline is **triggered by the tag** (CI builds and deploys on tag push), the push has already started that target's ship — Phase 7 then *watches* that pipeline instead of running its build/distribute commands.
 ---
 
 ### Phase 7 — Ship Targets  *(each target gated on its own; skipped when cut-only)*
 
-Ship the targets bound in Phase 0, in the order the guide lists them (the guide or groom plan may require an order — e.g. a backend before the app that depends on it). Blocked targets are skipped and reported.
+Ship the targets bound in Phase 0 — plus any publisher target Phase 6 bound **watch-only**, which is watched and verified but never gated, built or distributed, and whose state is recorded like any other target's — in the order the guide lists them (the guide or groom plan may require an order — e.g. a backend before the app that depends on it). Blocked targets are skipped and reported.
```
alexandrocuma, line 263: 🔵 **Nit — a watch-only target is indistinguishable from a shipped one downstream.**

Two small gaps now that the binding exists:

1. **Phase 9's table** (line 363 onward) gives it an ordinary row — `Build | —`, a channel, `shipped` — reading exactly like a target this run built and distributed. Worth a marker: `shipped (watch-only — published the release object)`, or a `watch-only` note in the Verification column.
2. **Phase 2's gate** lists *"Then ship, each target gated on its own"* from the Phase 0 binding, but this target is bound in Phase 6, i.e. after the gate — so the list the user consented to silently grows by one. It's ungated and watch-only, so nothing runs without consent, but the natural place to disclose it is the preflight line you just added: `Release object : … — published by <target>`.

skills/release/SKILL.md
```diff
@@ -176,33 +179,85 @@ Then apply each affected target's **Versioning** rule from the guide — typical
...
 
-If the guide says a target's pipeline is **triggered by the tag** (CI builds and deploys on tag push), the push has already started that target's ship — Phase 7 then *watches* that pipeline instead of running its build/distribute commands.
+**If that publisher fails or never runs**, follow the guide's steps for the target. Where it gives none, the generic shape is **confirm it is still unpublished, then delete it** — a re-run of the publisher, by you or by the user in the platform's UI, can publish it while you are standing here:
 
+```bash
+gh release view   "$tag" --json isDraft,isPrerelease   # must still show it unpublished
```
alexandrocuma: 🔵 **Nit — the "generic shape" recovery is `gh`-only.** The paragraph introduces it as what to do *where the guide gives no steps*, which is most likely in a repo whose guide is thin — including a GitLab one. `glab release delete <tag>` (and `glab release view`) have no counterpart here, and `glab release delete` takes `--with-tag`, the flag to keep away from for the same reason as `--cleanup-tag`.

**Suggested:** either add the GitLab line beside the two `gh` ones, or say "on GitHub; the equivalent `glab release view` / `glab release delete` on GitLab — never `--with-tag`".

alexandrocuma (commented):
## Re-review @ `88816e0`

**All 12 findings from the first review are fixed**, and the reordering is a genuine improvement rather than a patch: the stop-the-cut branch now fires at preflight, `Release object :` makes the classification visible to the user before the gate, and `tag=` threads the guide's tag format through every command that touches it. I re-ran each failure shape and re-walked A–F.

**Recommendation: one more small commit, then merge.** Two of the new lines need a fix — a Markdown structure break and a safety-gate escape hatch with nothing behind it. Everything else below is a nit that can ride along.

### Verification of the 12

| # | Finding | Status |
|---|---|---|
| 1 | GitLab `ls-remote` failed open | ✅ `\| grep -q .` + explicit stop, hoisted so it also guards a row-1 command — and the "row 1 forbids changing that command, not checking before it" framing is the right reading |
| 2 | "stop the cut" after the point of no return | ✅ table hoisted above the commit; Phase 1 classifies; Phase 2 prints the resolved command. Row 3 now even names the report string it produces |
| 3 | `sed` empty ⇒ silently row 2 | ✅ guard + `grep -niE` confirm (see nit 3 — the guard didn't travel to Phase 1) |
| 4 | unbounded delete | ✅ fresh `isDraft,isPrerelease`, no `--cleanup-tag`, published ⇒ roll forward |
| 5 | generic commands hardcoded `v<version>` | ✅ `tag=` used by `git tag`, `ls-remote`, `gh` and `glab` |
| 6 | publisher target unbound | ✅ watch-only binding + Phase 7 clause + Phase 8 gate (see must-fix 1 and nit 5) |
| 7 | Phase 9 missing outcome | ✅ |
| 8 | notes file | ✅ `${ticket}`, `[ -s ]`, bare-version note, comment no longer overpromises (see nit 4) |
| 9 | `--verify-tag` GitHub-only | ✅ |
| 10 | "verbatim" in both docs agents | ✅ — I re-extracted both `#### Release Guide` blocks at `88816e0` and `diff`ed: still **byte-identical**, 64 lines |

Also re-checked: `./install.sh --dry-run` ends `All done.`; `devexp/internal/skills` and `internal/assets` pass; the CHANGELOG addition is accurate (the `ls-remote`-exits-0 rationale and the non-empty-section requirement both match the file).

### Paths A–F at `88816e0`

- **(a) this repo** — preflight prints `Release object : from the guide's Cut: \`gh release create v<version> --draft …\``, the gate shows it, Phase 6 runs exactly it with `$notes` substituted, binds `cli`, Phase 7 watches `release.yml` for `isDraft:false`. ✅
- **(b) no guide** — cut-only; `sed` empty, `grep` finds nothing, row 2, generic with `$tag`. ✅ except the inventory line has no value for this case (nit 2).
- **(c) half-written / `[CONFIRM]`** — now caught in preflight, gate never offered, **nothing pushed**. This is the fix that mattered. ✅
- **(d) monorepo tag format** — `tag="<target>@<version>"` flows into `git tag`, the remote check and both create commands. ✅ (the tag *message* is still a literal `Release v<version>` in the block while the prose says to take the guide's format — the `tag=` line got an inline hint, this one could too)
- **(e) GitLab** — the guard stops on empty output before `glab release create "$tag"`; no `--verify-tag` claimed anywhere. ✅ (nit 6)
- **(f) no platform** — line 242 still reached; the notes file and the remote check that precede it are read-only/local, so nothing is wasted or harmful. ✅

### What I'd still change

Two must-fix (inline), four nits (inline). None is a safety regression; the worst case of must-fix 1 is retiring a delivery's artifacts while the release object is still a draft — the scenario Phase 6 was rewritten to prevent — via a clause with no prompt behind it.

---

*Re-reviewed read-only at `88816e0`.*

skills/release/SKILL.md
alexandrocuma: 🔵 **Nit (from the same re-review pass) — the "generic shape" recovery is `gh`-only.** The paragraph above introduces it as what to do *where the guide gives no steps*, which is likeliest in a repo whose guide is thin — including a GitLab one. There is no `glab release view` / `glab release delete` counterpart here, and `glab release delete` has `--with-tag`, the flag to keep away from for the same reason as `--cleanup-tag`.

**Suggested:** add the GitLab lines beside the two `gh` ones, or say "on GitHub; the equivalent `glab release view` / `glab release delete` on GitLab — never `--with-tag`".

skills/release/SKILL.md
alexandrocuma: Took the first option — fixed in 074ee1e. The clause is gone, and the gate now says why it doesn't need one:

> The release is complete when the cut succeeded, the release object is published, **and** every bound target is `shipped` or explicitly `skipped`. A release object still unpublished leaves its watch-only target `awaiting-external`, which already holds this phase — `/release <ticket>` retires the artifacts once the publisher has run.

Smaller diff, and it lands on "Waiting is a state, not a failure" instead of inventing a gate that no phase prompts for.

skills/release/SKILL.md
alexandrocuma: Fixed in 074ee1e — blank line restored, so the `---` is the phase separator again and that paragraph is a paragraph.

I also scanned the whole file for the shape (`---`/`===` directly under a non-empty, non-table, non-fence line): the only remaining match is the frontmatter's close under `argument-hint:`, exactly as you found.

skills/release/SKILL.md
alexandrocuma: Fixed in 074ee1e, with your first suggestion — the preflight now names both the range and the guard, and says why:

> **The cut itself**, once, and only where a guide exists: extract its **Cut** section read-only with Phase 6's `sed` range, apply Phase 6's empty-output guard (an unmatched heading is not a silent guide), and classify the result against Phase 6's table — improvising an extraction here is how the gate ends up promising the generic commands that Phase 6 then refuses to run.

Kept the snippet itself in Phase 6 so there's one copy to maintain.

skills/release/SKILL.md
alexandrocuma: Fixed in 074ee1e. The inventory line now reads:

```
  Release object : <from the guide's Cut: `<command>` (published by <target>) | generic — Cut is silent | generic — no release guide | blocked — Cut unfinished, run /devxp>
```

and the Phase 1 paragraph is explicitly conditional — *"once, and only where a guide exists"* … *"With no guide at all this is `generic — no release guide`; Phase 0's cut-only offer already covered it."*

skills/release/SKILL.md
alexandrocuma: Agreed, and fixed in 074ee1e — the whole notes block moved above the commit rather than leaving a pointer, since it reads nothing the commit produces:

> Build the release notes next, still before anything is pushed. They are always the version's own `CHANGELOG.md` section — which Phase 4 already finalised, so nothing here needs the commit — passed as a **file** …

and the paragraph after it ends with the rule you wanted stated once: *"Every refusal in this phase is now pre-push; keep it that way."* Order is now: classify → notes + `[ -s ]` → commit/tag/push → `ls-remote` check → create.

skills/release/SKILL.md
alexandrocuma: Both fixed in 074ee1e.

1. Phase 9's example table gained a marked row: `| cli | — | published release | shipped (watch-only — published it) | not a draft, 5 assets |` (re-checked the column widths — every row is still 111 chars).
2. The preflight line discloses the publisher: `from the guide's Cut: `<command>` (published by <target>)`, and Phase 1 says why it's named there — *"name it here, because Phase 6 binds it watch-only after the gate"*. So the gate's list matches what ends up bound.

skills/release/SKILL.md
alexandrocuma: Fixed in 074ee1e — added the GitLab half rather than a parenthetical, so the recovery is symmetric:

```bash
gh   release view   "$tag" --json isDraft,isPrerelease   # GitHub — must still show it unpublished
gh   release delete "$tag" --yes                         # never --cleanup-tag
glab release view   "$tag"                               # GitLab — same check, read the release's state
glab release delete "$tag" --yes                         # never --with-tag
```

Checked against the installed `glab`: `release delete` has `-t/--with-tag` and `-y/--yes`, and `release view [<tag>]` exists.

alexandrocuma: Fixed in 074ee1e — added the GitLab half rather than a parenthetical, so the recovery is symmetric:

```bash
gh   release view   "$tag" --json isDraft,isPrerelease   # GitHub — must still show it unpublished
gh   release delete "$tag" --yes                         # never --cleanup-tag
glab release view   "$tag"                               # GitLab — same check, read the release's state
glab release delete "$tag" --yes                         # never --with-tag
```

Checked against the installed `glab`: `release delete` has `-t/--with-tag` and `-y/--yes`, and `release view [<tag>]` exists.

alexandrocuma:
Re-review items are all in `074ee1e`, answered on their threads.

One extra commit, `4b957ba`: #155 landed on `main` while this was in review and the PR went `CONFLICTING`. The only conflict was `CHANGELOG.md`'s `[Unreleased]` block — both sides opened a `### Fixed` section — so main's entry is kept and #154's sits under the same heading. Nothing else overlapped: this branch touches the skill, the two docs agents and three docs; #155 touched the workflows, `scripts/govulncheck.sh` and the release guide's Build section (its Cut bullet, which this PR reads, is unchanged and still resolves to row 1).

All suites re-run on the merge result, and CI is green on `4b957ba` (`test`, `hooks`, `govulncheck`). PR is `MERGEABLE` again. Not merging — leaving that to you.

Issue #162, opened 2026-09-17
Security guards fail open when a scan is slow: add a fail-closed scan budget


alexandrocuma:
## Problem
Claude Code doesn't block a tool call when a `PreToolUse` command hook times out, and the default timeout is long. `hooks/registry.json` sets no timeout for devexp's security guards. So a guard that is slow on some input fails **open**: the tool call goes ahead unscanned after a long wait.

The opencode modules run synchronously, so a slow match freezes opencode instead.

Found in the PR #158 review. Slow patterns have come up in both content-scanning guards (#146, PR #158). This ticket is about the missing budget itself, whatever pattern is slow next.

## Acceptance Criteria
- [ ] Each security guard (`fail_closed` in the registry) enforces its own scan budget and **blocks** with a clear message when the budget is exceeded, in both the Claude Code hook and the opencode module.
- [ ] The registry sets an explicit hook `timeout` above that budget, so the guard's own block always wins over Claude Code's timeout.
- [ ] Tests: an input forced over budget (via a test seam or a deliberately slow pattern) blocks in both implementations; normal inputs are unaffected.
- [ ] `docs/reference/hooks.md` documents the budget and the fail-closed behaviour.

## Non-Goals
- Making every pattern linear (tracked per guard).

alexandrocuma:
## Groomed — plan

Validated against `8aa2d7a`. Every claim in the ticket holds.

### What the docs actually say

- A timed-out **command** hook does not block the tool call: *"A timed-out `command`, `http`, or `mcp_tool` hook doesn't block the tool call. The call continues through the normal permission flow, so don't count on a stalled hook to act as a gate."* So the premise is correct.
- `timeout` is in **seconds**, and the default for a `command` hook on `PreToolUse` is **600**. That is what makes a slow guard so expensive today.
- There is **no** field that expresses "block on timeout" — no `failClosed`, `blockOnTimeout` or equivalent. The only lever is the guard exiting 2 before the timeout fires.
- Trap for the next reader: the Agent SDK *callback* hooks page says a callback that exceeds its timeout **does** block. That is a different mechanism (in-process SDK callbacks) and does not apply to our command hooks.
[...]

PR #166, opened 2026-09-17
ci: scan weekly and gate the tagged release on the full suite (#155)

### LINKED ISSUE #155: CI: scheduled vulnerability scan, test-gated release workflow, stale CHANGELOG line refs

alexandrocuma:
Closes #155

## What changed
1. **Weekly vulnerability scan, plus a manual trigger.** `ci.yml` gains `schedule: "27 6 * * 1"` (Mondays 06:27 UTC — off the hour, which GitHub documents as when scheduled runs get delayed or dropped), so an advisory published while `main` is quiet is caught within a week instead of at the next push. A failure shows as a red `ci` run on `main`. It also gains `workflow_dispatch:` (added in review), because a public repo's schedules are disabled after 60 days without activity and a schedule that never fires raises no alarm — the manual trigger is how you confirm the scan works again, or scan a new advisory without waiting for Monday.
2. **A tagged release now runs the full suite first.** `ci.yml` gains `workflow_call`, and `release.yml` replaces its copy of the govulncheck job with `uses: ./.github/workflows/ci.yml` plus `needs: ci` on goreleaser. A local `uses:` resolves at the caller's ref, so the tagged commit is what gets tested. `goreleaser` stays the only job with `contents: write`; the scan keeps `contents: read` and `persist-credentials: false`.
3. **Stale references.** No doc cites a `CHANGELOG.md` line number any more (`grep -rn "CHANGELOG\.md:[0-9]"` is empty). Workflow citations name triggers and jobs instead of line numbers. Ten drifted Go-test citations were corrected by dropping the line number, since each already names its symbol.

Job graph before → after:
```
before  release.yml: govulncheck (a copy of ci.yml's) → goreleaser
after   release.yml: ci (uses ./.github/workflows/ci.yml: test | hooks | govulncheck) → goreleaser
```

## Verification

**The `workflow_call` path was executed on this PR, not just linted.** A temporary workflow (`zz-call-check.yml`) called `ci.yml` exactly as `release.yml` does — same job id `ci`, `permissions: contents: read`, no `with:` or `secrets:` — plus a `gate` job with a plain `needs: ci` standing in for `goreleaser`. Both the pass and the fail case were observed, then the scaffolding was removed.

| Case | Run | Result |
|---|---|---|
| **Pass** — call succeeds | [35287978159](https://github.com/alexandrocuma/devexp-toolkit/actions/runs/35287978159) (`2d99324`) | `ci / test`, `ci / hooks`, `ci / govulncheck` all success → run **success** |
| **Fail** — one inner job fails | [35288156553](https://github.com/alexandrocuma/devexp-toolkit/actions/runs/35288156553) (`f13d164`) | `ci / test` **failure**, `ci / hooks` and `ci / govulncheck` success → call job aggregates to **failure** (`needs.ci.result = failure` in the observer log), and `gate` — the goreleaser stand-in — is **skipped with 0 steps** |
| **Clean** — scaffolding removed | [35288368129](https://github.com/alexandrocuma/devexp-toolkit/actions/runs/35288368129) (`de48c95`) | `test`, `hooks`, `govulncheck` all pass; no `zz-call-check` run |

So on a tag: one failing test ⇒ the `ci` call job fails ⇒ `goreleaser` never starts ⇒ nothing is published. That is the gate, observed rather than reasoned about. These runs outlive the deleted scaffolding.

The three scaffolding commits (`2d99324`, `f13d164`, `de48c95`) are in the branch history but cancel out: `git diff df5a86c de48c95` is empty, and only `ci.yml` and `release.yml` remain under `.github/workflows/`.

Also checked:
- **actionlint 1.7.12** (downloaded to scratch, SHA-256 verified): clean on both files and on whole-repo discovery, before and after — including with all five `on:` triggers.
- **Job-graph assertions** parsed from the YAML — triggers, cron, per-job permissions, `uses`, no `steps`/`secrets` on the call job, `needs: ci`, `workflow_dispatch` and `workflow_call` coexisting, and that the jobs gating goreleaser are exactly `{test, hooks, govulncheck}`.
- **Cron:** next 5 firings enumerated, 7 days apart.
- All suites pass locally: Go (12 packages), every hook suite, and the three installer script suites.

## Review fixes (`96694e5`)

All six findings addressed; replies on each thread. Two are worth surfacing here:

- **The alerting design was documented wrong.** `testing.md` said GitHub emails "the workflow's author" on a scheduled-run failure. Since this change deliberately files no issue, that sentence *is* the alert. The docs actually specify three rules: the recipient defaults to the workflow's **creator**, moves to whoever **last modified the `cron` line**, and — overriding both — moves to whoever **re-enabled** the workflow if it was ever disabled. That last rule fires on exactly the 60-day-disable-then-re-enable sequence described two sentences later, so the rules now sit together, along with the fact that the previous recipient is never told they stopped being notified. 60 days is exact and scoped to public repositories.
- **`gh run rerun --failed` is now marked unverified on this path.** The failed job lives inside a called workflow, which is a different and historically rough re-run path. The guide keeps it as the first thing to try, adds re-running the whole run as the fallback and cutting the next patch after that, and asks for the bullet to be stamped with the tag that first confirms it. Verifying it honestly needs a real release, which this PR is not going to do.

Also: `scripts/govulncheck.sh`'s header no longer claims `release.yml` runs it; the copied-test-helper counts are four and two, not two and one; and `govulncheck`'s per-job `permissions` block is kept but both comments now explain that it restates the workflow default deliberately, rather than implying it grants something extra.

## Notes for reviewers
- **Check names — confirmed by observation, not inference.** Direct runs of `ci.yml` (push/PR) keep the plain names `test`, `hooks`, `govulncheck` — unchanged by this PR, as this PR's own checks show. When `ci.yml` is *called*, its jobs render as `<caller-job-id> / <called-job-id>`, i.e. `ci / test`, `ci / hooks`, `ci / govulncheck` on a tag, since `release.yml`'s call job is named `ci`. **The call job itself does not appear as its own check** — only the three nested names do, so anything matching on check names must match those three, not `ci`. No required-check names exist to update today: `main` has no branch protection (`/branches/main/protection` → 404) and no rulesets (`/rulesets` → `[]`). If they are ever added, they need both spellings.
- **At merge:** this repo has `squash_merge_commit_message=COMMIT_MESSAGES`, so the squash body is pre-filled by concatenating every branch commit message — including "ci: temporarily fail one test to observe the release gate skip", which reads badly in `main`'s log out of context. Trim the squash body by hand. (Handled by the merger; the branch history is deliberately left alone.)

- ~70–78 `<file>:<line>` citations remain live in `docs/` (count varies by regex). The drift is in-range-but-wrong, so only reading the cited line finds it: worth a CI guard or a convention change — this is now the second sweep to find real drift.
- The `Last verified … against commit <sha>` stamps in the kit docs are stale; they're left alone while #154 and #162 edit the same files.
- `ci.yml` has no `concurrency:` group, so a rapid push sequence runs overlapping full suites. Pre-existing, but `schedule` adds one more source of runs.


docs/architecture/overview.md
```diff
@@ -167,8 +167,8 @@ Hook commands point into the install root, so editing a registered script in the
 | `cli/internal/` | One package per asset kind or concern (see Layers) |
 | `cli/internal/assets/` | `assets.go` (`//go:embed all:agents all:skills all:hooks all:mcps devexp.config.json uninstall.sh`). The staged copies next to it are gitignored and produced by `scripts/stage-assets.sh`, and the package doesn't compile until they exist |
 | `cli/internal/agents/testdata/` | Agent fixtures for transformation tests |
-| `scripts/` | `stage-assets.sh` (rsync of assets into `cli/internal/assets/`), `remote-install.sh`, `govulncheck.sh` (govulncheck for every release platform, used by `ci.yml` and `release.yml`) |
-| `.github/workflows/` | `ci.yml` (Go tests, hook tests, installer script tests, govulncheck), `release.yml` (tag → govulncheck → goreleaser) |
+| `scripts/` | `stage-assets.sh` (rsync of assets into `cli/internal/assets/`), `remote-install.sh`, `govulncheck.sh` (govulncheck for every release platform, run by `ci.yml`'s `govulncheck` job — which `release.yml` reaches by calling `ci.yml`) |
```
alexandrocuma, line 170: 🟡 **This line is now right; its counterpart in the script itself was missed.**

You correctly rewrote this to "run by `ci.yml`'s `govulncheck` job — which `release.yml` reaches by calling `ci.yml`". But `scripts/govulncheck.sh:4` still says:

```
# Used by .github/workflows/ci.yml, .github/workflows/release.yml and locally.
```

**Consequence.** `release.yml` no longer runs this script at all. It matters more than an ordinary stale comment because `docs/development/testing.md:73` calls this script "the one place the scan is defined" — so its header is exactly where a reader goes to find out who runs it, and it now points them at a `govulncheck` step in `release.yml` that doesn't exist. The next person editing `release.yml` will go looking for it.

**Fix** (`scripts/govulncheck.sh:4`):

```sh
# Used by .github/workflows/ci.yml (job `govulncheck`), which release.yml calls
# on a tag, and locally.
```

The file isn't in this diff, but this diff is what invalidated it — same sweep, same fact, one line. I checked the rest of the repo (`CLAUDE.md`, `README.md`, `docs/README.md`, `skills/`, `agents/`, `hooks/`, `cli/`) and this is the only other place that still describes the old job graph.

docs/development/testing.md
```diff
@@ -66,13 +66,15 @@ Gotchas when running tests:
...
+There is one `govulncheck` job, in `ci.yml`. It sets up Go from `cli/go.mod` (so the go1.26.8 standard library the release is built with is scanned), runs the script with a read-only token and `persist-credentials: false`, and runs wherever `ci` runs: pull requests, pushes to `main`, the weekly schedule, and the tag, where `release.yml` calls `ci.yml` and goreleaser `needs:` the call — so a failed scan publishes nothing (see [`../guides/release.md`](../guides/release.md#build)).
+
+The weekly run exists because an advisory published while `main` is quiet would otherwise wait for the next push, PR or tag. It runs the whole `ci` workflow, not a scan-only copy of the job, and a failure is a red `ci` run on `main` in the Actions tab (GitHub also emails the workflow's author on a scheduled-run failure). It deliberately files no issue: that would need a write-scoped token on a read-only workflow plus dedup so a standing advisory doesn't file one issue a week. Scheduled workflows fire only from the default branch, and GitHub disables them after ~60 days without repository activity — no run, and so no failure, is the one way this can go quiet.
```
alexandrocuma: 🟡 **"GitHub also emails the workflow's author on a scheduled-run failure" is not what GitHub does.**

Per the current *Events that trigger workflows* reference:

> Notifications for scheduled workflows are sent to the user who last modified the cron syntax in the workflow file.

Not the workflow's author — **whoever last touched the `cron` line**.

**Consequence.** This paragraph is deliberate about filing no issue ("that would need a write-scoped token on a read-only workflow"), which makes this sentence the *entire* alerting design for the weekly scan. As written it promises a stable recipient. In reality the alerts hand off silently the first time anyone edits the cron — changing the day, the minute, or reformatting the line — and the previous recipient gets no signal that they've stopped being notified. That is precisely the "this can go quiet" mode the next sentence is trying to guard against. It's also contingent on that user having Actions notifications enabled at all.

**Suggested fix:**

> …and a failure is a red `ci` run on `main` in the Actions tab. GitHub emails the user who **last modified the `cron` line** (not necessarily its author) when a scheduled run fails, if they have Actions notifications enabled — so whoever edits the schedule inherits the alerts.

Two smaller things in the same sentence, worth folding in:

- "GitHub disables them after ~60 days" — the docs scope this to public repos: *"In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days."* True here (the repo is public) and it's exactly 60, not ~60. Naming the public-repo condition makes the caveat survive a visibility change.
- ✅ "Scheduled workflows fire only from the default branch" is exactly right (*"Scheduled workflows run on the latest commit on the default branch"*), and `27 6 * * 1` deliberately avoids the top of the hour, which the same page flags as when scheduled runs get delayed or dropped. Nice.

alexandrocuma (commented):
## PR Review — `ci: scan weekly and gate the tagged release on the full suite` (#155)

**Recommendation: Request changes** — two accuracy defects, one of them created by this PR. The design is right and I confirmed every load-bearing claim about it; neither finding touches the workflow YAML, which I'd merge as-is.

**Reviewed at `de48c95`.** I caught the branch mid-experiment at `f13d164` (both `zz-call-check.yml` and a `t.Fatalf` probe in `cli/internal/manifest/`), and `de48c95` removes both — `git diff df5a86c de48c95` is empty, so the final tree is byte-identical to the substantive state, and `git ls-tree` confirms no `zz*` survives. Everything below was verified against that tree.

The temporary caller was worth doing: it turned "the `workflow_call` path is validated statically" into an observed result. I checked all three runs rather than taking the summary on trust:

| Run | Workflow | Result |
|---|---|---|
| [35287978159](https://github.com/alexandrocuma/devexp-toolkit/actions/runs/35287978159) | `zz-call-check` | success — `ci / test`, `ci / hooks`, `ci / govulncheck` all green through the call |
| [35288156553](https://github.com/alexandrocuma/devexp-toolkit/actions/runs/35288156553) | `zz-call-check` | `ci / test` **failure** (other two green) → `gate-observer` printed `needs.ci.result = failure` → `gate` **skipped, steps=0** |
| [35288368129](https://github.com/alexandrocuma/devexp-toolkit/actions/runs/35288368129) | `ci` | success on `de48c95` itself |

`gate` reaching `skipped` with zero steps is the whole point: on a tag that job is `goreleaser`, and it never starts, so nothing is uploaded and no draft is published. These runs survive the scaffolding's deletion — worth linking them from the PR body so the evidence outlives the branch.

**Check names — independently confirmed, since the PR body now asserts it.** Through the call the jobs report as `ci / test`, `ci / hooks`, `ci / govulncheck`; the `ci` call job itself never renders as a check of its own (the run has no job named `ci`). On a direct PR run the names stay bare — `gh api .../commits/de48c95/check-runs` returns exactly `test`, `hooks`, `govulncheck`. So the "Notes for reviewers" warning is accurate, and any future required-check config has to list both spellings.

**The three scaffolding commits left in branch history: harmless for the tree, but they will leak into `main`'s log.** This repo squash-merges (the last four commits on `main` are single-parent, `(#NN)`-suffixed), so `2d99324`, `f13d164` and `de48c95` collapse into one commit whose tree is `de48c95`'s — the probe files never reach `main`, and the deliberately-red run stays a PR run, invisible to `gh run list --workflow ci.yml --branch main` (what `release.md`'s post-release check uses). The one catch is the message: the repo has `squash_merge_commit_message=COMMIT_MESSAGES`, so GitHub pre-fills the squash body by concatenating **every** branch commit message. Merged without editing that box, `main`'s history permanently records "ci: temporarily fail one test to observe the release gate skip" as part of shipping #155, which reads like the release gate was knowingly broken. **Trim the body at merge time** — it's a one-time manual step, not a code change.

---

### 🟡 Important

**1. `scripts/govulncheck.sh:4` now states something false — and this PR is what made it false.**

```
# Used by .github/workflows/ci.yml, .github/workflows/release.yml and locally.
```

`release.yml` no longer runs this script; it reaches it only by calling `ci.yml`. You fixed the exact counterpart claim in `docs/architecture/overview.md:170` in this same sweep, so this reads as an oversight rather than a scoping call. It matters more than a normal stale comment because `docs/development/testing.md:73` calls this script "the one place the scan is defined" — its header is where a reader goes to learn who runs it, and it now sends them hunting for a `govulncheck` step in `release.yml` that doesn't exist. Inline, with a suggested rewrite.

**2. `docs/development/testing.md:77` names the wrong notification recipient.** GitHub notifies "the user who last modified the cron syntax", not the workflow's author. Since the PR deliberately files no issue, this sentence *is* the alerting design for the weekly scan — and it describes a recipient that hands off silently the first time anyone edits the cron line. Inline, with the doc quote and a suggested rewrite.

### 🔵 Minor

3. `ci.yml` gains a `schedule:` but no `workflow_dispatch:` — one line, in the block this diff rewrites, and the PR's own docs argue for it.
4. `docs/guides/release.md:45` documents a `gh run rerun --failed` recovery that is now a reusable-workflow re-run — a different, historically buggy code path, asserted without being exercised.
5. `docs/development/testing.md:45` — `captureStdout` is defined in four test files, not two.
6. `ci.yml:62-65`'s per-job `permissions: contents: read` is now redundant with the new workflow-level block.

---

### ✅ What's done well

- **The gate is proven, not assumed** — see the run above.
- **Every GitHub Actions claim in the diff checks out** against current docs: local `uses:` ref resolution, permissions direction, and no need for `secrets: inherit` (table below).
- **No double-run, and no way past the gate.** `ci.yml`'s `push` is branch-filtered, so a tag push starts exactly one `ci` — the called one. `release.yml` has no `workflow_dispatch`, `goreleaser` carries no `if:`, and `contents: write` is still on `goreleaser` alone while the scan keeps `persist-credentials: false`. (The duplicate `test`/`hooks`/`govulncheck` that appeared on this PR was the temporary caller, now gone.)
- **The re-run of the suite on the tagged commit is justified, not accidental** — `release.md:41` gives the real reason (the tag can be pushed while `ci` is still running on the release commit, since `main` has no protection). Three re-run jobs to close that window is a good trade, and worth the explicit sentence.
- **`release.md:46-49` already handles the draft-left-behind case correctly.** With `use_existing_draft: true`, a skipped `goreleaser` leaves the `/release` draft unpublished and asset-less; the guide says delete the draft, keep the tag, cut the next patch — and the post-release table checks `isDraft`. The residual risk (a human publishing that empty draft by hand) is inherent to the draft flow, not introduced here.
- **This change makes the draft race safer.** `/release` creates the draft around the tag push; lengthening the pre-`goreleaser` path from one job to three widens the window in which the draft is guaranteed to exist before goreleaser looks for it.
- **Cron hygiene.** `27 6 * * 1` avoids the top of the hour, which GitHub documents as when scheduled runs get delayed or dropped.
- **The doc sweep found real drift.** Every Go-test line number it dropped was genuinely wrong; every one it kept is right.

---

### What I verified

**actionlint 1.7.12** — downloaded to scratch, SHA-256 `aba9ced2dee8d27fecca3dc7feb1a7f9a52caefa1eb46f3271ea66b6e0e6953f` matched the published `actionlint_1.7.12_checksums.txt`. Clean (exit 0) on `ci.yml` + `release.yml` explicitly, on whole-repo discovery, and on `main` as a baseline.

**All four suites** — Go `go test ./... -race -cover` exit 0 across 12 packages; all `hooks/claude-code/*.test.sh` pass; all `hooks/opencode/*.test.js` pass; all three root `*.test.sh` pass.

**GitHub Actions semantics, against current docs:**

| Claim | Verdict |
|---|---|
| A local `uses: ./…` resolves at the caller's ref | ✅ "the called workflow is from the same commit as the caller workflow" — so a tag push tests the tagged commit's `ci.yml` against the tagged code |
| Called workflow needs no `secrets: inherit` here | ✅ "The called workflow is automatically granted access to `github.token` and `secrets.GITHUB_TOKEN`". `setup-go`'s cache uses the Actions cache service (not `GITHUB_TOKEN`), and `go install …/govulncheck` goes to the public module proxy — no user-defined secret is touched |
| `ci.yml`'s comment "A caller can only lower these" | ✅ Accurate in effect — "permissions passed from the caller can be only downgraded (not elevated) by the called workflow", and `contents: read` on both sides caps it either way |
| `needs: ci` skips `goreleaser` on any inner failure | ✅ **Observed** — run 35288156553, `needs.ci.result = failure`, `gate` skipped |
| Check names become `ci / test` etc. on a tag | ✅ Observed in the same run — your "Notes for reviewers" is correct |
| Scheduled runs: UTC, default branch only | ✅ "Scheduled workflows run on the latest commit on the default branch" |
| ~60-day auto-disable | ✅ "In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days" — this repo is public, so it applies |

**Doc citations:**

- `grep -rn 'CHANGELOG\.md:[0-9]'` across the whole repo → **empty**. Claim confirmed.
- **No new `<file>:<line>` introduced.** Eight appear on added lines, all carried over from surrounding prose; I read every one and all are accurate: `.goreleaser.yaml:25` (the `-X …version` ldflag), `:5-7` (the stage-assets before-hook), `:44-49`/`44-52` (the `release:` block, `use_existing_draft` exactly at `:49`), `cli/go.mod:3-5`, `cli/cmd/root.go:15` and `:9-15`, `cli/cmd/install.go:78`, `cli/cmd/uninstall.go:80`, `cli/internal/repo/repo.go:412` (`var userCacheDir`), `scripts/remote-install.sh:50`/`:51-55`/`:53`, `install.sh:7-20`.
- **The dropped Go-test line numbers were all genuinely drifted** — `writeRegistry` cited `:774`, actually `cmd/install_test.go:1378`; `captureStdout` cited `:615`, actually `:1155`; `fakeCLI` cited `:700-710`, actually `:1240`; `blockedDir` cited `:839`, actually `:1443`; `TestAddClaude_NonExec` cited `mcp_test.go:201`, actually `:202`. Every symbol the rewritten text names resolves: `createScript`, `writeSettingsHooks`, `writeAgentFiles`, `writeSkillDir`, `writeRegistry`, `captureStdout`, `captureOutput`, `fakeCLI`, `blockedDir`, `withTempCache`, `userCacheDir`, `badHomes`, and both `RefusesBadHome` tests.
- Citation count in `docs/` dropped from 114 to 70 by my regex (your 78 uses a wider one).

**Repo-wide sweep for other descriptions of the old job graph** — `CLAUDE.md`, `README.md`, `docs/README.md`, `skills/`, `agents/`, `hooks/`, `cli/` are all clean. `scripts/govulncheck.sh:4` is the only miss.

---

### Follow-up candidates (listing, not filing)

- ~70 `<file>:<line>` citations remain live in `docs/`; the drift is in-range-but-wrong, so only reading the cited line finds it. A CI guard or a convention change would pay for itself — this is the second sweep.
- The `Last verified … against commit <sha>` stamps in the kit docs are stale (correctly left alone while #154 and #162 touch the same files).
- `ci.yml` has no `concurrency:` group, so a rapid push sequence runs overlapping full suites. Pre-existing, but the new `schedule` adds one more source of runs.
- Branch protection / required checks remain absent (an explicit #155 non-goal). If they're ever added, note the names are `ci / test`, `ci / hooks`, `ci / govulncheck` on a tag but bare `test`, `hooks`, `govulncheck` on a PR — a required-check config has to cover both.

.github/workflows/ci.yml
```diff
@@ -4,6 +4,21 @@ on:
...
   push:
     branches: [main]
+  # Weekly run of the whole suite on the default branch. Without it an advisory
+  # published while `main` is quiet waits for the next push, PR or tag; a
+  # failure here is a red `ci` run on `main`. The jobs are defined once — a
+  # scan-only workflow would be a second copy of `govulncheck` — and the two
+  # test jobs cost little and also catch drift in the runner image and the
+  # pinned actions.
+  schedule:
+    - cron: "27 6 * * 1" # Mondays 06:27 UTC
```
alexandrocuma, line 14: 🔵 **Consider adding `workflow_dispatch:` here rather than deferring it.**

The PR lists this as out of scope, and I'd normally respect that — but the `on:` block *is* what this diff rewrites, it's one line, and the PR's own documentation argues for it. `docs/development/testing.md:77` says:

> no run, and so no failure, is the one way this can go quiet

With `schedule` as the only unattended trigger and no manual one, there is no way to run the scan on demand. That bites in exactly the situations that matter:

- After the 60-day auto-disable and a `gh workflow enable`, you can't confirm the scan works again without waiting until Monday or pushing a no-op commit.
- An advisory drops on a Tuesday and you want an answer now — the only options are re-running a stale run (which scans the *old* commit) or pushing something.
- Changing the cron can't be smoke-tested.

```yaml
  schedule:
    - cron: "27 6 * * 1" # Mondays 06:27 UTC
  # Manual re-run of the same suite — after a 60-day auto-disable, or to check a
  # freshly published advisory without waiting for Monday.
  workflow_dispatch:
  # Called by release.yml on a `v*` tag, so the same jobs gate the publish.
  workflow_call:
```

`workflow_call` and `workflow_dispatch` coexist fine and actionlint 1.7.12 accepts the combination — I checked. No downside I can find; the button is only visible to people who can already push.

.github/workflows/ci.yml
```diff
@@ -4,6 +4,21 @@ on:
...
+  # Called by release.yml on a `v*` tag, so the same jobs gate the publish.
+  workflow_call:
+
+# Every job here only reads the repo. goreleaser, the one job that writes, is
+# in release.yml and raises this for itself. A caller can only lower these.
+permissions:
```
alexandrocuma, line 27: 🔵 Nit. This new workflow-level block makes the `govulncheck` job's own `permissions: contents: read` (lines 62-65) redundant — it grants nothing extra now.

Harmless, and a defensible belt-and-braces. But the comment directly above says "Every job here only reads the repo", which leaves the surviving per-job block looking like it must be doing something the others don't. Either drop it, or add four words (`# redundant with the workflow default; kept explicit because this job handles untrusted advisory data`) so the next reader doesn't go hunting for the difference.

✅ The block itself is correct and the reasoning in the comment holds. I confirmed against the docs that a called workflow's permissions *"can be only downgraded (not elevated) by the called workflow"*, so with the caller also at `contents: read` this composes exactly as described, and `goreleaser` remains the sole holder of `contents: write`.

docs/guides/release.md
```diff
@@ -17,33 +17,33 @@ Asset edits reach the two targets at different times. Clone users get them on th
...
 
 If the run fails, read which job and exit status:
 
-- **`govulncheck` exited 1 (infrastructure, not a finding)** — a fetch or network error (vuln.go.dev, the module proxy), marked `infrastructure failure, not a finding` in the log. Re-run: `gh run rerun <run-id> --failed`. This re-runs `govulncheck` and then `goreleaser`, which still finds the draft.
-- **`govulncheck` exited 3 (called vulnerability), or `goreleaser` failed and can't be fixed by re-running** — this tag won't ship:
+- **`ci / govulncheck` exited 1 (infrastructure, not a finding)** — a fetch or network error (vuln.go.dev, the module proxy), marked `infrastructure failure, not a finding` in the log. Re-run: `gh run rerun <run-id> --failed`. This re-runs the failed `ci` jobs and then `goreleaser`, which still finds the draft.
```
alexandrocuma: 🔵 **This recovery step now describes a different GitHub code path than the one it was written for, and it hasn't been exercised.**

> Re-run: `gh run rerun <run-id> --failed`. This re-runs the failed `ci` jobs and then `goreleaser`, which still finds the draft.

Before this PR, `govulncheck` was a plain job in `release.yml` and the behaviour was unambiguous. Now the failed job lives inside a **called reusable workflow**, which is a distinct path in GitHub's re-run logic and a historically rough one: "Re-run failed jobs" used to disappear entirely on runs containing reusable workflows ([community #13501](https://github.com/orgs/community/discussions/13501), since fixed), and there is a still-open report that for a reusable workflow **in a matrix**, failed jobs re-run without their dependents ([community #52505](https://github.com/orgs/community/discussions/52505)).

This call isn't a matrix, so it most likely behaves as written — but the PR notes say the `workflow_call` path "is validated statically here rather than executed", and this is the one sentence a releaser follows when a release is half-done and vuln.go.dev is flaky. Being wrong here costs a version number.

**Two ways to close it:**

1. Hedge it — append: *"If the run offers no failed-jobs re-run (reusable-workflow runs have had gaps here), re-run all jobs: `gh run rerun <run-id>`. It repeats the passing suites, which is only a few minutes, and `goreleaser` still finds the draft."*
2. Or confirm it once on the first real release and stamp the sentence with the tag you saw it on — the guide already stamps its verification claims that way.

The rest of this recovery block is good: ✅ the `ci / test` / `ci / hooks` additions to the next bullet correctly widen the "this tag won't ship" case to the new jobs, and ✅ the draft handling in 46-49 is right — with `use_existing_draft: true` a skipped `goreleaser` leaves the `/release` draft sitting unpublished with no assets, and "delete the draft, keep the tag, cut the next patch" is exactly the correct recovery.

docs/development/testing.md
```diff
@@ -42,13 +42,13 @@ node hooks/opencode/secret-guard.test.js
...
-- **Fixtures / factories:** built per test by `t.Helper()` functions that write files into a `t.TempDir()` — `createScript`/`writeSettingsHooks` (`internal/hooks/installer_test.go`), `writeAgentFiles` (`internal/agents/installer_test.go`), `writeSkillDir` (`internal/skills/installer_test.go`), `writeRegistry` (`cmd/install_test.go:774`). There is no shared test-helper package: copy the helper you need (`captureStdout` exists in both `cmd/install_test.go:615` and `internal/mcp/mcp_test.go:170`; `captureOutput` in `internal/ui/ui_test.go:131`).
+- **Fixtures / factories:** built per test by `t.Helper()` functions that write files into a `t.TempDir()` — `createScript`/`writeSettingsHooks` (`internal/hooks/installer_test.go`), `writeAgentFiles` (`internal/agents/installer_test.go`), `writeSkillDir` (`internal/skills/installer_test.go`), `writeRegistry` (`cmd/install_test.go`). There is no shared test-helper package: copy the helper you need (`captureStdout` exists in both `cmd/install_test.go` and `internal/mcp/mcp_test.go`; `captureOutput` in `internal/ui/ui_test.go`).
```
alexandrocuma: 🔵 Small factual slip in a line this sweep rewrote. `captureStdout` isn't in "both" of two files — it's defined in **four**:

```
cli/cmd/install_test.go:1155
cli/internal/agents/installer_test.go:397
cli/internal/mcp/mcp_test.go:171
cli/internal/skills/installer_test.go:248
```

Same for `captureOutput`, which is in `internal/ui/ui_test.go:131` *and* `internal/hooks/opencode_test.go:22`.

Minor, but it slightly undersells the sentence's own point — "there is no shared test-helper package: copy the helper you need" is a stronger observation at four copies than at two. Suggest "`captureStdout` is copied into four test packages (`cmd`, `internal/agents`, `internal/mcp`, `internal/skills`); `captureOutput` into two".

✅ Separately: dropping the line numbers here was correct — the old `cmd/install_test.go:615` was really at 1155 and `internal/mcp/mcp_test.go:170` pointed at the doc comment rather than the `func`. Every symbol this rewrite names resolves; I checked all of them.

docs/architecture/overview.md
alexandrocuma: Fixed in 96694e5 — took the suggested rewrite almost verbatim:

```sh
# Used by .github/workflows/ci.yml (job `govulncheck`), which release.yml calls
# on a tag, and locally.
```

You are right that this was an oversight and not a scoping call: same sweep, same fact. Thanks for checking the rest of the repo — that saved me re-running the search.

docs/development/testing.md
alexandrocuma: Fixed in 96694e5, and your quote sent me back to the docs, where there turned out to be a third rule that matters more here than the one I had wrong.

The events reference says what you quoted (recipient = whoever last modified the `cron` syntax). The notifications page adds two more: the default recipient is the user who **created** the workflow, and — this is the one — *"If a scheduled workflow is disabled and then re-enabled, notifications will be sent to the user who re-enabled the workflow rather than the user who last modified the cron syntax."*

That override fires on exactly the sequence the next sentence describes: 60-day auto-disable, then someone re-enables. So the two facts belonged in the same paragraph rather than adjacent ones, and the section now states all three rules, notes that the previous recipient is never told they stopped being notified, and that it depends on the account having Actions notifications on.

Also folded in both of your smaller points: 60 days exactly, scoped to *"In a public repository"* so the caveat survives a visibility change, and the off-the-hour cron rationale now has the docs sentence behind it.

.github/workflows/ci.yml
alexandrocuma: Added in 96694e5 — your argument beat my scoping call. The deciding point was that the docs in this very PR name "no run, and so no failure" as the failure mode and then offer no way out of it.

```yaml
  # Manual re-run of the same suite: after the 60-day auto-disable, or to scan a
  # freshly published advisory without waiting for Monday. The schedule is
  # otherwise the only unattended trigger, and a run that never starts is the
  # one failure mode that raises no alarm.
  workflow_dispatch:
```

Confirmed independently: actionlint 1.7.12 clean with all five triggers, and a parse assertion that `workflow_dispatch` and `workflow_call` coexist (both null-valued, no inputs). testing.md now points at it as the recovery after a re-enable, and it is in the changelog entry.

docs/guides/release.md
alexandrocuma: Hedged in 96694e5 — took option 1, and added the last-resort step:

> Try `gh run rerun <run-id> --failed` first: it should re-run the failed job and then `goreleaser`, which still finds the draft. **Unverified on this path** — since #155 the failed job sits inside a called workflow, and GitHub's re-run-failed-jobs handling of runs containing a reusable workflow has had gaps. If the run offers no failed-jobs re-run, or `goreleaser` doesn't follow the retry, re-run the whole run: `gh run rerun <run-id>`. That repeats the passing suites — a few minutes — and `goreleaser` still finds the draft. If neither gets there, treat it as the next case and cut the next patch. Stamp this bullet with the tag the first real release confirms it on.

Option 2 stays open and the sentence now invites it. I did not want to claim it verified — the only honest way to exercise it is a real release, and the one thing I was not going to do on this PR is push a tag.

docs/development/testing.md
alexandrocuma: Fixed in 96694e5. Confirmed your counts before changing anything — `captureStdout` in `cmd`, `internal/agents`, `internal/mcp`, `internal/skills`; `captureOutput` in `internal/ui` and `internal/hooks` (and the two `captureOutput` copies do not even share a signature: `(string, string)` vs `string`).

Took your phrasing, since you are right that it strengthens the sentence's own point: "`captureStdout` is copied into four test packages (`cmd`, `internal/agents`, `internal/mcp`, `internal/skills`) and `captureOutput` into two (`internal/ui`, `internal/hooks`)".

.github/workflows/ci.yml
alexandrocuma: Reworded rather than dropped, in 96694e5 — you named the real problem, which was the comment implying a difference that was not there.

Kept the block because this is the job that fetches and evaluates third-party advisory data, so I would rather its scope be legible at the job. Both comments now say that outright: the workflow-level one names `govulncheck` as restating it deliberately, and the job carries `# Same as the workflow default above, restated because this is the job that fetches and evaluates third-party advisory data.` `persist-credentials: false` untouched.

PR #167, opened 2026-09-18
fix(hooks): run every fail-closed guard under a scan budget (#162)

### LINKED ISSUE #162: Security guards fail open when a scan is slow: add a fail-closed scan budget

alexandrocuma:
Closes #162

## Why
Claude Code's hooks reference is explicit: *"A timed-out `command`, `http`, or `mcp_tool` hook doesn't block the tool call."* `timeout` is in seconds, defaults to 600, and no field expresses "block on timeout". So a guard that is slow on some input fails **open** after a long wait. The opencode modules run synchronously with no timeout at all, so there the failure mode is a freeze.

(Trap for future readers: the Agent SDK's *callback* hooks page says a callback that exceeds its timeout does block. Different mechanism, doesn't apply to command hooks.)

## What changed
- **A shared scan budget, `DEVEXP_SCAN_BUDGET_MS`, default 15 s, ceiling 44 s**, per guard invocation, honoured by both twins. Over budget ⇒ **block** (exit 2 / throw), never allow. Only a plain **ASCII** non-negative integer counts, surrounding spaces trimmed, on both sides — so a typo can neither widen the budget nor disable the guard; `0` is the deterministic test seam.
  - The default is chosen from measured worst cases on v0.9.2's own pathological inputs (1.0 s for a 2 MB write, 1.7 s for a 1 MB command, 0.07 s ordinary) — ~9× headroom, above the ceilings the existing timing suites already accept, and two orders of magnitude under the 600 s default.
  - A value above the ceiling is **clamped, not honoured**, with a notice on stderr (once per process). Above the registered hook timeout, Claude Code cancels the guard first and a cancelled command hook doesn't block — the one knob on offer could otherwise put the bug back. The ceiling sits just under the registered 45 s, and both `scan-budget.test.sh` and the Go registry test fail if a timeout ever drops to or below it.
- **Claude Code:** `scan-budget.sh` re-runs the guard under a python watchdog in its own process group and SIGKILLs the group on a hit, so the budget covers the whole scan including `grep`. It exits 2 on a hit, on any child status that is neither 0 nor 2, and when the watchdog itself can't start.
  - The window *before* the budget exists is floored by a prologue trap in each guard, not an `if`: under `set -e` bash abandons the script at the failing `.` and never reaches the body, and for five distinct failures it then reports **0** to an EXIT trap — which reads as *allow*. Reaching that trap at all is a block, and the trap is lifted only where a status is actually known (in the budgeted run, and *after* the `python3` line — before it, the unbound constant is dereferenced on that very line). Six failure modes are tested: missing, unreadable, no entry point, unparsable, a directory, and a helper carrying the entry point but not the constants above it.
  - The marker that tells the budgeted run apart travels in **argv**, which only the watchdog can write. As an environment variable it would be inherited from whatever started Claude Code — a settings `env` entry, a shell profile, a CI image — and would silently switch the budget off for every guard.
  - A second marker, a depth counter, *is* read from the environment, because the only thing it can do is block. Writing the mutant for the one above showed why: with the argv marker unrecognised, every budgeted run starts a watchdog of its own, and since each level gets its own session a watchdog only kills its own child's group — the guard forks without bound (my first sweep reached ~479 processes). It now stops with an internal error in 0.16 s instead.
- **opencode:** the deadline is checked as the scan runs and throws a `ScanBudgetError`. Upstream `Plugin.trigger` has no try/catch around hook dispatch, so a throw reliably aborts the tool call. One cadence across the guards: every unit where units are few and expensive (the 11 regexes of a write), sampled every 1024 where they are many and cheap (a token, a line), and every 65536 characters/nodes inside `dangerous-cmd-guard`'s masking pass — its largest single piece of work. The distinct error type is what lets `maskInert`'s catch (which turns a refused parse into "scan everything") pass a spent budget through instead of swallowing it.
- **Registry:** the three `fail_closed` guards declare `claude_code.timeout: 45` — 3× the default budget and just above the ceiling, so the guard's own exit 2 always lands before Claude Code cancels it. The installer now writes that timeout into `settings.json`, and **converges it on an existing registration** (`registerHook` used to short-circuit on an already-registered command, so every existing install would have kept the 600 s default forever). A hook that declares no timeout leaves whatever is registered alone.

## Verification
- **New suites:** `scan-budget.test.sh` 64, `scan-budget.test.js` 81. Go `internal/hooks` gained `TestInstallClaude_Timeout` (5 subtests, each also asserting a second install is byte-identical), `TestInstallClaude_TimeoutKeepsOtherFields`, `TestRepoRegistry_FailClosedTimeouts`.
- **Budget covers the whole scan:** a 2 MB write that takes ~1 s is cut off at a 300 ms budget and blocks in 0.35 s; the same write passes under the default.
- **Parity:** the JS suite runs the shell twin on the equivalent envelope and compares verdicts — 3 guards × {over budget, ordinary input, real finding} all agree, the message's seconds match at 0/1/250/1250 ms, and the values that used to divide the twins (a non-ASCII digit, a digit-like character, spaces around the value) now agree too. Default and ceiling are pinned from all three of shell, JS and Go.
- **Mutations: 41 mutants, all killed bar one equivalent**, one per behaviour and one per review finding — the prologue trap removed, each pre-check removed, the floor flipped to allow, the marker back in the environment, the sentinel dropped, the re-entry limit off by one, the clamp removed or silenced, either ceiling drifted, `str.isdigit()` restored, trimming dropped on either side, the masking pass unbudgeted, its catch swallowing a spent budget, `maskInert` called without the budget, and the isolation test's allowlist widened.
  - The sweep found four gaps of its own, all the same shape — the test watched for an outcome that something else also produced:
    - the **rule loop**'s check was no longer what refused its input once masking carried the budget (a command containing `esac` isolates it: `maskInert` hands back anything it cannot classify without parsing it);
    - the **clamp** was pinned by its notice, not its effect (`DEVEXP_SCAN_BUDGET_CEILING_MS` now lowers the ceiling — and only lowers it — so a 60 s budget under a 300 ms ceiling can be watched blocking at 0.3 s);
    - the prologue's **`command -v`** check was subsumed by the floor, which blocked anyway with the same message — the case now also requires no shell error in the output;
    - the **masking pass**'s budget was pinned by "did it refuse", which it did either way, just later — the check is counted now, not timed.
  - The last survivor turned out to be a real hole rather than an equivalent mutant. The masking *walk* sampled per pipeline and per simple command, and one command can hold millions of words — so the part of the walk that costs anything (~570 ms on an 8 MB command, after the parse had stopped ticking) was unwatched. It samples per word now, and both the "walk unsampled" and "word loop unsampled" mutants die.
  - One equivalent survivor, as before: removing the `ms <= 0` short circuit. `child.wait(timeout=0)` raises immediately and the child is killed before it can act, so nothing observable changes. Kept so that "zero budget = no scan at all" is a property of the code rather than of scheduling.
- Every hook suite, the installer script suites and `go test -race ./...` pass after merging `main`.

## Cost
One extra `python3` **and** one extra `bash` per guarded tool call — the guard is re-run as a child, not `exec`'d. Median of 30 warmed runs against `origin/main`, same machine: `dangerous-cmd-guard` 66 → 129 ms, `secret-guard` 46 → 130 ms, `secret-in-write-guard` 46 → 130 ms, one Bash tool call (two guards in parallel) 66 → 130 ms. So **+60 to +85 ms**, about double a guard's run. An independent run on the same head measured 69 → 144, 50 → 135 and 50 → 137 ms. That buys one uniform budget that also covers the shell's `grep` phase; the alternative (arming a deadline inside each guard's own python) is two mechanisms and doesn't cover `grep`. The opencode side's checks measured within noise (−6% to +2%).

## Worth knowing
- **Any external command a guard runs before it can exit 2 is a fail-open path.** A first wiring used `$(dirname …)`; with `dirname` unavailable the guard exits 1, which Claude Code reads as non-blocking. Replaced with pure bash parameter expansion.
- **An `if ! . …` is not a floor** — see the prologue trap above. On bash 3.2 an EXIT trap that exits `$?` would have turned three of these failures into exit **0**, worse than the exit 1 it replaced; the trap therefore asks no questions.
- `interpreter-isolation.test.sh` keeps its broad `hooks/claude-code/*.sh` walk, with an explicit `HELPERS` allowlist for files that have no envelope of their own (today just `scan-budget`), so the next shared helper that runs `python3` has to be exempted on purpose.
- `docs/reference/hooks.md` carried a stale claim that there is no time budget; corrected, with a new "The Scan Budget" section.

- opencode upstream: in `session/prompt.ts` `handleSubtask`, the `tool.execute.before` trigger sits outside the `Effect.catchCause` wrapping `taskTool.execute`, so a throwing hook escapes there rather than becoming a per-tool error.
- `time.monotonic()` isn't comparable across processes here; any future cross-process timing test must use wall clock.
- Raised in review and left for their own tickets: a hostile `python3` earlier on `PATH`, and the opencode column of the cost table.
- Noted in review, no action: a sentinel handed in via an edited registered command still switches the budget off — tamper-evident, and in a file the installer rewrites; and an ambient depth value bricks every guarded call with a confusing message, but fails closed.


alexandrocuma (commented):
## Security review — scan budget for the fail-closed guards (#162)

**Recommendation: Request changes.** One in-diff fail-open, plus a silent kill switch and a missing ceiling on the one documented knob.

The design is right and the hard parts are done well. I verified the premises against the live hooks reference rather than the PR body: a timed-out `command` hook does not block the tool call, the default is 600 s, exit 2 blocks, and **exit 1 / 127 / any other non-zero is a non-blocking error — the action proceeds**. That last point is what makes finding 1 a real fail-open. `timeout` is spelled on the handler object in seconds, which is exactly where the installer writes it.

### What I verified (not just read)

- **Process-group kill is correct and tight.** `start_new_session=True` + `killpg` — I planted a grandchild under a deliberately slow guard: budget hit at 1.29 s for a 1.2 s budget, grandchild reaped, zero orphans, caller untouched. This is the part that could have gone very wrong, and it didn't.
- **Large stdin survives the extra hop.** The envelope reaches the re-run guard intact at 6 MB and 60 MB (1.33 s), with a secret at the very tail still blocking. No truncation, no double-read.
- **`python3 -I -S` neutralises `PYTHONPATH`, `PYTHONHOME` and `PYTHONSTARTUP`** — confirmed with hostile values, still blocks correctly.
- **`$BASH` can't be hijacked** from the environment (bash resets it), so `"${BASH:-bash}"` is safe.
- **Unknown child status and a watchdog that can't start both block** — confirmed with an empty `PATH` and with children exiting 7 / 1 / 127.
- **Installer** (drove `InstallClaude` directly against a scratch settings.json): exactly the three `fail_closed` guards get `timeout: 45`, advisory hooks get none; a fresh install is byte-identical on re-run; a pre-timeout install converges while `env`, `permissions` and a user's own extra handler fields survive; an advisory hook's user-set `timeout: 120` is preserved; a guard's user-set timeout is brought to 45. The #137 byte-preservation property holds.
- **Mutation testing**: 16 mutations of my own across the shell helper, both opencode twins and the Go settings writer — **16/16 killed**. I checked each mutation actually applied first. The suites are genuinely load-bearing. No survivors to report.
- All four CI suites pass on my scratch copy.

### Findings

**1. Critical — the "helper can't be sourced" block never runs; the guard exits 1.** In-diff, all three guards. Under `set -euo pipefail`, bash aborts the script when `.` cannot open the file, *before* the `if !` body. Status 1, message never printed → non-blocking error → the tool call proceeds **unscanned**. The PR body calls this block out as the mitigation for exactly this class; it doesn't fire. Details inline.

**2. Important — the re-entrancy marker is an inherited environment variable.** `scan-budget.sh` skips the watchdog whenever `DEVEXP_SCAN_BUDGET_RUNNING` is already `1` in the environment. Hooks inherit Claude Code's environment including the `env` block of settings.json, so an ambient value silently turns the budget off on every guard with no warning. A marker that only the watchdog can set (an argv sentinel — command hooks get no arguments) closes it.

**3. Important — nothing keeps the budget below the hook timeout.** The registry pins `timeout: 45` at 3× the 15 s default, but `DEVEXP_SCAN_BUDGET_MS` is unclamped: both twins happily accept 600000. Above 45 s the hook is cancelled before its own exit 2 can land, which reinstates precisely the fail-open this PR removes — via the one knob the docs tell people about.

**4. Important — the twins disagree on budget parsing, in both directions.** The JS side trims, the Python side doesn't; Python's `isdigit()` accepts non-ASCII digits, JS's `\d` doesn't. Three concrete divergences inline. Both file headers and `docs/reference/hooks.md` claim "both twins decide alike" and "only a plain non-negative integer counts".

**5. Minor — the docs understate the cost.** Details inline, with numbers I measured.

**6. Minor — `maskInert()` runs before the first budget check** on the opencode side, so the masking pass is unbudgeted.

**7. Nit — inconsistent clock sampling** between the two opencode guards.

**8. Nit — `interpreter-isolation.test.sh` narrowed its own net** to registry-registered scripts.

### Follow-up candidates (pre-existing, not for this PR)

- A `python3` earlier on `$PATH` that exits 0 turns every guard into a no-op — I confirmed the call is allowed. Pre-existing (the guards already trusted `python3`), but the wrapper widens the blast radius from "the parse step" to "the whole guard", so it may be worth a note in the threat model.
- `docs/reference/hooks.md` gives opencode's budget cost as "none"; it's about +9% on a token-heavy command.

### Done well

- The failure taxonomy in `scan-budget.sh` (hit / unknown status / can't start, all → 2) is the right shape, and the header comment explains *why* a trap and `timeout(1)` were rejected — future readers won't re-litigate it.
- Deriving the helper directory with parameter expansion instead of `dirname` is the correct instinct, and the no-PATH test pins it.
- The cross-twin parity suite (same input, same verdict, same seconds) is a good pattern and worth the precedent `docs/development/testing.md` now sets for it.
- The installer change converges an existing registration instead of short-circuiting — without it every existing install would have sat on the 600 s default for ever. That's the detail most people would have missed.

hooks/claude-code/dangerous-cmd-guard.sh
```diff
@@ -20,6 +20,19 @@
...
 set -euo pipefail
 
+# The whole scan runs under a wall-clock budget and blocks when it is exceeded:
+# a slow scan must never let a tool call through unchecked (#162). The helper is
+# found without running anything: an external command that is missing would end
+# the guard with a status Claude Code reads as a non-blocking error, which is
+# the fail-open this guards against. Failing to load it blocks.
+devexp_dir="${BASH_SOURCE[0]%/*}"
+if [ "$devexp_dir" = "${BASH_SOURCE[0]}" ]; then devexp_dir="."; fi
+if ! . "$devexp_dir/scan-budget.sh"; then
```
alexandrocuma: **Critical — this block does not fire, and the guard fails open.**

Under `set -euo pipefail`, bash aborts the script when `.` cannot open the file. It never reaches the `if !` body: the message is not printed and the status is **1**, not 2. Per the hooks reference, exit 1 is a non-blocking error and the tool call proceeds — **unscanned**. Same for `secret-guard.sh:32` and `secret-in-write-guard.sh:35`.

Repro (the shipped `set -euo pipefail`, macOS bash 3.2):

```
$ cat > p.sh <<'EOF'
#!/usr/bin/env bash
set -euo pipefail
if ! . /no/such.sh; then echo BODY; exit 2; fi
EOF
$ bash p.sh; echo "rc=$?"
p.sh: line 3: /no/such.sh: No such file or directory
rc=1              # BODY never printed
```

Against the real guard, with the helper removed or mode 000: `rc=1`, no devexp message. Drop `set -e` and the same script gives `rc=2` — so the block only works in the one configuration the guards don't use.

Adjacent case: if the helper is present but truncated so the function is missing, sourcing succeeds and `devexp_scan_budget` is `command not found` → **127**, also non-blocking.

Both are reachable in the field — a partial or interrupted extraction, a stale `settings.json` pointing at a tree from before this existed, a restrictive umask on the cache dir.

Suggested fix — don't rely on `.`'s return status, and put a floor under everything before the watchdog takes over:

```bash
# Any exit that is neither allow (0) nor block (2) before the budget is armed
# is a fail-open; make it a block.
devexp_budget_floor() {
    rc=$?
    case $rc in 0|2) ;; *)
        echo "[devexp dangerous-cmd-guard] internal error -- the guard did not run (exit $rc). Blocking to be safe." >&2
        rc=2 ;;
    esac
    trap - EXIT
    exit "$rc"
}
trap devexp_budget_floor EXIT

devexp_dir="${BASH_SOURCE[0]%/*}"
if [ "$devexp_dir" = "${BASH_SOURCE[0]}" ]; then devexp_dir="."; fi
[ -r "$devexp_dir/scan-budget.sh" ] || exit 1      # the trap turns this into a block
. "$devexp_dir/scan-budget.sh"
command -v devexp_scan_budget >/dev/null || exit 1
devexp_scan_budget dangerous-cmd-guard "$@"
```

The trap also covers the 127 case and any future `set -e` abort in this prologue. It must be cleared before the guard's own `exit 2` path, or arranged so a legitimate 0/2 passes through unchanged (the `case` above does that).

Worth a test, too: `scan-budget.test.sh` has no case for a missing or unreadable helper. Every suite *does* fail when the helper is deleted, but only because the guards then can't run at all — nothing asserts the exit status is 2.

hooks/claude-code/scan-budget.sh
```diff
@@ -0,0 +1,117 @@
...
+    raise SystemExit(rc)
+block('[devexp %s] internal error -- the guard exited %d, which is neither allow (0) nor '
+      'block (2), so its decision is unknown. Blocking to be safe.\n' % (guard, rc))
+PY
+
+# devexp_scan_budget — run the calling guard under its budget, then exit with
+# its status. Returns without doing anything when it is already the budgeted
+# run, so the guard's own body runs exactly once.
+devexp_scan_budget() {
+    if [ "${DEVEXP_SCAN_BUDGET_RUNNING:-}" = "1" ]; then
```
alexandrocuma: **Important — this marker is a plain inherited environment variable, so an ambient value disables the budget on every guard.**

The hooks reference confirms handlers "run in the current directory with Claude Code's environment", and that the settings `env` block "reach[es] every subprocess Claude Code starts". So if `DEVEXP_SCAN_BUDGET_RUNNING=1` is present in the parent environment — a settings `env` entry, a shell profile, a devcontainer or CI image, a wrapper script — every `fail_closed` guard runs its body directly with no watchdog and no message. Verified: the guard still blocks on its own merits, but the budget is simply gone, and nothing at runtime notices.

(Good news: per the same docs, variables exported inside a Bash tool call do *not* reach later hooks, so this isn't reachable from a tool call. It's still a silent off switch for a security control.)

The concern in the header comment — "an unrelated nested session" — is the same shape: anything the guard's own scan spawns inherits the marker.

Suggested fix: make the marker something only the watchdog can produce. Claude Code passes no arguments to a command hook, so argv is a clean channel:

```python
child = subprocess.Popen(cmd + ['--devexp-budgeted'], start_new_session=True)
```

```bash
devexp_scan_budget() {
    local guard="$1"; shift
    if [ "${1:-}" = "--devexp-budgeted" ]; then return 0; fi
    ...
}
```

A random per-run nonce works too, but argv alone is enough here and keeps the env clean.

hooks/claude-code/scan-budget.sh
```diff
@@ -0,0 +1,117 @@
...
+IFS= read -r -d '' DEVEXP_SCAN_BUDGET_PY <<'PY' || true
+import os, signal, subprocess, sys
+
+guard, raw, default = sys.argv[1], sys.argv[2], sys.argv[3]
+cmd = sys.argv[4:]
+
+# Only a plain non-negative integer counts; "10s", "-1", "1e3" and an empty or
+# unset variable all fall back to the default, so a typo cannot silently widen
+# or disable the budget.
+ms = int(raw) if raw.isdigit() else int(default)
```
alexandrocuma: **Important — `str.isdigit()` is not "a plain non-negative integer", and it disagrees with the JS twin.**

Three divergences from `hooks/opencode/utils.js:33` (`/^\d+$/.test(raw.trim())`), all measured:

| `DEVEXP_SCAN_BUDGET_MS` | Claude Code | opencode |
|---|---|---|
| `"  250  "` | 15000 (no trim) | 250 (trims) |
| `"\u0663"` (Arabic-Indic 3) | **3 ms** — blocks every call | 15000 |
| `"\u00b2"` (superscript 2) | `isdigit()` true, `int()` raises → raw `Traceback` on stderr, exit 1 → remapped to 2 | 15000 |

The header comment here, the one in `utils.js`, and `docs/reference/hooks.md` all say the twins "decide alike" and that only a plain non-negative integer counts. `scan-budget.test.js` asserts the `"  250  " → 250` reading; nothing asserts what the shell does with it.

The superscript case fails closed, so it's not a hole — but the user gets a Python traceback and then "the scan budget could not run (exit 1)", which points at the wrong thing.

Suggested fix — one spelling, explicitly ASCII, on both sides:

```python
import re
ms = int(raw) if re.fullmatch(r'[0-9]+', raw.strip()) else int(default)
```

and add `"  250  "`, a non-ASCII decimal digit and a non-decimal digit character to both suites' bad-value tables (`scan-budget.test.sh` already iterates `abc -1 1e3 10s "" " " 1.5 0x10` — these are three more rows).

hooks/claude-code/scan-budget.sh
```diff
@@ -0,0 +1,117 @@
...
+
+# The default budget, in milliseconds. Sized from the worst cases measured on
+# the crafted inputs the #146/#158 timing tests use: ~1.0 s for a 2 MB write
+# through secret-in-write-guard, ~1.7 s for a 1 MB command through
+# dangerous-cmd-guard, ~0.07 s for ordinary input. That is ~9x headroom, and it
+# is above the ceilings those suites already accept as "not slow" (8 s for a
+# 2 MB write, 15 s for a 400 KB pipeline), so a much slower machine still never
+# trips it. hooks/registry.json sets each guard's hook `timeout` well above it,
+# so the guard's own block always lands first.
+DEVEXP_SCAN_BUDGET_DEFAULT_MS=15000
```
alexandrocuma, line 48: **Important — nothing keeps the budget below the registered hook timeout.**

The registry pins `timeout: 45` and the docs explain it as "three times the budget, so the guard's own exit 2 always lands first". But `DEVEXP_SCAN_BUDGET_MS` is unclamped — both twins accept `600000` without comment. Above 45 s the ordering inverts: Claude Code cancels the hook first, a cancelled command hook doesn't block, and the tool call proceeds unscanned. The one knob the docs advertise can reinstate the exact bug this PR fixes, and there's no warning.

Plausible route: someone hits the budget on a genuinely slow machine or a huge input, reads the message, and raises the budget until the blocks stop.

Suggested fix — clamp against the registered timeout, with the ceiling derived from the same constant rather than hardcoded twice:

```python
# A budget at or above the hook timeout is the fail-open this exists to
# prevent: the hook would be cancelled before it could exit 2.
ms = min(ms, ceiling)
```

passing the registry's timeout (or a `DEVEXP_SCAN_BUDGET_MAX_MS` derived from it) in, and printing a one-line notice on stderr when a value is clamped so it isn't silent. `docs/reference/hooks.md → The Scan Budget` should state the ceiling next to the default, and `scan-budget.test.sh` can pin `budget >= timeout` → clamped.

hooks/opencode/utils.js
```diff
@@ -6,6 +6,56 @@ import { spawn } from 'child_process';
...
+ * "already over budget" and blocks at the first check; anything that is not a
+ * non-negative integer is ignored in favour of the default.
+ * hooks/claude-code/scan-budget.sh reads the same variable and carries the same
+ * default, so both twins decide alike.
+ */
+export const SCAN_BUDGET_DEFAULT_MS = 15000;
+
+export function scanBudgetMs(env = process.env) {
+  const raw = env?.DEVEXP_SCAN_BUDGET_MS;
+  return typeof raw === 'string' && /^\d+$/.test(raw.trim()) ? Number(raw.trim()) : SCAN_BUDGET_DEFAULT_MS;
```
alexandrocuma: Twin of the parsing comment on `scan-budget.sh:56` — this side trims and is ASCII-only, the Python side does neither. Whichever spelling wins, both should use it, and the shared claim in this file's header ("so both twins decide alike") should be true.

Also the budget ceiling from `scan-budget.sh:44` applies here in spirit: opencode puts no timeout on a plugin hook, so a 600 s budget here means a session that can stall for ten minutes on one tool call. Worth the same clamp even though there's no fail-open on this side.

hooks/opencode/dangerous-cmd-guard.js
```diff
@@ -724,22 +729,37 @@ export function maskInert(command) {
...
+ * rather than called per line because reading the clock costs real time, and a
+ * command can hold hundreds of thousands of lines.
+ */
+export function blockReason(command, overBudget = () => {}) {
+  overBudget();
   const lines = maskInert(command).split('\n');
-  for (const { test, label } of BLOCK_PATTERNS) for (const line of lines) if (test(line)) return label;
+  for (const { test, label } of BLOCK_PATTERNS) {
+    for (let i = 0; i < lines.length; i++) {
+      if ((i & 1023) === 0) overBudget();
```
alexandrocuma, line 788: **Minor — `maskInert()` runs before the first budget check, so the masking pass is unbudgeted.**

`overBudget()` is called on line 741, then line 742 runs the whole character-by-character masking pass with no further check; sampling only starts in the rule loop below. Measured: 172 ms for a 2 MB command, 1.68 s for 20 MB. Linear, so it's not a hang, but `docs/reference/hooks.md` describes this side as covering "the scan, at the granularity of one unit of work", and the masking pass — often the single largest piece of work — isn't in it.

Either pass `overBudget` into `maskInert` and sample it there, or say in the docs that masking is outside the budget and why that's acceptable (bounded and linear).

hooks/opencode/secret-guard.js
```diff
@@ -58,12 +61,14 @@ export function plausiblePath(tok) {
...
 
-export function secretInCommand(cmd) {
+/** `overBudget` is called once per token; it throws when the budget is spent. */
+export function secretInCommand(cmd, overBudget = () => {}) {
   if (!cmd) return null;
   const stripped = cmd.replace(HEREDOC, ' ');
   const tokens = stripped.match(/(?:[^\s"']+|"[^"]*"|'[^']*')+/g) ?? [];
   let skipPattern = false;
   for (const token of tokens) {
+    overBudget();
```
alexandrocuma: **Nit — inconsistent sampling between the two guards.**

`dangerous-cmd-guard.js:745` samples every 1024 lines with an explicit rationale ("reading the clock costs real time, and a command can hold hundreds of thousands of lines"). This one reads the clock on every token. Measured: +9% on an 800k-token command (78 ms → 85 ms).

Not a problem in itself, but the two files now document opposite conclusions about the same cost. `(i & 1023) === 0` here too, or drop the rationale there.

docs/reference/hooks.md
```diff
@@ -203,6 +208,72 @@ In `dangerous-cmd-guard.sh`, all parsing happens in the `python3 -I` step, where
...
+brings an existing one to the registry's value, so a machine installed before
+this existed stops running on the 600 s default at its next install.
+
+**How it is enforced.**
+
+| | Claude Code | opencode |
+|---|---|---|
+| Mechanism | `scan-budget.sh` re-runs the guard under a `python3` watchdog in a session of its own and `SIGKILL`s the process group at the deadline | a deadline started at handler entry, checked at every unit of the scan; the check throws, which is how `tool.execute.before` refuses a call |
+| Covers | the whole hook run — reading the envelope, `json.load`, the regex work and every `grep` | the scan, at the granularity of one unit of work: one pattern, one rule against one line, one token |
+| Cost | one extra interpreter start per guarded tool call | none |
```
alexandrocuma: **Minor — "one extra interpreter start" understates it, and there's no number.**

It's one extra `python3` *and* one extra `bash` (the guard is re-run as a child, it isn't `exec`'d). Measured on an M-series laptop, warm, 20 runs each:

| | before | after |
|---|---|---|
| `dangerous-cmd-guard` | 72 ms | 138 ms |
| `secret-guard` | 51 ms | 118 ms |
| `secret-in-write-guard` | 52 ms | 121 ms |
| one Bash tool call (2 guards in parallel) | 72 ms | 143 ms |

Roughly a doubling — consistent with the PR body's 69 → 124 ms, and fine for what it buys. But the PR body has the number and the doc doesn't, and this row is what a future reader will find. Suggest "one extra interpreter *and shell* start per guarded tool call — roughly +70 ms per guard on current hardware".

hooks/claude-code/interpreter-isolation.test.sh
```diff
@@ -77,10 +77,23 @@ check graphify-read-guard       '{"tool_name":"Read","tool_input":{"file_path":"
...
-for f in "$DIR"/*.sh; do
-  case "$f" in *.test.sh) continue ;; esac
-  name=$(basename "$f" .sh)
+# Every hook that starts a Python interpreter must be covered above. A hook is a
+# script hooks/registry.json registers; a helper the hooks source — scan-budget.sh
+# — has no envelope of its own and is covered through them: every check above runs
+# it from the crowded working directory as well, so a module it picked up there
+# would show in the sentinel.
+scripts=$(python3 -I -c '
+import json, os, sys
```
alexandrocuma: **Nit — this narrows the test's own net.**

The old loop caught *any* `hooks/claude-code/*.sh` that starts a Python interpreter. The new one only sees scripts `registry.json` registers, so the next shared helper that runs `python3` — the situation that prompted this change — is silently uncovered again.

The reasoning in the comment is sound for `scan-budget.sh` specifically (it has no envelope of its own, and `-I` keeps the working directory off `sys.path`). Consider keeping the broad `*.sh` walk and adding an explicit allowlist of helpers covered transitively, so the next one has to be added on purpose:

```bash
HELPERS="scan-budget"   # no envelope of their own; covered through the guards that source them
```

That keeps the property the test was written to hold while still letting `scan-budget.sh` through.

hooks/claude-code/dangerous-cmd-guard.sh
alexandrocuma: Confirmed and fixed, and thank you — this was the same class of fail-open the PR is about, one layer up.

One correction worth recording: the suggested trap, taken literally, would have made it **worse** on the bash the guards actually run under. `$?` inside an EXIT trap is not the aborting status for several of these on bash 3.2:

| failure | `$?` in an EXIT trap | no trap |
|---|---|---|
| `. missing-file` | **0** | 1 |
| `. unparsable-file` | **0** | 2 |
| unset variable under `set -u` | **0** | 1 |
| `. a-directory` | 1 | 1 |
| `false`, command not found | 1, 127 | 1, 127 |

So `exit "$rc"` after a `case $rc in 0|2)` would have turned three of them into exit **0** — allow — where today they are a non-blocking 1. The trap therefore asks no questions: reaching it at all means the budget never started, so it blocks unconditionally.

```bash
devexp_budget_why="stopped before its scan budget could start"
devexp_budget_floor() {
    trap - EXIT
    echo "[devexp secret-guard] internal error -- the guard $devexp_budget_why, so it did not run. Blocking to be safe." >&2
    exit 2
}
trap devexp_budget_floor EXIT
```

The `-r` and `command -v` pre-checks you suggested are there too, setting `devexp_budget_why` so each case names itself; they `exit 1` into the same trap rather than duplicating the message. `devexp_scan_budget` clears the trap the moment it takes over, in both the first run and the budgeted one, so from there the watchdog's existing 0-or-2 check is the floor — including for the 127 case, which is now unreachable before the trap and floored after it.

`scan-budget.test.sh` gained five cases, all checking exit 2 by itself: missing, unreadable (a broken symlink, so it holds as root too), no entry point, unparsable, and a directory. Mode 000 verified by hand as well. Four mutants cover it: trap not installed, either pre-check removed, floor flipped to allow.

hooks/claude-code/scan-budget.sh
alexandrocuma: Fixed — the marker is now an argv sentinel, as you suggested, and the environment variable is gone entirely.

```python
child = subprocess.Popen(cmd + [sentinel], start_new_session=True)
```

```bash
for arg in ${1+"$@"}; do
    if [ "$arg" = "$DEVEXP_SCAN_BUDGET_SENTINEL" ]; then return 0; fi
done
```

It scans all of argv rather than `$1` alone: Claude Code passes a command hook no arguments today, but the sentinel is appended, so this holds if that ever changes.

Writing the mutant for this turned up something worth fixing on its own: with the marker not recognised, each budgeted run starts a watchdog of its own, and since every level gets `start_new_session=True` a watchdog only ever kills its own child's group — so the guard forks without bound. My first sweep did exactly that and reached ~479 processes before I stopped it.

So there is now a second marker, a depth counter, read from the environment *on purpose* — the only thing it can do is block, so an ambient value cannot switch anything off:

```python
if depth >= MAX_DEPTH:
    block('... the scan budget re-entered itself, so the guard did not run. Blocking to be safe.')
child = subprocess.Popen(cmd + [sentinel], start_new_session=True,
                         env=dict(os.environ, **{depth_var: str(depth + 1)}))
```

With the sentinel check deliberately broken, the guard now stops in 0.16 s with that message instead of multiplying.

`scan-budget.test.sh` runs each guard with an ambient `DEVEXP_SCAN_BUDGET_RUNNING`, `DEVEXP_SCAN_BUDGET_SENTINEL` and `DEVEXP_BUDGET_CHILD` set to both `1` and `--devexp-budgeted`, with a zero budget, and requires a block every time — six cases; an ambient depth of 1, 2 or 9 blocks while 0, empty, `abc` and `-1` read as none; and a sandboxed copy with the sentinel check broken has to stop, and stop within 10 s, rather than fork. Three mutants: the marker put back in the environment, the watchdog no longer marking the budgeted run, and the re-entry limit off by one.

hooks/claude-code/scan-budget.sh
alexandrocuma: Fixed. `DEVEXP_SCAN_BUDGET_MAX_MS=44000` / `SCAN_BUDGET_MAX_MS = 44000`, one number per twin, and a larger value is clamped rather than honoured:

```
[devexp dangerous-cmd-guard] Note: DEVEXP_SCAN_BUDGET_MS of 600 s is at or above this guard's
registered hook timeout, which would let Claude Code cancel the guard before it could block. Using 44 s.
```

Once per process on the opencode side, since that one runs on every tool call.

I kept the ceiling as a constant next to the default rather than reading `registry.json` at hook time — that would put file I/O on every guarded tool call, for a number that only changes when someone edits the registry. The "derived, not hardcoded twice" property is enforced by tests instead: `scan-budget.test.sh` and `TestRepoRegistry_FailClosedTimeouts` both fail if any `fail_closed` guard's `timeout × 1000` is not strictly above the ceiling, so raising one without the other is caught. The Go test now checks the ceiling rather than the default, since the ceiling is the largest budget a guard can actually run with.

`docs/reference/hooks.md` states the ceiling next to the default and says why. Mutants: clamp removed, clamp silenced, either twin's ceiling drifted.

hooks/claude-code/scan-budget.sh
alexandrocuma: Fixed, with your spelling:

```python
ms = int(raw) if re.fullmatch(r'[0-9]+', raw.strip()) else int(default)
```

All three rows of your table now read the same on both sides: `"  250  "` → 250, `U+0663` → default, `U+00B2` → default (no traceback). The comment in the Python says why `str.isdigit()` is not the test, so it does not come back.

Both tables gained the rows: `scan-budget.test.sh` iterates `abc -1 1e3 10s "" " " 1.5 0x10 ٣ ² "12 34"`, plus a separate check that `"  0  "` is trimmed and read as 0 rather than falling back. `scan-budget.test.js` has the same values in its `scanBudgetMs` table, and the cross-twin section drives both guards with `U+0663`, `U+00B2`, `"  0  "` and `"  30000  "` and requires the same verdict. Four mutants: `isdigit()` restored, trimming dropped on either side, a non-ASCII digit accepted in JS.

hooks/opencode/utils.js
alexandrocuma: Both done on this side: the shell now uses the ASCII-only, trimmed spelling this file already had, and the ceiling is shared rather than being a Claude-Code-only concern — `SCAN_BUDGET_MAX_MS = 44000`, clamped in `scanBudgetMs`, with a notice printed once per process.

The header claim is now true and tested from three directions: `scan-budget.test.sh` and `scan-budget.test.js` each read the other twin's constants out of the source and compare, and `TestRepoRegistry_FailClosedTimeouts` checks both against the registry. Your point about a ten-minute budget stalling a session is in the doc as the reason this side has a ceiling at all.

hooks/opencode/dangerous-cmd-guard.js
alexandrocuma: Fixed by budgeting it rather than documenting around it — it is too much of the work to leave out.

`maskInert(command, overBudget)` now samples by position: `Parser.tick(i)` fires about once every 65536 characters, from the `parseScript` loop and from the four loops that walk characters one at a time (`word`, `dquote`, `ansi`, `blanks`), and the masking walk samples every 65536 nodes. Position rather than a counter because `i` only moves forward through the command, so the cadence holds however the parse jumps between loops.

The catch was the interesting part: it turns anything thrown into "scan the whole command", so it would have swallowed a spent budget and then scanned 2 MB anyway. A spent budget is now a distinct `ScanBudgetError` and the catch rethrows it.

Measured: on `echo <2 MB>` the refusal now lands at ~2 ms instead of running the full pass. Overhead against the previous commit on five shapes (one long word, 800k tokens, 250k lines, a long quoted string, an ordinary command) is −6% to +2%, i.e. inside noise. Three mutants: ticking disabled, the rethrow removed, `maskInert` called without the budget.

hooks/opencode/secret-guard.js
alexandrocuma: Fixed, and made deliberate rather than consistent-by-accident. There is now one rule, stated once in `startScanBudget`'s doc comment in `utils.js` and referenced from both guards:

> check every unit when units are few and each is expensive (one regex over the whole write: 11 of them), and sample every 1024 when units are many and each is cheap (one token, one line of a command that may hold hundreds of thousands).

So `secret-guard` samples `(i & 1023) === 0` over tokens, `dangerous-cmd-guard` keeps it over lines, `secret-in-write-guard` keeps its per-pattern check, and the masking pass samples by position. Your 9% figure is quoted in the rule as the reason. `docs/reference/hooks.md` says the same thing in one paragraph instead of two files disagreeing.

docs/reference/hooks.md
alexandrocuma: Fixed, with your numbers and your wording:

> one extra `python3` **and** one extra `bash` — the guard is re-run as a child, not `exec`'d — so roughly **+70 ms** per guarded tool call on current hardware (measured, warm, 20 runs: `dangerous-cmd-guard` 72 → 138 ms, `secret-guard` 51 → 118 ms, `secret-in-write-guard` 52 → 121 ms; one Bash tool call, two guards in parallel, 72 → 143 ms)

The PR body's older 69 → 124 ms has been replaced by the same table, so the two no longer disagree. The opencode column now carries a number too: the masking-pass checks measured within noise (−6% to +2%).

hooks/claude-code/interpreter-isolation.test.sh
alexandrocuma: Fixed exactly as suggested — the broad `hooks/claude-code/*.sh` walk is back, with an explicit allowlist:

```bash
HELPERS="scan-budget"   # no envelope of their own; covered through the guards that source them
```

Plus a check that every allowlisted name actually exists, so the list cannot rot into a silent exemption for a file that was renamed. The comment says why the exemption is sound for this one (covered transitively — every `check` above runs it from the crowded working directory too) and that adding another is deliberate.

A mutant widening the allowlist to a second, non-existent helper is in the set and dies.

hooks/claude-code/dangerous-cmd-guard.sh
alexandrocuma: Follow-up: the mutation sweep showed the `command -v` check was *subsumed* by the floor — delete it and the guard still blocks, via the trap, with the same message, because `devexp_budget_why` is set on the line above. The difference is what the user sees: without it bash's own `command not found` reaches them first.

So the case now also requires the guard to say what happened itself, with no shell error in the output. With that, the mutant dies.

hooks/claude-code/scan-budget.sh
alexandrocuma: Follow-up: the sweep caught that I had pinned the *notice* and not the *effect* — delete `ms = ceiling` and the notice still prints while the budget stays at the value asked for.

Watching a 44 s budget bite means waiting 44 s, so there is now a `DEVEXP_SCAN_BUDGET_CEILING_MS` that may **lower** the ceiling and never raise it (`ceiling = min(ceiling, override)`): the seam cannot undo the cap, and the worst an ambient value can do is make a guard block sooner. Both suites now check that a 60 s budget under a 300 ms ceiling blocks at 0.3 s, and that a 600 s ceiling changes nothing. Documented next to the ceiling itself.

hooks/opencode/dangerous-cmd-guard.js
alexandrocuma: Follow-up: my first test for this asserted that the pass *refused*, which it does either way — with the parse unwatched, the masking walk's own first check catches it a parse later. Timing it turned out to be a poor discriminator too (the parse alone is ~230 ms on the shape I used, close to the bound).

So it is counted instead: `maskInert` is handed a budget that only increments a counter, and the pass has to consult it more than ten times on an 8 MB command (it manages about 120; unwatched it manages one, on the way out). The handler-level case stays time-bounded, which is what catches `blockReason` not passing the budget down at all.

One mutant I invented does survive and I have left it: removing the *walk*'s sampler. The parse runs first and consults the budget many times, so the walk's samples add nothing a test can see. Kept as defence in depth rather than relying on walk cost staying proportional to parse cost.

hooks/claude-code/scan-budget.sh
```diff
@@ -0,0 +1,198 @@
...
+# run, so the guard's body runs exactly once.
+devexp_scan_budget() {
+    local guard="$1"
+    shift
+
+    # The guard's prologue trap covers only the window before this function
+    # exists. From here the watchdog's 0-or-2 check is what floors an unknown
+    # status, the budgeted run included, and leaving the trap installed would
+    # make it fire on the guard's own clean exit.
+    trap - EXIT
```
alexandrocuma: **The last fail-open: this clears the floor before the function knows it can do its job.**

`trap - EXIT` is the first statement, but the constants this function needs are read afterwards. If the helper defines the entry point without them, `command -v devexp_scan_budget` in the prologue passes, the trap is already gone, and `set -u` ends the guard at **exit 1** — a non-blocking error, so the tool call proceeds unscanned.

Repro:

```
# helper = the function only, constants stripped
$ sed -n '/^devexp_scan_budget()/,$p' scan-budget.sh > <install>/scan-budget.sh
$ echo '{"tool_name":"Bash","tool_input":{"command":"ls -la"}}' \
    | bash <install>/dangerous-cmd-guard.sh; echo "rc=$?"
scan-budget.sh: line 19: DEVEXP_SCAN_BUDGET_PY: unbound variable
rc=1
```

This was the only non-2 result in a 17-case damaged-install sweep (missing, unreadable, syntax error, directory, empty, no entry point, `exit 9`/`exit 0` mid-source, sources-something-missing, symlink loop, dangling symlink, no-op stub, program text stripped, truncated early, truncated late — every one of those blocks correctly, and the truncation cases block precisely because the constants come first).

It's a narrow shape — a partial write that landed the tail but not the head, a hand-edit, or a future refactor that moves the constants below the function. But it's structural rather than accidental: the floor is lifted before the thing that replaces it is known to work.

Fix — clear the trap only once a decision is actually in hand, at the two exit points instead of the entry:

```bash
devexp_scan_budget() {
    local guard="$1"
    shift

    local arg
    for arg in ${1+"$@"}; do
        if [ "$arg" = "$DEVEXP_SCAN_BUDGET_SENTINEL" ]; then
            # The budgeted run is floored by the watchdog's 0-or-2 check from
            # here, and the trap must not fire on the guard's own clean exit.
            trap - EXIT
            return 0
        fi
    done

    local rc=0
    python3 -I -S -c "$DEVEXP_SCAN_BUDGET_PY" \
        ... || rc=$?
    if [ "$rc" != 0 ] && [ "$rc" != 2 ]; then
        echo "[devexp $guard] internal error -- ..." >&2
        rc=2
    fi
    trap - EXIT
    exit "$rc"
}
```

Note the clear has to go *after* the `python3` line, not before it — I tried the before-it variant first and it still exits 1, because the unbound constant is dereferenced on that very line.

Verified with this shape: function-only helper → **2**; healthy + allow envelope → 0; healthy + block envelope → 2; forced budget hit → 2; no `python3` on `PATH` → 2. No healthy path changes.

Worth a case in `scan-budget.test.sh` alongside the ones you added — the sweep you wrote covers everything else in this class.

hooks/claude-code/scan-budget.sh
```diff
@@ -0,0 +1,198 @@
...
+# character such as U+00B2 all fall back to the default. str.isdigit() accepts
+# the last two — one silently, the other by raising — and disagrees with the JS
+# twin, which is ASCII-only. Both sides trim surrounding spaces.
+ms = int(raw) if re.fullmatch(r'[0-9]+', raw.strip()) else int(default)
+
+# A budget that outlives the hook timeout is a fail-open, so it is capped
+# rather than honoured, and the cap is said out loud.
+if ms > ceiling:
+    sys.stderr.write(
+        "[devexp %s] Note: DEVEXP_SCAN_BUDGET_MS of %s s is at or above this guard's registered hook "
```
alexandrocuma: **Minor — this notice says something that isn't true, and the JS twin says it correctly.**

The clamp fires at `ms > ceiling` (44000), but the message claims the value is "at or above this guard's registered hook timeout" — which is 45000. For anything in 44001–44999 that's simply wrong, and it points the reader at the registry when what actually bit is the shared ceiling.

Side by side, same input (`DEVEXP_SCAN_BUDGET_MS=44500`):

```
shell: Note: ... of 44.5 s is at or above this guard's registered hook timeout,
       which would let Claude Code cancel the guard before it could block. Using 44 s.
js:    Note: ... of 44.5 s is above the 44 s ceiling the guards share. Using 44 s.
```

The JS wording is the accurate one and names the number the reader can act on. Suggest adopting it verbatim here — it also keeps the twins' messages identical, which is the property the parity suite exists to hold.

Separately: the shell prints this on every invocation (so, every tool call), while the JS side prints it once per process via `clampNoticed`. Not worth a mechanism in a short-lived shell process, but if the wording is unified it's worth a line saying why the frequency differs.

docs/reference/hooks.md
```diff
@@ -203,6 +208,113 @@ In `dangerous-cmd-guard.sh`, all parsing happens in the `python3 -I` step, where
...
+this existed stops running on the 600 s default at its next install.
+
+**How it is enforced.**
+
+| | Claude Code | opencode |
+|---|---|---|
+| Mechanism | `scan-budget.sh` re-runs the guard under a `python3` watchdog in a session of its own and `SIGKILL`s the process group at the deadline | a deadline started at handler entry, checked as the scan runs; the check throws a `ScanBudgetError`, which is how `tool.execute.before` refuses a call |
+| Covers | the whole hook run — reading the envelope, `json.load`, the regex work and every `grep` | the scan, at the granularity of one unit of work: one pattern, one rule against one line, one token, and the masking pass `dangerous-cmd-guard` runs first |
+| Cost | one extra `python3` **and** one extra `bash` — the guard is re-run as a child, not `exec`'d — so roughly **+70 ms** per guarded tool call on current hardware (measured, warm, 20 runs: `dangerous-cmd-guard` 72 → 138 ms, `secret-guard` 51 → 118 ms, `secret-in-write-guard` 52 → 121 ms; one Bash tool call, two guards in parallel, 72 → 143 ms) | none measurable — the masking pass's budget checks are sampled by position and came out within noise (−6% to +2%) |
```
alexandrocuma: **Minor — these numbers are from my first-pass measurement, before the sentinel, depth counter and ceiling landed; the prologue got a little heavier since.**

Re-measured on this head against `origin/main`, same machine, warm, 30 runs each:

| | main | this head | quoted here |
|---|---|---|---|
| `dangerous-cmd-guard` | 69 ms | **144 ms** | 72 → 138 |
| `secret-guard` | 50 ms | **135 ms** | 51 → 118 |
| `secret-in-write-guard` | 50 ms | **137 ms** | 52 → 121 |

So "+70 ms" is really +75 to +87 ms, and it's closer to 2.7× than 2× on the two cheaper guards. The claim's shape is right and the cost is still the right trade — but the row is presented as measured, and `CHANGELOG.md:73` repeats the +70. Worth refreshing both to the current figures.

alexandrocuma (commented):
## Re-review — 03a85d3

**Recommendation: one change, then merge.** Everything from the first pass is closed, both departures from my suggestions were correct, and I found one residual fail-open with a one-line fix.

### On the two departures — you were right, and about more cases than you said

**The EXIT-trap claim checks out, and it's worse than you described.** On this machine's bash 3.2 I measured what `$?` reads inside an `EXIT` trap:

| what failed | `$?` in the trap |
|---|---|
| source a **missing** file | **0** |
| source an **unreadable** file | **0** |
| source an **unparsable** file | **0** |
| **unset variable** under `set -u` | **0** |
| a sourced helper that itself sources something missing | **0** |
| source a **directory** | 1 |
| `false` / `exit 1` | 1 |
| command not found | 127 |

So my `case $rc in 0|2) ;; *) rc=2` would have turned **five** distinct non-blocking failures into an **explicit allow** — strictly worse than the 1 it replaced, since 0 is a clean "no decision, proceed" with not even an error notice. And the 0-vs-1 split isn't uniform, so no `$?`-reading trap could have been made safe. A trap that asks no questions is the right call; please keep the reasoning in the comment where it is.

**The fork storm is real and the counter earns its place.** Measured with the sentinel deliberately unrecognised:

| | stop time | peak guard processes |
|---|---|---|
| depth counter on | **0.21 s** | 4 |
| depth counter off (`MAX_DEPTH` raised) | 15.18 s | **536** |

Same order as your ~479. Worth noting the budget alone does eventually contain it — the outermost watchdog kills its group and exits 2 at the deadline, so even the storm fails closed — but 536 processes for 15 s is a real event and 0.21 s is much better. Reading the counter from the environment is sound: I confirmed every ambient value can only block or be ignored (`1`/`99`/`" 1 "` → block, `0`/`abc`/unset → normal), never switch anything off.

On the storm recurring: the depth travels on the child's own env via `Popen(env=…)` with no shell in between, so there is nothing to scrub it. The only way to lose it is to make `${BASH:-bash}` a wrapper — and I confirmed in the first pass that bash resets `$BASH` at startup, so it can't be pointed at one from the environment. No remaining path.

### The one thing to change

**Residual fail-open: the prologue trap is cleared before the function can be sure it can do its job.** `devexp_scan_budget` clears the trap as its first statement, and only then reads the constants it needs. If the helper defines the entry point but not the constants, `command -v` passes, the trap is gone, and `set -u` ends the guard at **exit 1** with no floor — the one non-2 result left in a 17-case damaged-install sweep. Repro and fix inline; I verified the fix closes it with no change to any healthy path.

### Everything else, verified

- **Prologue sweep, 17 damaged installs** — missing, unreadable, syntax error, directory, empty, no entry point, `exit 9` mid-source, `exit 0` mid-source, sources-something-missing, symlink loop, dangling symlink, no-op stub, program text stripped, truncated early, truncated late, plus the healthy control: **all block with a status of 2** and a message that names the case. Only "function without its constants" escapes.
- **The old ambient switch is inert.** The env marker no longer does anything; the sentinel travels in argv, which the watchdog writes directly via `Popen` with a list — no shell, no quoting, so the settings.json command string and any `sh -c` wrapper can't disturb it. A sentinel handed in by an edited registered command still turns the budget off (the scan itself still runs), but that's a tamper-evident edit to a file the installer rewrites, not an inherited value. Fine as is.
- **Budget parsing is now one spelling on both sides.** `"  250  "` → 250 both, `"  0  "` → 0 both, U+0663 → default both, U+00B2 → default both (no traceback), and `-1 / 1e3 / "" / " " / 0x10 / 1.5 / +5` → default both.
- **The ceiling behaves.** 44001 clamps, 44000 and below are silent, `DEVEXP_SCAN_BUDGET_CEILING_MS` only ever lowers (`999999999` is ignored, `0` blocks). The registry's 45 s sits just above it and both test suites pin the relationship.
- **The masking budget is not swallowed.** `ScanBudgetError` is re-thrown out of `maskInert`'s catch and reaches the handler as a refusal, while an unclassifiable command still falls back to "scan it all". The plugin dispatch has no catch around `tool.execute.before`, so nothing downstream swallows it either.
- **The isolation allowlist** is exactly the right shape — broad `*.sh` walk restored, `HELPERS` explicit, and the allowlisted helper must exist.
- **All suites pass** on this head: 8 Claude Code, 6 opencode, `go test ./...`.
- **Mutations: 20/20 killed** — including the walk sampler you logged as an equivalent survivor; the later test commits close it. No survivors.

### Two small factual points

- The cost table now quotes my first-pass numbers, but the sentinel, depth and ceiling work made the prologue a little heavier. Re-measured on this head against `origin/main`, same machine, 30 runs warm: `dangerous-cmd-guard` 69 → **144** ms, `secret-guard` 50 → **135** ms, `secret-in-write-guard` 50 → **137** ms. The shape of the claim holds; the numbers are ~15 ms optimistic per guard, and "+70 ms" is closer to +75–87 ms.
- The two twins word the clamp notice differently, and the shell's is the inaccurate one. Detail inline.

Nice work on the whole thing — the failure taxonomy, the argv sentinel and the counted second marker are a better design than what I suggested.

hooks/claude-code/scan-budget.sh
alexandrocuma: Fixed as suggested, and reproduced first: function-only helper → `rc=1`, `DEVEXP_SCAN_BUDGET_PY: unbound variable`, on both guards. After the fix, `rc=2` on both.

The trap now comes off at the two points where a status is known — in the budgeted run, where the watchdog's 0-or-2 check takes over, and after the `python3` line. Thank you for the note about *after*: I'd have put it before and shipped the same bug, since the constants are dereferenced on that very line.

One addition: the prologue sets `devexp_budget_why="could not run its scan budget"` once `command -v` passes, so this window names itself instead of inheriting the previous step's reason ("defines no entry point", which would have been misleading here).

`scan-budget.test.sh` gained the case, built the way your repro builds it (`sed -n '/^devexp_scan_budget()/,$p'`), alongside the other five. Four mutants cover the placement and all die: the floor lifted on entry again, lifted just before the `python3` line, left installed in the budgeted run, and left installed at the last exit.

hooks/claude-code/scan-budget.sh
alexandrocuma: Both fixed. The shell now carries the JS wording verbatim:

```
[devexp dangerous-cmd-guard] Note: DEVEXP_SCAN_BUDGET_MS of 44.5 s is above the 44 s ceiling the guards share. Using 44 s.
```

You're right that the old one was simply false between 44000 and 45000, and pointed at the registry rather than the number that bit.

The parity suite now compares the two notices directly — it spawns the shell twin and a child node process on the same over-ceiling value and requires the lines to be identical — so this cannot drift again. Two mutants die on it: the old wording restored, and the notice naming the asked-for value instead of the ceiling.

On frequency: a Claude Code hook is a process per tool call, so printing once there already *is* once per process; there is nothing to dedupe against without state on disk, which is not worth an I/O on every guarded call. The JS side needs its flag because that module outlives many tool calls. That is now a comment next to the notice rather than an unexplained difference.

docs/reference/hooks.md
alexandrocuma: Re-measured and published my own numbers with the method, as asked. Median of 30 warmed runs (3 warm-up runs discarded), `origin/main` materialised with `git archive` so neither tree is advantaged:

| | main | this head |
|---|---|---|
| `dangerous-cmd-guard` | 66 ms | 129 ms |
| `secret-guard` | 46 ms | 130 ms |
| `secret-in-write-guard` | 46 ms | 130 ms |
| one Bash call (two guards in parallel) | 66 ms | 130 ms |

So +60 to +85 ms, about double a guard's run — the same shape as yours, a few ms lower on both sides, which I'd put down to machine load rather than anything structural. Both the doc row and `CHANGELOG.md` now say +60 to +85 and quote this table, with your independent run noted alongside it.

PR #169, opened 2026-09-18
fix(hooks): allow only against proof that a guard's own scan ran (#168)

### LINKED ISSUE #168: Security guards: a no-op python3 on PATH makes them allow; opencode budget cost is misdocumented
## Problem
Both found in the PR #167 review (#162); both pre-date that PR.

1. **A `python3` earlier on `PATH` that exits 0 turns a guard into a no-op.** The Claude Code guards run their parsing and matching through `python3`. A different `python3` first on `PATH` — a wrapper, a shim, a broken virtualenv — that exits 0 without doing anything makes the guard allow the call. Verified in review: the tool call proceeds. The scan budget (#162) widens the blast radius from the parse step to the whole guard, since the wrapper also runs through `python3`.

   Worth considering: have each guard prove its interpreter did the work (a required marker on stdout/exit path that a no-op interpreter can't produce), and block when the proof is missing.

2. **`docs/reference/hooks.md` gives opencode's scan-budget cost as "none".** It is about +9% on a large input, because the deadline is sampled inside the scan loops. Correct the figure.

## Acceptance Criteria
- [ ] A `python3` on `PATH` that exits 0 without running the guard's code causes a **block**, not an allow, in every `fail_closed` guard.
- [ ] A test covers it with a stub interpreter on `PATH`.
- [ ] The opencode cost figure in `docs/reference/hooks.md` matches a measurement.

## Non-Goals
- Pinning an absolute interpreter path (breaks legitimate setups).

alexandrocuma:
Closes #168

## The problem
The Claude Code guards decided by reading `python3`'s exit status, so any `python3` earlier on `PATH` that exits 0 without doing the work — a wrapper, a shim, a broken virtualenv — made them **allow**. The scan budget added in v0.9.3 widened it: the wrapper runs through `python3` too, so the guard body never ran at all.

Measured before the fix, with stand-ins on `PATH` and envelopes that block under a healthy interpreter: **12 of 15 interpreter cases allowed**, across all three `fail_closed` guards, plus a `grep` that never matches allowing a destructive command.

## What changed
A guard now allows only against **proof that its own scanning code ran**, whatever the exit status says. Three independent checks, because an interpreter honest at one level and a no-op at the other still scans nothing (verified):

1. **The scan** — `scan-budget.sh` mints a per-process token; each guard passes it to its scanning program, which emits it as the first line of the output the guard already captures, *after* the scan has a verdict. The shell blocks when it is absent and strips it before reading the verdict.
2. **The budget watchdog** — a second per-invocation token, reported back on a dedicated file descriptor only after the guarded run's status is in hand. `close_fds` keeps that descriptor out of the guarded child, so no grandchild can hold it open past the deadline; the guard's stdout passes through untouched.
3. **`grep`** — `dangerous-cmd-guard` first asks one question whose answer it knows and blocks unless the count is exactly right. Catches never-match, always-match, error and miscount.

**Why accident can't satisfy it:** silence, an empty line, a generic success message, an echo of stdin or argv, a crash mid-scan and a wrapper that runs some other program all arrive without the token — each is a test case, each blocks. Tokens are per-invocation, so nothing can be baked in, and each is matched whole, so a marker with anything around it doesn't count.

**Stated limit** (in the code, the docs and the changelog): a program written to imitate the protocol on purpose is out of reach by construction — the program producing the proof is the one under suspicion. The property established is *the scanning code ran*, not *the interpreter is honest*. Pinning an absolute interpreter path remains the wrong fix; it breaks virtualenvs and version managers (#168 non-goal).

**v0.9.3's properties are intact:** prologue floor, argv sentinel, depth counter, budget and ceiling, the 0-or-2 rule, silence on an allow — `scan-budget.test.sh` 64/64 and `fail-closed.test.sh` 10/10 still pass. One subtlety handled: the constants are read into locals before the new command substitution, so the "helper with the entry point but not its constants" case keeps its exact prologue-trap behaviour.

**Also:** `docs/reference/hooks.md` said the opencode budget costs nothing measurable. Measured: ~+1 µs on ordinary input and +2–4% on the largest inputs. The separate "+9%" in the paragraph below the table was always a different comparison (per-token checking vs sampling), measured at 6–10% where the token loop dominates. Both figures are now labelled.

## Verification
- **After the fix,** every stand-in case exits 2 with a named reason; healthy paths still allow silently and block on their own reason.
- **New `interpreter-proof.test.sh`, 71 cases:** 7 stand-in shapes × 3 guards × 2 envelopes. It reads `registry.json`, so a future `fail_closed` guard is covered the day it's registered, and it asserts each guard actually calls the result check.
- **All suites pass,** Claude Code and opencode, installer scripts and Go.
- **Mutations: 10/11 killed.** The survivor is equivalent (emitting the proof before the masking pass): the only way the scan stops after the proof is written is an exception, which exits non-zero and is already blocked. Ordering kept as defence against a future early return.
- **Overhead: +0 to +6 ms per guarded call** (one extra `grep`, no extra interpreter start), against the +60 to +85 ms the budget itself costs.

## Out of scope (noted, not changed)
- `${BASH:-bash}` in the watchdog resolves to the interpreter already running the guard, so a hostile one means the guard is hostile.
- `cat` reading the envelope can only lie by emitting a different valid envelope; empty or malformed input already fails closed. A bash builtin read would drop a fork but truncate at a NUL byte — a new fail-open.
- A rogue interpreter's own stdout still reaches the hook's stdout while the guard blocks. That's the rogue program's noise; the guard adds none of its own, which the suite asserts.


alexandrocuma (commented):
## Security review — proof of work for the fail-closed guards

**Recommendation: request changes.** The two-level design does what it claims and I could not break it by accident: replayed tokens, tokens on stderr, noise printed before the token, and a stand-in honest at one level but not the other all block. One in-diff finding is a live fail-open of exactly the class this PR closes, so it should be fixed here rather than followed up.

I re-ran the v0.9.3 properties and the damaged-install sweep, mutated the new code, and measured the cost claim independently. Everything below was reproduced locally; repros are described at class level.

### Suites, as run here
`interpreter-proof.test.sh` 71/71 · `scan-budget.test.sh` 64/64 · `fail-closed.test.sh` 10/10 · `interpreter-isolation.test.sh` 11/11.

### Cost claim: verified
25 alternating warmed pairs per case against `main`, medians: `dangerous-cmd-guard` ordinary **+0.6 ms**, `secret-guard` ordinary **+0.7 ms**, `secret-in-write-guard` ordinary **+0.8 ms**, `dangerous-cmd-guard` on a 150 KB command **+0.9 ms**. Inside the stated **+0 to +6 ms**, and branch maxima were no worse than main's. (One earlier run showed a wild number for one case; it was CPU contention from a suite running in parallel, and it did not reproduce on a quiet machine.)

---

## 🔴 The `grep` canary certifies a code path the guard never takes

`dangerous-cmd-guard.sh` — the canary asks its question as `grep -cE <pattern>`: count mode, pattern as a positional operand. Every real check goes through `matches`, which is `grep -q <flags> -e <pattern>` on a here-string, sometimes with `-i`. So the question and the answer live in different modes.

A `grep` that is honest under `-c` but reports "no match" under `-q` passes the canary and then reports **every** blocked pattern as clean, silently, exit 0. I confirmed it end-to-end on the filesystem-wipe, force-push, hard-reset and drop-database envelopes: all four allowed with the stand-in, all four blocked with the real `grep`. That is the "a grep that never matches lets a destructive command through" case #168 set out to close, and the PR body, the CHANGELOG and the new **Proof of Work** docs section all state it is closed.

Same root cause, also uncovered: the canary's pattern is anchored literal alphanumerics, which is byte-identical in BRE and ERE, so it cannot notice `-E` being dropped — and on GNU `grep` an ERE containing `(` or `|` degrades to a silent no-match rather than an error. `-e` handling and `-i` are likewise never exercised.

**Suggested fix** — ask the question through the same call shape the checks use: `grep -q <flags> -e <pattern>` on a here-string, with an ERE-only construct (alternation, `+`) in the pattern, and require *both* a hit and a miss, plus one `-iE` hit. I prototyped a three-probe form of that: it blocks the `-q`-blind, `-q`-always and `-E`-dropping stand-ins, keeps blocking the never-match/always-match ones the current canary already catches, and leaves the healthy path allowing.

---

## 🟠 `close_fds` is load-bearing, stated in a comment, asserted nowhere

The proof channel changes the guard's wait from *wait for the watchdog process* to *wait for every holder of the pipe*. The only thing keeping a guarded descendant off the proof descriptor is `subprocess.Popen`'s default `close_fds=True`.

I verified the default does hold today — probing the descriptor from inside the guard's own scan child reports it closed. But turning it off passes all 71 cases of the new suite, and the guard's wall time then tracks the longest-lived grandchild instead of the budget: **0.2 s → 25.1 s** when the guarded run leaves a 25 s sleeper behind. That reinstates the #162 fail-open, since a hook that runs past the hook timeout does not block the tool call.

**Suggested fix** — pass `close_fds=True` explicitly so the dependency is visible, and pin it: the guarded run must not see the proof descriptor, and a lingering grandchild must not extend the guard's wall time.

---

## 🟠 Token freshness is claimed everywhere and pinned nowhere

"Minted per invocation, so nothing can be baked in" is the load-bearing claim for both tokens. Replacing either with a fixed literal **survives the whole new suite**. The `guess` stand-in only tries the literal text `devexp-scanned`, which is not the shape a baked-in constant would take.

The runtime behaviour is right — a token captured from one invocation and replayed into the next is refused, which I checked directly. It is only untested. **Suggested fix**: mint twice and assert the two differ, and assert the captured-then-replayed token is refused.

---

## 🟡 The new `exec 8>&1` adds a precondition the guard never had

With stdout closed, v0.9.3 behaved normally (ordinary input allowed, dangerous input blocked). On this branch both exit 2, preceded by a raw `scan-budget.sh: line 292: 1: Bad file descriptor` and then a branded message naming the wrong cause — it says the guard "could not run its scan budget", but it never got that far.

Fail-*closed*, so not a security hole, but it turns a missing stdout into "every tool call blocked" with a misleading diagnostic. Cheapest fix is to set the prologue's reason string truthfully just before the `exec`; better is to attach the descriptor to the command group rather than via `exec`, so a closed stdout does not end the shell mid-prologue.

---

## 🟡 Docs and CHANGELOG overstate the `grep` proof

The **Proof of Work** section says a grep that never matches, always matches, errors or miscounts is caught; the CHANGELOG says `grep` now has to answer a known question before it is trusted. Both are true only for breakage in the canary's own mode. Please scope the claim, or fix the canary and the claim becomes accurate as written.

---

## 🔵 Minor

- The new suite's behavioural loop iterates the hard-coded guard list, while the registry-derived list is only used for an equality assertion and a text check. The equality assertion does force a human edit when a new `fail_closed` guard lands, so the property holds — but computing the registry list first and looping over *it* would make "covered the day it's registered" literally true. The per-guard check is `grep -q` for two identifiers, which a mention in a comment would satisfy.
- File mode and the CI glob are fine — the new suite is picked up.

---

## ✅ What holds up

- **The two levels genuinely cover each other.** A stand-in that answers the scan's question but not the watchdog's is caught by the watchdog, and vice versa.
- **Replay, channel and ordering attacks all block**: a previous invocation's token, the token on stderr, and anything printed before the token each end at 2 with a named reason.
- **The guard's stdout still passes through untouched** — verified with a marker written from inside the guard.
- **Verdict stripping is exact.** The strip is non-greedy, so a multi-line verdict survives whole; a command containing token-shaped text still blocks; and the first line can only ever be written by the guard's own code, so a finding message cannot forge it.
- **Damaged-install sweep re-run**: every truncated-helper shape (missing token constant, missing result function, missing descriptor constant, entry point only, empty, absent) and the depth counter end at 2.
- The comment explaining why the constants are read into locals *before* the substitution is right, and the prologue-trap behaviour it protects still holds.

## Out of scope (noted, not filed)

`BASH_ENV`/`ENV` reaching the budgeted run through the inherited environment; `${BASH:-bash}`; an exported `SHELLOPTS=xtrace` making a guard emit trace output — including both tokens — on stderr. All pre-existing.

## Mutation survivors

4 of 14: constant watchdog token, constant scan token, proof-written-before-scan (the equivalent mutant the PR already calls out), and `close_fds` disabled.

hooks/claude-code/dangerous-cmd-guard.sh
```diff
@@ -557,6 +564,20 @@ matches() { # $1=grep flags  $2=pattern
...
 
+# `matches` reads grep's exit status and nothing else, so every check below is
+# only as good as that status (#168). One question with a known answer proves it
+# carries information: two lines, one of which matches a pattern made up here,
+# so the count is 1 for a working grep — and neither 0 (a grep that never
+# matches would report every command as clean) nor 2 (one that matches
+# everything, which over-blocks but would also mean the status says nothing).
+# The canary is alphanumerics and dashes only, so it is its own regex.
+devexp_grep_canary="devexp-grep-canary-$$-${RANDOM}"
+devexp_grep_answer=$(grep -cE "^$devexp_grep_canary\$" <<<"$devexp_grep_canary"$'\nnot-the-canary') || devexp_grep_answer='-'
```
alexandrocuma: **🔴 The canary asks its question in a mode `matches` never uses.**

This is `grep -cE <pattern>` — count mode, pattern as a positional operand. Every real check goes through `matches` just above, which is `grep -q <flags> -e <pattern>` on a here-string, and sometimes with `-i`. So the question certifies a code path the guard never takes.

A `grep` that is honest under `-c` but reports "no match" under `-q` passes this check and then reports **every** blocked pattern as clean, exit 0, in silence. I reproduced it end-to-end on the filesystem-wipe, force-push, hard-reset and drop-database envelopes: all four allowed with the stand-in, all four blocked with the real `grep`. That is the case #168 set out to close, and the PR body, CHANGELOG and docs all say it is closed.

Same cause, also uncovered: the pattern here is anchored literal alphanumerics, which is byte-identical in BRE and ERE, so it cannot notice `-E` being dropped — and on GNU `grep` an ERE containing `(` or `|` degrades to a *silent no-match* rather than an error. `-e` handling and `-i` are never exercised either.

**Suggested fix:** ask through the same call shape the checks use — `grep -q <flags> -e <pattern>` on a here-string — with an ERE-only construct (alternation, `+`) in the pattern, and require *both* a hit and a miss, plus one `-iE` hit. I prototyped a three-probe form: it blocks the `-q`-blind, `-q`-always and `-E`-dropping stand-ins, keeps blocking the never-match/always-match ones this canary already catches, and leaves the healthy path allowing.

hooks/claude-code/scan-budget.sh
```diff
@@ -165,6 +230,15 @@ except subprocess.TimeoutExpired:
     block(over)
 
 if rc in (0, 2):
+    # Proof that this watchdog ran the guard and has its decision in hand
+    # (#168). Written here and nowhere else: not on the way in, not on any
+    # path that never started the guard. Every exit above is a block, which
+    # needs no proof -- only an allow does.
+    try:
+        os.write(int(proof_fd), (proof + '\n').encode())
```
alexandrocuma: **🟠 `close_fds` is load-bearing but asserted nowhere.**

The proof channel changes the guard's wait from *wait for the watchdog process* to *wait for every holder of this pipe*. The only thing keeping a guarded descendant off this descriptor is the default `close_fds=True` on the `Popen` above — stated in the comment at the top of the file, checked by nothing.

The default does hold today: probing the descriptor from inside the guard's own scan child reports it closed. But turning it off **passes all 71 cases of the new suite**, and the guard's wall time then tracks the longest-lived grandchild instead of the budget — measured **0.2 s → 25.1 s** with a 25 s sleeper left behind by the guarded run. That reinstates the #162 fail-open, since a hook that outruns the hook timeout does not block the tool call.

**Suggested fix:** pass `close_fds=True` explicitly so the dependency is visible, and pin it — the guarded run must not see this descriptor, and a lingering grandchild must not extend the guard's wall time.

hooks/claude-code/scan-budget.sh
```diff
@@ -73,12 +101,49 @@ DEVEXP_SCAN_BUDGET_SENTINEL='--devexp-budgeted'
...
+#
+# A redirection needs a literal number, so the `9>&1` in devexp_scan_budget
+# spells it out; interpreter-proof.test.sh fails if the two ever disagree.
+DEVEXP_SCAN_BUDGET_PROOF_FD=9
+
+# The token a guard's own scanning step must report back before the guard may
+# allow. Minted per process, so it cannot be baked into anything; it travels to
+# the scanning step as an argument and comes back as the first line of the
+# output the guard already captures. See devexp_scan_result.
+DEVEXP_SCAN_PROOF="devexp-scanned-$$-${RANDOM}${RANDOM}${RANDOM}"
```
alexandrocuma: **🟠 Freshness is the load-bearing claim and no test pins it.**

"Minted per invocation, so nothing can be baked in" is what stops a stand-in carrying a token in advance. Replacing this with a fixed literal **survives the entire new suite** — same for the watchdog token below. The `guess` stand-in only tries the literal text `devexp-scanned`, which is not the shape a baked-in constant would take.

The runtime behaviour is correct: I captured a token from one invocation and replayed it into the next, and it was refused. It is only untested. Suggest minting twice and asserting the two differ, plus a captured-then-replayed case.

hooks/claude-code/scan-budget.sh
```diff
@@ -195,22 +269,49 @@ devexp_scan_budget() {
...
+    # channel existed. Inside a substitution `set -u` would only end the
+    # subshell, and the guard would name the wrong thing.
+    local program="$DEVEXP_SCAN_BUDGET_PY" \
+          default="$DEVEXP_SCAN_BUDGET_DEFAULT_MS" \
+          ceiling="$DEVEXP_SCAN_BUDGET_MAX_MS" \
+          sentinel="$DEVEXP_SCAN_BUDGET_SENTINEL" \
+          depth_var="$DEVEXP_SCAN_BUDGET_DEPTH_VAR" \
+          proof_fd="$DEVEXP_SCAN_BUDGET_PROOF_FD"
+
+    exec 8>&1
```
alexandrocuma: **🟡 New precondition: stdout must be open and dup-able.**

With stdout closed, v0.9.3 behaved normally — ordinary input allowed, dangerous input blocked. On this branch both exit 2, preceded by a raw `scan-budget.sh: line 292: 1: Bad file descriptor` and then a branded message naming the wrong cause: it says the guard "could not run its scan budget", but it never got that far.

Fail-*closed*, so not a security hole — but it turns a missing stdout into "every tool call blocked" with a misleading diagnostic. Cheapest fix is to set the prologue's reason string truthfully immediately before this line; better is to attach the descriptor to the command group rather than via `exec`, so a closed stdout does not end the shell mid-prologue.

docs/reference/hooks.md
```diff
@@ -315,6 +320,63 @@ parity tests use a zero budget or the real one.
...
+answer either — it breaks the virtualenvs and version managers people actually
+use.
+
+**`grep` is proved the same way** in `dangerous-cmd-guard`, where `matches`
+reads its exit status and nothing else. v0.9.1 made a grep *error* (a status
+above 1) block, but "no match" was still believed, so a `grep` that answered "no
+match" to everything would have reported every command as clean. Before the
+patterns run, the guard now asks one question whose answer it knows — two lines,
+one of which matches a pattern made up on the spot — and blocks unless the count
+comes back as exactly one. A grep that never matches, always matches, errors or
```
alexandrocuma: **🟡 Overstated.** This is true only for breakage in the mode the canary itself uses (`grep -cE`). A `grep` that never matches *under `-q`* — the mode every real check actually uses — passes the canary and reports every command clean; I reproduced that end-to-end. Please scope the claim, or fix the canary (see the comment on `dangerous-cmd-guard.sh`) and this sentence becomes accurate as written. The CHANGELOG entry carries the same claim.

hooks/claude-code/interpreter-proof.test.sh
```diff
@@ -0,0 +1,206 @@
...
+# ── Every fail-closed guard asks for the proof ──────────────────────────────
+# A new guard that forgets to is exactly the hole this closes, and it would
+# pass every other suite.
+registered=$(python3 -I -c '
+import json, sys
+print(" ".join(h["name"] for h in json.load(open(sys.argv[1]))
+                if h.get("opencode", {}).get("fail_closed")))' "$DIR/../registry.json")
+check "the fail-closed guards are the ones tested here" \
+    "$([ "$registered" = "$GUARDS" ] && echo 0 || echo 1)" "registry says \"$registered\", this file tests \"$GUARDS\""
+for guard in $registered; do
```
alexandrocuma: **🔵 Minor.** The behavioural stand-in loop above iterates the hard-coded `GUARDS`, while the registry-derived list is used only for an equality assertion and this text check. The equality assertion *does* force a human edit when a new `fail_closed` guard lands, so the property holds — but computing `registered` first and looping over it would make "covered the day it's registered" literally true. Also worth noting this check is `grep -q` for two identifiers, which a mention in a comment would satisfy.

hooks/claude-code/dangerous-cmd-guard.sh
alexandrocuma: Confirmed and fixed in `7726526`. I reproduced your stand-in first: all four destructive envelopes exit 0 in silence, all four block with a healthy `grep`. You were right that this was the class the PR claims to close.

The fix follows your prototype. There is now one place a pattern reaches grep — `devexp_grep "$flags" "$pattern" "$subject"` — and both `matches` and the probes go through it, so the question and the answers share a call shape by construction rather than by care. Three probes against a subject made up on the spot: an ERE-only hit, a miss, and an `-iE` hit.

On the `-E` case you flagged as reasoned rather than demonstrated: I built an argument-preserving stand-in (rebuilding `$@` with `set --` instead of re-splitting, which is what broke yours) and ran it end to end. On this machine's BSD `grep` a BRE containing `(` is an *error*, so the guard already blocked on grep's status — no fail-open to demonstrate here, and no GNU grep available to show the silent-no-match path you described. Rather than leave the property resting on the platform, the suite asserts the block is *attributed to the probes* for the `-E`-dropping stand-in, so the test fails if the probes stop being what catches it. Mutating all three probe patterns to BRE-safe equivalents kills that test (mutating only one does not — the other two still carry ERE constructs, which is why my first attempt at that mutation survived).

New stand-ins in the suite: `grep-q-blind`, `grep-q-always`, `grep-no-E`, `grep-i-blind`, alongside the never/always/error/miscount ones. `grep-i-blind` gets its own case with a command only a case-insensitive check blocks, so dropping the `-iE` probe is now a killed mutation instead of a survivor.

hooks/claude-code/scan-budget.sh
alexandrocuma: Confirmed and fixed in `7726526`. `close_fds=True` is now spelled out on the `Popen` with a comment saying why it is load-bearing.

Your measurement reproduced and then some: on my machine the same shape went 0.2 s → **544 s** with the sleeper (I let it run to completion rather than capping it), so the hook outruns any timeout by a wide margin.

Two tests, since the default holding today is not the property:
- the guarded run cannot see the descriptor — a synthetic guard probes it from inside the budgeted run and must find it closed;
- a process the guarded run leaves behind does not extend the guard — it backgrounds a 20 s sleeper and the guard must still finish in under 8 s. The sleeper's pid goes to a sweep file the suite kills on exit, so the test leaves nothing running.

`close_fds=False` is now a killed mutation.

hooks/claude-code/scan-budget.sh
alexandrocuma: Confirmed and fixed in `7726526`. Both constants were survivors; both are now killed mutations.

Three tests:
- the token differs between two invocations (sourcing the helper twice);
- a **captured, real** token replayed into a later invocation is refused — for the watchdog channel and, separately, for the scan step, using a stand-in that is honest for the watchdog so the scan check is what has to catch it;
- the `guess` stand-in now also emits a marker built to the right *shape* (`devexp-scanned-1-234567`, `devexp-budget-1-234567`), not just the literal text it tried before.

One thing worth recording: my first version of the replay test passed for the wrong reason. It recovered the tokens with `grep -o` over a file of recorded argv, and that file holds the guards' program text, so grep reported `Binary file … matches` and the "replay" was garbage. The capture stand-in now matches by prefix and writes each token to its own file, and a check asserts both were actually captured — a replay test with nothing to replay proves nothing.

hooks/claude-code/scan-budget.sh
alexandrocuma: Confirmed and fixed in `7726526`, taking the second of your two options.

fd 8 is now attached to a call (`devexp_scan_budget_run … 8>&1`) rather than to the shell, since a failed `exec` redirection is what ended the guard mid-prologue. Before that, the guard asks quietly whether stdout can be duplicated at all — `{ : >&8; } 2>/dev/null 8>&1`, with stderr silenced *before* the failing dup, or the shell's own complaint is the noise this was meant to remove — and when it cannot, the guarded run's stdout goes to `/dev/null`.

So a caller with no stdout now gets a decision instead of a block: ordinary input allowed, dangerous input blocked, nothing on stderr — v0.9.3's behaviour, which is what your comparison showed. Covered for all three guards in the suite.

docs/reference/hooks.md
alexandrocuma: Fixed in `6a57e34`. The claim now says what the probes establish rather than "never matches, always matches, errors or miscounts":

> a grep that answers wrongly **in the mode the checks use** is caught — never matching, always matching, blind or always-true under `-q`, blind under `-i`, silently dropping `-E`, erroring, or returning a status that carries no information. A grep that answers all three probes correctly and then lies about one particular pattern is not caught, and is the same class as an interpreter built to imitate the protocol.

The CHANGELOG entry carries the same scoping, and the PR body is updated.

While I was in that section I also had to correct something of my own. You noted you had not verified the revised cost figures; re-running them showed the +2 to +4% I published does not reproduce — paired runs swing between −3% and +12% on a 250 ms scan, so I was reporting GC and JIT variance. The section now counts the clock reads instead, which is deterministic: 145 on an 800k-token command, 2,025 on a 200k-line command, 13 on `ls -la`, at ~22 ns each — under 0.05 ms on any input. The 9% figure stays as what it always was (checking at every token instead of sampling), now with the arithmetic that makes it plausible.

hooks/claude-code/interpreter-proof.test.sh
alexandrocuma: Fixed in `7726526`. `GUARDS` is now computed from the registry and every behavioural loop iterates it, so a new `fail_closed` guard is driven by all the stand-ins the day it is registered.

The equality assertion is gone, replaced by something that fails more usefully: each registered guard must have both envelopes in the file, named in the failure message. And you were right about the text check — it is now anchored as code (`^devexp_scan_result <guard>( |$)` and `^[^#]*"\$DEVEXP_SCAN_PROOF"`), so a mention in a comment no longer satisfies it.

alexandrocuma (commented):
## Re-review at `6a57e34`

**All six findings are fixed, and I verified each one rather than taking the description.** Recommendation: **merge**, with two small follow-ups below — neither blocks.

Suites here: `interpreter-proof` 103 · `scan-budget` 64 · `fail-closed` 10 · `interpreter-isolation` 11 · `dangerous-cmd-guard` 380 · `secret-guard` 37 · `secret-in-write-guard` 321, all passing. **All four of my round-1 mutation survivors now die.**

### Confirmed fixed

- **The single funnel is real.** I audited every `grep` invocation across the three guards and the helper: there is exactly one, inside `devexp_grep`. All eight mode-level stand-ins now block — `-q`-blind, `-q`-always, `-E`-dropped, `-i`-dropped, never-match, always-match, error, bogus-count — checked against four dangerous shapes plus a benign one. The round-1 bypass is gone.
- **The descriptor.** Closed in the guarded run *and* in a grandchild. Wall time stayed at **0.2 s** with two 20 s sleepers left behind, including one detached into its own session — the escape route the process-group kill cannot reach. Well under the 8 s bound.
- **Freshness.** 16 parallel invocations → 16 distinct scan tokens and 16 distinct watchdog nonces; 6 same-second serial runs → 6 distinct; two different guards at once → 2 distinct. (My first pass reported "1 distinct" — that was a stray `%%s` in my own instrumentation, not the guard. Flagging it because I nearly filed it.)
- **The dup probe.** With stdout closed, a benign envelope now allows and a dangerous one blocks, with **no noise on stderr** either way; pipe, file, `/dev/null` and a read-only fd 1 all behave. Guard stdout still reaches the caller when there is one.
- **Docs, CHANGELOG, registry-driven guard list, anchored structural check** — all as described.

### 🟡 Follow-up: the probes certify the call shape, not the regex dialect

The probes now share `-q`, `-e` and the here-string with the real checks, which is what mattered. What they still do not share is the **pattern vocabulary**: they use only anchors, alternation, a digit class, `+` and `-i`, while the real patterns lean on `\s`, `\b`, `{m,n}` and bracket classes.

A stand-in that answers all three probes correctly but refuses patterns containing one of those constructs gets through. Measured on this machine:

| stand-in refuses patterns containing | result |
|---|---|
| `\s` | **all four dangerous commands allowed, exit 0, in silence** |
| `\b` | two of four allowed |
| `{` | two of four allowed |
| `[a-z]` | one of four allowed |

In every case I confirmed directly that all three probes pass. Separately, the probe subject is a single line, so a grep that only ever sees the first line passes all three and then allows a dangerous command that is not on line 1.

The docs file this under *"answers all three probes correctly and then lies about one particular pattern … the same class as an interpreter built to imitate the protocol."* I'd push back on that one sentence: `\s` is a GNU extension, not POSIX ERE, so a strict-POSIX or busybox-style grep reads it as a literal and under-matches **by accident**, not by imitation. It belongs in the same family as the `-E` case this round already closed.

Cheap fix, and I prototyped it: one more probe pair whose pattern carries the same constructs the real patterns use, against a multi-line subject whose hit is not on the first line. That blocks the `\s`, `\b`, `{` and first-line stand-ins while the healthy path still allows. Add a bracket class to the probe pattern and the fourth row goes too. Then the docs sentence becomes accurate as written.

### 🔵 Three unpinned properties (mutants that survive all 103 cases)

- **`matches` routing through `devexp_grep`.** Re-inlining the grep call in `matches` passes everything. Behaviour-preserving today because the shape happens to match — but the single funnel is the whole basis of this round's fix, and nothing enforces it. The file already anchors a structural check for `devexp_scan_result`; the same treatment for `devexp_grep`, plus "no other `grep -q` in this file", would close it.
- **The dup probe's positive branch.** Forcing it to always report "no stdout" — so every guard's output goes to `/dev/null` — passes everything. Passthrough does work at this head; it is just unasserted, while the docs claim it.
- **The first of the three probes** can be deleted and the suite stays green. Largely redundant with the `-iE` probe, so low impact, but the suite cannot tell one probe from three.

### 🔵 Cost wording

The three probes in isolation measure **5.18 ms** per guarded call here, matching the "5.2 ms against 2.2 ms" in the CHANGELOG — but that is a +3.0 ms delta in grep alone, while the same sentence says "+0 to +2 ms per guarded call". End-to-end, interleaved pairs against v0.9.3 gave **+0.4 to +4.8 ms** (ordinary input ~+1 ms; one large-input pair at +4.8 ms). **"+0 to +5 ms"** would be safer than "+0 to +2 ms".

### On the opencode retraction

Counting clock reads is a fair substitute when the end-to-end difference sits below run-to-run variance, and it has the advantage of being reproducible. The arithmetic checks out: 2,025 × 22 ns = 0.045 ms, 145 × 22 ns = 0.003 ms, 800k × 22 ns ≈ 18 ms. The 6–10% figure is now correctly scoped as per-token versus sampled, with the arithmetic beside it. One clause worth adding: the count covers the **clock reads**, not the rest of the budget (the per-unit branch, the sampling counter, the deadline object), so "under 0.05 ms" is a floor for the budget's cost rather than the whole of it — worth saying, since the table row is labelled *Cost*.

### On the GNU-grep caveat

It does not matter for the merge. The dropped-`-E` case is now caught by the probes directly, so nothing rests on how it would have degraded on GNU. The underlying class — a dialect difference that silently under-matches instead of erroring — I was able to demonstrate directly on this machine's BSD grep via `\s`, `\b` and `{`, so no GNU box is needed to see it.

alexandrocuma:
Thanks — the follow-up was a live allow, so it is fixed in `a75611e` rather than deferred.

## The dialect gap

Reproduced your table first, on this machine: `\s`-blind allowed all four dangerous commands silently; `\b` and `{` two of four; `[a-z]` one of four. And you are right that it does not belong in the "purpose-built imitation" bucket — `\s` is a GNU extension, so a strict-POSIX or busybox grep under-matches by accident. That sentence in the docs is rescoped: the bounded claim is now *a grep that answers wrongly in the mode **or the dialect** the checks use is caught*, and the same wording is in the CHANGELOG and the PR body.

I took your probe shape but derived the vocabulary from the patterns instead of listing it. Each probe pattern now carries `\s`, `\b`, `\S`, a POSIX class, a bracket range, a literal brace, alternation, a group, `+`, `*`, `?` and both anchors, against a two-line subject whose answer is on the second line. And the drift you warned about is a test: the suite extracts the pattern-bearing lines from the guard and the probe patterns, and fails when a construct appears in the first and in none of the second. That check already earns its place — mutating the probes to drop `[[:alpha:]]` fails with `no probe exercises: [[:class:]]` rather than with a mysterious stand-in result.

Two notes on the design. I folded your extra pair into the existing three rather than adding to them, because a fourth and fifth grep cost more than they add: the hit, the miss and the `-iE` hit each already carry the full vocabulary, and each is now the only probe that catches one of the stand-ins. And `{m,n}` is deliberately absent — the catalogue says the real patterns do not use a bounded repeat, only a literal `{` inside a bracket expression, so that is what the probes exercise. If a pattern ever gains one, the drift check will say so.

New stand-ins, all blocking: `grep-s-blind`, `grep-b-blind`, `grep-brace-blind`, `grep-class-blind`, `grep-first-line`.

## The three unpinned properties

- **The funnel** is structural now, the way you suggested: exactly one `grep` invocation across the three guards and the helper, and it must fall inside `devexp_grep`'s line span. Re-inlining it in `matches` fails both halves. (The check skips quoted heredocs, so the `'grep'` inside the embedded Python program is not miscounted.)
- **The dup probe's positive branch**: a synthetic guard writes a marker to stdout after the budget takes over and the test requires it back. Forcing "no stdout" now fails.
- **The first probe**: with the vocabulary folded in, each of the three is load-bearing. `drop-first`, `drop-miss` and `drop-iE` are all killed mutations.

## Cost

You were right that "+0 to +2 ms" did not square with a +3 ms grep delta, and chasing it turned up something better than a wider range. Over 40 paired runs: the two secret guards are **+1 ms**, `dangerous-cmd-guard` is **+6 ms** (128.8 → 134.7 median), the two-guard call **+3 ms** (186.9 → 189.5). The three probes on their own are 4.9 ms of grep against 2.0 ms.

The scatter you saw is a real bimodality, and it is not the probes' doing: this machine occasionally spends ~50 ms extra on a run, in v0.9.3 too (2 runs in 40). With two more short-lived processes it happens more often — 14 in 40, and 17 in 40 for the two-guard call. Bare `grep`, `true` and `python3 -c pass` spawns are not bimodal, so the tail belongs to the environment and the probes just buy more tickets in it. Both the medians and the tail are in the docs rather than a single tidy number.

## The opencode clause

Added, in the table row and in the paragraph: the count covers the clock **reads**, not the per-unit branch that decides whether to read, the sampling counter it tests, or the deadline object each handler builds — so "under 0.05 ms" is a floor for the budget's cost, not the whole of it.

## Suites you did not run this round

Ran here: opencode 434 / 130 / 61 / 81 / 38 / 301, installer 21 / 24 / 250, Go `gofmt` + `vet` clean and `go test -race ./...` green, alongside the nine Claude Code suites (`interpreter-proof` is now 120). Mutations across the three sets: **23 of 24 killed**, the survivor being the acknowledged equivalent (proof emitted before the masking pass). `matches` re-inlining, both constant tokens and `close_fds=False` all die now.

On the six empty review events: noted, nothing done.

